PR #134, opened 2020-06-01
Read promised text

### LINKED ISSUE #128: Pluto seems to work fine in Firefox but not in safari.
Hi,

I just downloaded Pluto to try it out, but was surprised that I was stuck on 
image
but same when run on Firefox, it seems to be up and dandy.

CiaranOMara:
Appears to resolve #128.

Chrome
image

Safari
image

Firefox
image

### DIFF
--- a/assets/client.js
+++ b/assets/client.js
@@ -79,8 +79,30 @@ class PlutoConnection {
         return this.send(messageType, body, cellID, true)
     }
 
+    readBlob(blob) { // Event-based fill-in for newer Blob.text().
+
+        const reader = new FileReader()
+
+        const p = new Promise((res, rej) => {
+            // on read success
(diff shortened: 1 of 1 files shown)


fonsp:
I'm not sure what to do here - of course I appreciate making it work on safari (although I would still like to do more tests), but avoiding compatibility code like this is one of the reasons why only Firefox and Chrome are supported right now. My idea was to add other browsers (including safari) using polyfills later, but if this is the _only_ incompatibility, then it makes sense to add this in.

Another problem is that I don't have a WebKit browser available - so I would not be able to test any future changes myself, i.e. WebKit would officially remain unsupported.

But I'm no expert in all of this, what do you think? And @dralletje, can you comment?

PR #136, opened 2020-06-01
use javascript modules

dralletje:
Instead of using `window` to store all shared variables globally, use the es6 module system (natively supported in Chrome and firefox).
This doesn't make it magically clean, but it does allow for a more manageable javascript file structure.

For now I let all the files in place, but made as much as possible use of imports still


dralletje:
Also please squash this 😁

dralletje:
(Next to merge there is a dropdown where you can select "Squash and merge" to make this all one commit)

fonsp:
Yay! I will look at it soon 👀

Does it work on safari?

dralletje:
Yes

Issue #77, opened 2020-04-12
A nicer filenotfound error page


fonsp:
Not great:

![image](https://user-images.githubusercontent.com/6933510/79081576-118af900-7d1f-11ea-9282-a2852a1ecb77.png)

fonsp:
Tips:

Write a basic `error.html` file (I already started), and write something like `{{ error message here }}` inside the .HTML script. When there is an error (`Static.jl`), open the file, and use string replace to put the error in.

TheCedarPrince:
Question: just to confirm, is the error file you are referencing this one: [`assets/error.jl.html`](https://github.com/fonsp/Pluto.jl/blob/master/assets/error.jl.html)? Thanks for clarifying @fonsp

TheCedarPrince:
Started working on draft PR #148

PR #148, opened 2020-06-05
Better File Note Found Page

TheCedarPrince:
So far, this is what I have been working on (started from #77):

image

I am thinking the error page can be very similar to the original welcome page but can also display the error message. Working on adding the recent files part as well. 

Running into a couple difficulties on how to do this however. @fonsp how do I:

1. Use an `HTTP.Response(404)` to redirect the user to another page? I tried looking at the documentation but was confused by how to do this.

2. Do I want to pipe an error to this page from the back-end or do I want to explicitly say, when triggering a 404 error, "file not found" essentially?

Thanks!


fonsp:
1. The _page response_ is not a 404 but a 200 (the default) - trying to view `http://localhost:1234/some_unknown_address` would be a 404. 
2. It has to be the Julia error - file not found is not the only possible error.

fonsp:
Also, it looks like this is the same branch as your previous PR - can you delete this one and start from a clean fork? Or does that not make sense

TheCedarPrince:
Yup, this should be on a separate branch. Do you mind deleting this PR while I set-up a new PR with a new branch?

fonsp:
Try to fork off the latest version of the master branch from `fonsp/Pluto` (I don't know how git is hard)

PR #149, opened 2020-06-05
Nice File Not Found Page

TheCedarPrince:
(Accidentally started on an old branch so am restarting this PR again from PR #148 on a new development branch)

So far, this is what I have been working on (started from #77):

image

I am thinking the error page can be very similar to the original welcome page but can also display the error message. Working on adding the recent files part as well.


TheCedarPrince:
Ok, this PR is ready for review. Here is how it looks:

image


I added in a place for you to put in error messages. I assume this is something you need to do, but I cannot get the `filepicker` or `Recent Sessions` to load. Let me know if there is anything else needed! :smile:

fonsp:
Hmmmm I'm not sure about this design

Can you make the title at most 3 words, the error message left-aligned (try it out with an actual multi-line error message instead of the placeholder), and remove all other content? You could include a link back home, but your browser already has a button for that.

TheCedarPrince:
Hey @fonsp here is the new design:

image

I tested it out on different screen dimensions and it seems to render properly. Since I was duplicating myself a lot in the welcome.css, I decided to make an error.css to not mess with the main css layout for the web interface.

What do you think now?

Issue #115, opened 2020-05-16
include file containing structs


karlwessel:
It currently is not possible to reevaluate a cell that includes a file which defines a `struct`. 

For example when I have a file `mystruct.jl` that contains 
```
struct A end
```

and include it in a notebook with
```
include("mystruct.jl")
```
it works when I execute the cell the first time, but the second time I get the following error message:
```
LoadError: cannot assign a value to variable workspace10.A from module workspace11
top-level scope at mystruct.jl:1
include(::Module, ::String) at ./Base.jl:377
include(::String) at .julia/packages/Pluto/3bf5m/src/react/WorkspaceManager.jl:65
in expression starting at mystruct.jl:1
```

fonsp:
This can be fixed! And!

The workspace cleaning could be made more general, and also clean up after code from another file. For example, when your `include`d script adds an extension to `Base.sqrt`, this could be deleted by Pluto when you remove the `include`. This would set the stage for notebooks-importing-notebooks reactivity.

fonsp:
Related, but not quite the same, is that right now you have to restart your workspace (i.e. restart Pluto) when you change a package and want to import it again. 👉 #78

karlwessel:
If this is more complicated to solve and is only a long term goal should it maybe display a more informative error message mentioning that this is currently not supported and what a workaround may be?

fonsp:
Hm yes perhaps that's a good idea. Maybe it can be done [with the existing error rewriting mechanism](https://github.com/fonsp/Pluto.jl/blob/master/assets/editor.js#L874-L887) - give it a try if you like!

karlwessel:
I tried to implement this using the rewrite mechanism but can't get the regular expression to work (I have no experience with javascript):

```
{
    to: "Reloading cells that include files defining structures does not work yet. <Placeholder, replace with something better>",
    from: RegExp("/LoadError: cannot assign a value to variable workspace\\d+\\..+ from module workspace\\d+/")
}
```

fonsp:
Hm I see, maybe you can replace
`"LoadError: cannot assign a value to variable workspace"`
with
`"<nice error message>\n\nLoadError: cannot assign a value to variable workspace"`

This would still show the original error, but that way you don't have to use a regular expression :)

karlwessel:
Ok, that is a solution that would work.

Does the `replace` method not support regexes? According to the documentation I thought it would.

fonsp:
[...]

PR #156, opened 2020-06-08
helpful error message when including files that define a struct

karlwessel:
Including files that define structs is not yet supported (see issue #115). Until that is fixed this should give a more helpful error message saying 
```
Tried to reevaluate a cell that includes a file containing structs.
This is not yet supported :(.
As a workaround you can put the included code in a module and use it with `using <module>` in the notebook.
However, changes to that file will only take effect after restarting Pluto.
```

### DIFF
--- a/assets/editor.js
+++ b/assets/editor.js
@@ -904,7 +904,11 @@ function cellRedirect(event) {
 const errorRewrites = [
     {
         from: "syntax: extra token after end of expression",
-        to: "Multiple expressions in one cell.\n<a href=\"#\" onclick=\"errorHint(event)\">Wrap all code in a `begin ... end` block.</a>",
+        to: "Multiple expressions in one cell.\n<a href=\"#\" onclick=\"errorHint(event)\">Wrap all code in a `begin ... end` block.</a>"
+    },
+    {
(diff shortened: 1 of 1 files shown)


fonsp:
The workaround still doesn't let you reimport, so... isn't the workaround just to _not_ re-evaluate the `include`?

fonsp:
Thanks for the specific regex pattern by the way!

fonsp:
So how about:

```html
<p>Tried to reevaluate a cell that includes a file containing structs, this is not yet supported 😢.</p>
<p>As a workaround, you can <em>restart</em> the notebook in the <a href="./">main menu</a>.</p>
```

karlwessel:
That workaround was the best that came to my mind. `using` at least doesn't throw an error message when it is (accidentally) evaluated. 

Another workaround, if Pluto is used to actively develop the code from the included file, is to move the structs (and maybe their methods) out of the file into the notebook itself and only if it solidifies move it to a module and include it with `using`. That at least, was the workaround I used.

But that was to long to explain in an error message. So your proposal is fine :-)

fonsp:
Thanks! I've committed this to the main branch directly because of git troubles: https://github.com/fonsp/Pluto.jl/commit/a3f17f25793691c059e7e7ac88a26040b9f0ad3e

Issue #179, opened 2020-06-11
Open default browser on run


mgautam98:
Can we open the default browser with the localhost:port on run?

mgautam98:
I think we can detect the platform and run a system call within julia to open the default browser with the mentioned port. 
Can I take this issue?

fonsp:
Cool! It should just be [a single line of code](https://github.com/fonsp/Pluto.jl/blob/master/src/webserver/WebServer.jl#L222), but I also want it to work for the special case that you are running on WSL. Do you have WSL by any chance?

mgautam98:
Yes, I am running WSL2 I have tested it with WSL2 It works fine.

fonsp:
Nice! Make it a PR!

fonsp:
Thanks again! Sorry it took so long

MikaelSlevinsky:
This is a really cool feature :+1: but I might never change Safari as my default browser. For Pluto, is there a way to try/catch and look for Chrome/Firefox instead of default?

MikaelSlevinsky:
Oh I'm sorry, I didn't realize Safari support has been fixed! https://github.com/fonsp/Pluto.jl/pull/289

PR #190, opened 2020-06-15
add launbrowser on run

mgautam98:
Towards #179 
Opens the default browser with the specified port. 

Verified and tested on:
1. Windows 10 Pro
2. WSL2 (Distro: Ubuntu 18.04) on Windows
3. Debian (Kali Linux)


src/webserver/WebServer.jl
```diff
@@ -219,7 +219,11 @@ function run(host, port::Integer; launchbrowser::Bool = false)
     println("Press Ctrl+C in this terminal to stop Pluto")
     println()
     
-    launchbrowser && @warn "Not implemented yet"
+    if launchbrowser
+        Sys.islinux() && Base.run(`xdg-open $address`)
```
fonsp, line 223: This does not work on WSL - is WSL2 different? Did you install `xdg-open` yourself?

fonsp:
And can you put everything in a `try catch` block?

src/webserver/WebServer.jl
mgautam98: No I have not installed xdg-open I will check again on a fresh distro

fonsp: Try running `xdg-open https://github.com/` in WSL2

mgautam98: > Try running `xdg-open https://github.com/` in WSL2

This works fine for me in WSL2. I you don't put http/https then it can not open.

mgautam98:
Hey I found that on WSL1 we [can not open](https://unix.stackexchange.com/questions/510624/ubuntu-for-windows-cant-open-pdf-files-via-xdg-open) Host system browser(Or any application) I am trying to find a workaround this.

src/webserver/WebServer.jl
fonsp: I found this for WSL Legacy: https://github.com/sindresorhus/open/blob/master/index.js#L71

```julia
Base.run(`cmd.exe /s /c start "" /b "https://github.com/"`)
```

It works on my machine :)

fonsp: The same command also works for Windows

mgautam98: if you want to just find wls1 we can do it https://github.com/microsoft/WSL/issues/4555#issuecomment-536862561 because using calling cmd from wsl show the following log
```
'\\wsl$\Ubuntu-18.04\home\mgautam\.julia\dev\Pluto'
CMD.EXE was started with the above path as the current directory.
UNC paths are not supported.  Defaulting to Windows directory.
```

mgautam98:
@fonsp can you check now

fonsp:
Looking good! Just some style comments:
- the same command can be used for the Windows case and the WSL case
- the error message should go
- Julia has a built-in way to check for WSL

RocketRoss:
I'm keen to help finish this off, only confused about @fonsp 
```Julia has a built-in way to check for WSL```
I cannot find related information... 

I also don't see an issue with leaving the error message? Just curious about what it was requested to go, won't argue <3

mgautam98:
> I'm keen to help finish this off, only confused about @fonsp
> `Julia has a built-in way to check for WSL`
> I cannot find related information...
> 
> I also don't see an issue with leaving the error message? Just curious about what it was requested to go, won't argue <3

There is currently no built in way to do this. But I opened a PR in JuliaLang too for this, https://github.com/JuliaLang/julia/pull/36425

All the work is already done I don't think there is anything to complete @RocketRoss

RocketRoss:
Yeah I see that. Nice! 
Could you share the julialang request?
I just want it to be merged. Keen to have the observable notebook environments. So even if my comment just serves to relight the fire (:

fonsp:
Ah you're right, not sure where I got that from

fonsp:
Sorry @mgautam98 I'll merge this soon!

fonsp (approved):

fonsp:
Thanks @mgautam98!

PR #219, opened 2020-07-16
🎨 Make customization a bit easier again

Pocket-titan:


fonsp:
Nice, I was also hoping to have a `light.css` that defines the current theme colors using CSS variables, and then `editor.css` can use those variables. Changing the theme would be swapping out `light.css` for `rainbow.css`. That way, `editor.css` can stay the same for all themes (and we will still only have one style sheet to maintain).

frontend/components/CellInput.js
```diff
@@ -1,7 +1,19 @@
 import { html, useState, useEffect, useRef } from "../common/Preact.js"
-
 import { utf8index_to_ut16index } from "../common/UnicodeTools.js"
 
+// For customization, see: https://codemirror.net/doc/manual.html#config
+// Note that some options will require you to add additional addons in editor.html
+const custom_codemirror_options = {
```
fonsp, line 6: Why does this make customization easier?

fonsp: Maybe add a comment explaining how people can customize these? (A fork?)

Pocket-titan:
> Nice, I was also hoping to have a `light.css` that defines the current theme colors using CSS variables, and then `editor.css` can use those variables. Changing the theme would be swapping out `light.css` for `rainbow.css`. That way, `editor.css` can stay the same for all themes (and we will still only have one style sheet to maintain).

That sounds good, but by "theme" do you mean the Pluto styles excluding synax highlighting/the codemirror editor, or including? Right now I have them separated (they could be separated more still) to match how codemirror handles themes

frontend/components/CellInput.js
Pocket-titan: Right now it's just a separate object containing settings the user should be able to change (e.g. don't change `mode: "julia"`), but I'm not sure yet about the implementation (forking isn't ideal b/c every update will give merge conflicts), maybe a settings menu that persists to localstorage?

fonsp:
Both, they should have matching color schemes

PR #247, opened 2020-08-02
Update editor.css to add CSS for table outputs

mthelm85:
Adds some CSS for table outputs. Ultimately, I think it would be nice to have a full-blown data table (like this one: https://www.npmjs.com/package/react-data-table-component#demo-and-examples) but I think this is an improvement.

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -151,6 +151,8 @@ pluto-output code {
 pluto-output pre {
     display: inline-block;
     margin: 0px;
+    white-space: pre-wrap;
+    word-break: break-all;
     tab-size: 4;
     -moz-tab-size: 4; /* https://bugzilla.mozilla.org/show_bug.cgi?id=737785 */
     font-family: "Roboto Mono", monospace;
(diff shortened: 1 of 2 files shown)


fonsp:
Could you post a screenshot and a notebook that generates a table to test it on?

mthelm85:
Sure! 

Here's what it looks like right now:

table_output_standard

And here's what it looks like with the simple CSS additions:

table_output_css

Would it be worth it to look at TableView.jl to see what it would take to make that package compatible with Pluto or should the solution ultimately be a React solution? I looked at Nextjournal and the nice table output that they have seems to be produced with Tailwind CSS. I don't have any experience with that framework but I'd be happy to look into it...

fonsp:
Thanks! For those who don't know how nextjournal formats tables (including me) - what do you specifically like to see? You mention React - do you want a dynamic table view, with built-in sorting, pagination, etc? This is not trivial to implement, because you need to lazily access and compare table data on the Julia process. But it's possible, and interesting!

Same for TableView.jl - what do you specifically like about it? Why do you think that it is not compatible?

fonsp:
Ah I see that TableView.jl is WebIO-based. That might be worthwile to support.

mthelm85:
> Thanks! For those who don't know how nextjournal formats tables (including me) - what do you specifically like to see? You mention React - do you want a dynamic table view, with built-in sorting, pagination, etc? This is not trivial to implement, because you need to lazily access and compare table data on the Julia process. But it's possible, and interesting!
> 
> Same for TableView.jl - what do you specifically like about it? Why do you think that it is not compatible?

Ultimately, I think a dynamic table view with built-in sorting, filtering and pagination would be great. I'll be presenting two Pluto notebooks at work this coming Thursday and I need to display some tabular data for the audience and having that ability to manipulate the table within the output cell would certainly be the ideal situation : )

mthelm85:
Hi! Do I need to do anything else?

fonsp:
Hey! Thanks for the initiative! I modified your CSS to match the overall look & feel that I am going for.

image

image

Large tables scroll horizontally:

image

mthelm85:
Looks good! Thanks for all your hard work on this, I'm quickly becoming addicted to Pluto and I'll be presenting a couple of notebooks at work on Thursday : )

fonsp:
Oh that's nice to hear! You suggestions have been helpful, keep them coming!

Issue #252, opened 2020-08-04
kwagrs for macro calls


mcmcgrath13:
I started working on supporting rendering of the vega and vegalite mime types (nominally working!), but I ran into the below issue when trying to have two plots:

![image](https://user-images.githubusercontent.com/24885580/89247318-14b23980-d5db-11ea-9d24-3c74aac1553b.png)

I ran `@macroexpand` on the expressions and got:
```julia
:(dataset("cars") |> (VegaLite.VegaLite).VLSpec(VegaLite.convert_frag_tree_to_dict(VegaLite.fix_shortcut_level_spec((VegaLite.VLFrag)(VegaLite.Any[:point], VegaLite.OrderedDict{VegaLite.String, VegaLite.Any}("x" => :Horsepower, "y" => :Miles_per_Gallon))))))
```

I'm not sure why it's registering the x and y as variables in the main scope.  Any idea what's happening here?

fonsp:
Thanks!

Pluto works by figuring out the set of _assigned variables_ and _referenced variables_ for each cell. If one cell references a symbol `x`, and and another assigns to `x`, then that is a reactive link between them.

In this case, a bug in the analysis treats the keyword argument `x=:asdf` as an assignment. I wrote a special case for function calls, but not for macros.

A more general problem is that Pluto doesn't understand macros (#196), but luckily this macro can be parsed as if it is a function call - it does not need to be expanded for the syntax analysis to work.

fonsp:
Should be fixed now, can you try again?>

fonsp:
You need to run the master branch of Pluto

mcmcgrath13:
I'm using the master branch of Pluto, but still seeing the same issue:

![image](https://user-images.githubusercontent.com/24885580/91051678-710ee480-e5ee-11ea-9079-f44c65ad76eb.png)

fonsp:
This can't be fixed without expanding the macro (#196):
`@f(x=y)`
and
`@f x = y`
produce the same `Expr`, and something that looks like the second expression should be recognised as an assignment to `x`. (`@benchmark x = y` for example)

Try using a semicolon:
```julia
@vlplot(
	:a;
	x = :b,
	...
)
```

fonsp:
This has been fixed by #670

Issue #33, opened 2020-03-23
Check whether port is in use before starting server


fonsp:

fonsp:
I don't know how to do this.

On WSL, the Pluto server does not show up in `netstat`, I suppose that this is why I am able to open two Plutos on the same port.
I tried pinging the to be used IP address, but `HTTP.jl` behaves badly when the socket was not closed properly by the last Pluto. Also related to [this issue](https://github.com/JuliaWeb/HTTP.jl/issues/524).

fonsp:
Looks like this is only a problem on WSL

fonsp:
Fixed!

PR #253, opened 2020-08-04
wip: add support for vega mime types

mcmcgrath13:
This adds support to render vega plots interactively (current vega versions only).  

To do:
- [ ] Add tests (which file should they go in? new file?
- [ ] Figure out #252 
- [ ] how to tear down plots correctly

### DIFF
--- a/frontend/components/CellOutput.js
+++ b/frontend/components/CellOutput.js
@@ -70,6 +70,21 @@ const OutputBody = ({ mime, body, cell_id, all_completed_promise, requests }) =>
             return html`<div><${ErrorMessage} cell_id=${cell_id} requests=${requests} ...${JSON.parse(body)} /></div>`
             break
 
+        case "application/vnd.vegalite.v4+json":
+        case "application/vnd.vega.v5+json":
+            let vg = window.require("vega-lite@4", "vega@5", "vega-embed")
+                .then((res) => {
+                    res.embed(`#vega-${cell_id}`, JSON.parse(body))
+                })
+                .catch((err) => {
+                    // fall back to plain svg
+                    return html`<${OutputBody} mime="image/svg+xml" cell_id=${cell_id} body=${body} all_completed_promise=${all_completed_promise} requests=${requests} />`
+                })
+
+            return html`
+                <div id="vega-${cell_id}" style="width:100%;height:100%;"></div>
+            `
+            break
         case "text/plain":
         default:
             if (body) {
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -273,7 +273,6 @@ data:image/png;base64,ahsdf87hf278hwh7823hr...
 """
 function show_richest(io::IO, @nospecialize(x); onlyhtml::Bool=false)::MIME
     mime = Iterators.filter(m -> Base.invokelatest(showable, m, x), allmimes) |> first
-    println("mime: ", mime)
     t = typeof(x)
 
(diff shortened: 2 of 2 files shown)


frontend/components/CellOutput.js
```diff
@@ -70,6 +70,21 @@ const OutputBody = ({ mime, body, cell_id, all_completed_promise, requests }) =>
...
 
+        case "application/vnd.vegalite.v4+json":
+        case "application/vnd.vega.v5+json":
+            let vg = window.require("vega-lite@4", "vega@5", "vega-embed")
+                .then((res) => {
+                    res.embed(`#vega-${cell_id}`, JSON.parse(body))
+                })
+                .catch((err) => {
+                    // fall back to plain svg
+                    return html`<${OutputBody} mime="image/svg+xml" cell_id=${cell_id} body=${body} all_completed_promise=${all_completed_promise} requests=${requests} />`
```
fonsp, line 81: body is a JSON-formatted object, right? Can it render as svg?

fonsp:
Thanks so much for figuring this out! 

Would it be possible to include this in VegaLite itself? It would be a `MIME"text/html"` show method that outputs this HTML.

Pluto's JS runtime handles `<script>` tags, so it could be:
```jl
function Base.show(io::IO, m::MIME"text/html", v::VLSpec)
    divid = "vg" * randstring(3)
    print(io, """
<script src="https://cdn.jsdelivr.net/npm/vega@5"></script>
<script src="https://cdn.jsdelivr.net/npm/vega-lite@4"></script>
<script src="https://cdn.jsdelivr.net/npm/vega-embed@6"></script>
<div id="$divid" style="width:100%;height:100%;"></div>
<script>
var spec = """)
our_json_print(io, v)
print(io,"""
      var opt = {
        mode: "vega-lite",
        renderer: "$(Vega.RENDERER)",
        actions: $(Vega.ACTIONSLINKS)
      }
      vg_embed("#$divid", spec, opt);
</script>
""")
end
```

frontend/components/CellOutput.js
mcmcgrath13: Right now it renders as an svg from the show (so a vega plot in general can), but I do need to test that this piece of code actually works (it might not yet).

fonsp:
Maybe there is a reason why VegaLite does not want to have a `text/html` show method defined? Maybe because other HTML renderers don't support loading `<script>` tags from `src`?

In that case (there are probably more similar cases), we could add a mime type `application/vnd.pluto.html+xml` to Pluto, which has precedence over `text/html`, but is handled exactly the same. That way packages could only display to HTML if running in Pluto, or display different HTML for Pluto. (If absolutely necessary!)

fonsp:
This also prevents the situation where `VegaLite.jl` wants to update to `vega@6` or `vega-lite@5`. Then support for either the old or new version would have to be dropped if these versions are baked into Pluto.

mcmcgrath13:
Makes sense to me! Some of the other html consumers of VegaLite.jl don't seem to like the script tags, so I'm guessing that's why this isn't set up.  I can start a PR over there as well and see what David prefers.

mcmcgrath13:
https://github.com/queryverse/VegaLite.jl/pull/324

mcmcgrath13:
This problem will be solved by [VegaLite.jl #324](https://github.com/queryverse/VegaLite.jl/pull/324) and [Vega.jl #33](https://github.com/queryverse/Vega.jl/pull/33)

PR #254, opened 2020-08-05
Update dev_instructions.md

karlwessel:
There where some bits from the previous version of the document left. I guess this is how it should look like.


fonsp:
Fixing things only by _deleting_ lines is the best

Issue #122, opened 2020-05-25
Collapsible typeparameters


karlwessel:
For long stacktraces like they happen for example for 

```[1,2]*[1,2]```

it would be nice if the type parameters where displayed as some collapsed `SomeType{>...}` where you could show or hide the complete parameter list by clicking the arrow.

For example implementing
```
function Base.show(io::IO, ::MIME"text/html", x::DataType)
        show(io, x.name)
	if length(x.parameters) > 0
		print(io, "{")
		print(io, """<jltree class="collapsed" onclick="onjltreeclick(this, event)">""")
    	        print(io, "<jlarray>")
    	        Pluto.PlutoRunner.show_array_row.([io], enumerate(x.parameters))
    
    	        print(io, "</jlarray>")
    	        print(io, "</jltree>")
		print(io, "}")
	end
end
```
already displays types that are the result of a cell almost like I imagine. However:
1. it does not work for stacktraces and
2. the parameter list is not really collapsible but is always displayed in full, just the format is changed by clicking the arrow.

If you can give me some hints on how to solve these two issues I could try to implement some PR for this.

**Comment by fonsp** [comment]
For reference, [this](https://user-images.githubusercontent.com/6933510/82835048-2da7bb80-9ec3-11ea-9e84-ad8f4d94db82.png) is what the error message currently looks like.

**Comment by fonsp** [comment]
Wow that's a cool suggestion!

Copy pasting your code into a notebook does not change the error message for me. Where do you define the custom `show` method?

1. Could you also give an example that shows a useless stacktrace? I am surprised that it does not work, that would mean that `show` is not used by Julia to convert the datatype to a string in the stack frame `show` process.
2. I'm not able to verify, but it sounds like it is _wrapping_ the list into multiple lines, right? You could either disable wrapping, or hide all elements when the list is collapsed. This is done using CSS; you need to change something to the DOM generation (the code you wrote) to be able to differentiate this case from normal arrays - maybe add a class `parameter-list` to `<jltree>`.

You can try out new CSS rules inside your notebook:
```
html"""
<style>
jltree.parameter-list {
    color: purple;
}
</style>
"""
```

or add them to `treeviewer.css` and refresh.

I think that the nicest solution would be to give a custom `showerror` method for `MethodError` so that it also uses the custom formatting of stack traces. This also requires some work on the front-end - right now it does the formatted display only for the _stack trace_. I could give you some more info about this if you like!

[...]

PR #255, opened 2020-08-05
Disable automatic word wrap for MethodErrors

karlwessel:
This makes the method candidates displayed in `MethodError`s more readable by displaying each of them in one line that can be scrolled horizontally.

See also #122 
Before:
before

And after:
after


fonsp:
Fantastic!

PR #260, opened 2020-08-06
fix broken svg display

jveres:
The last update broke the svg display in cells. This PR fixes it and still uses the image tag to be able to right click save.

### DIFF
--- a/frontend/components/CellOutput.js
+++ b/frontend/components/CellOutput.js
@@ -61,7 +61,7 @@ const OutputBody = ({ mime, body, cell_id, all_completed_promise, requests }) =>
         case "image/bmp":
         case "image/svg+xml":
             console.log(body)
-            return html`<div><img src=${body} /></div>`
+            return html`<div><img src=${body.replace("%0A", "")} /></div>`
             break
         case "text/html":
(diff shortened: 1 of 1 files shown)


fonsp:
Could you post an SVG file that's broken?

fonsp:
It looks like this breaks all other image MIMEs by giving them all the svg MIME and base64 encoding already-encoded data?

fonsp:
I realise that I never really tested this new function, but the idea was that svg image data does not need to be base64-encoded.

jveres:
e.g. `using GR; inline("svg"); contourf(peaks())`

fonsp:
Thanks, I'm having a look now

jveres:
> It looks like this breaks all other image MIMEs by giving them all the svg MIME and base64 encoding already-encoded data?

Right, I totally overlooked it. Fixed in https://github.com/fonsp/Pluto.jl/pull/261

jveres:
> I realise that I never really tested this new function, but the idea was that svg image data does not need to be base64-encoded. I'll experiment with it again.

Correct, can be with or without base64. The root cause was the missing encoding.

PR #283, opened 2020-08-13
Reduce #title's margin-top

zdroid:
Before:
Annotation 2020-08-13 154713

After:
Annotation 2020-08-13 154642

With a smaller margin the scrollbar disappears (which looks better IMO). Screenshots from 1920x1080 screen.

### DIFF
--- a/frontend/index.css
+++ b/frontend/index.css
@@ -4,7 +4,7 @@
     /* position: absolute; */
     /* left: 0; */
     /* right: 0; */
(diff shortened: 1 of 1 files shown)


fonsp:
I like it in the middle :)

It looks very nice when you open Pluto for the first time (no recent notebooks). And when there are a lot of recent notebooks, it doesn't clutter the screen by showing all of them.

Issue #286, opened 2020-08-13
Safari no longer supported after Pluto v0.11.4?


goedman:
Hi,

Going over the issues threads this morning I decided to update Pluto, first to v0.11.5 and later to v0.11.6. Neither seems to work on Safari.

The start-up screen appears fine (on port 1234, a second session will take port 1135), but after selecting to open a new notebook, the top right of the screen shows `Loading` and `Disconnected` repeatedly. Nothing is output to the Terminal window and no cells show up.

It does work on Chrome (but that is not a good option for me right now).

Let me know if I can help out with testing.

Rob

fonsp:
Safari is not an officially supported browser because I can't test on it. But if someone submits a minor PR to add support, I will add it in!

A screenshot of your browser's developer console would be helpful for anyone who wants to fix it

PR #289, opened 2020-08-14
Fixes Safari support (#286)

oem:
This PR fixes the safari support: https://github.com/fonsp/Pluto.jl/issues/286

## Problem

The reason for pluto failing in safari is `frontend/common/PlutoConnection.js:104`, specifically the usage of `Blob.arrayBuffer()`.
The support for this is still [pretty spotty](https://developer.mozilla.org/en-US/docs/Web/API/Blob/arrayBuffer).

Screenshot 2020-08-14 at 13 56 54

## Solution

Instead uses `Response(blob).arrayBuffer()`, whose support is [much better](https://developer.mozilla.org/en-US/docs/Web/API/Body/arrayBuffer).

### DIFF
--- a/frontend/common/PlutoConnection.js
+++ b/frontend/common/PlutoConnection.js
@@ -101,7 +101,7 @@ export class PlutoConnection {
 
     async handle_message(event) {
         try {
-            const buffer = await event.data.arrayBuffer()
+            const buffer = await new Response(event.data).arrayBuffer()
             const buffer_sliced = buffer.slice(0, buffer.byteLength - this.MSG_DELIM.length)
             const update = unpack(new Uint8Array(buffer_sliced))
             const by_me = "initiator_id" in update && update.initiator_id == this.client_id


fonsp:
Awesome, thanks!

fonsp:
Wait we have a polyfill for this, why didn't that work?

https://github.com/fonsp/Pluto.jl/blob/master/frontend/common/Polyfill.js#L1-L20

fonsp:
Aha because the polyfill is loaded afterwards:

https://github.com/fonsp/Pluto.jl/blob/master/frontend/index.js

https://github.com/fonsp/Pluto.jl/blob/master/frontend/index.js

fonsp:
It seems like your polyfill is better because it avoids allocating the data twice?

oem:
I must admit, I did not see the polyfill for `arrayBuffer`, just for `text` (err, I mean I of course tried to optimize an existing solution that I gleamed from careful reading of the source code! ;) )

fonsp:
Could you check if this works?

https://github.com/fonsp/Pluto.jl/tree/polyfill-updates

goedman:
Pluto v0.11.6+ works again on MacOS/Safari! Thanks.

PR #310, opened 2020-08-20
Allow Pluto.run(host), i.e. without specifying port.

fredrikekre:


src/webserver/WebServer.jl
```diff
@@ -35,7 +35,7 @@ The default `host` is `"127.0.0.1"`. For wild setups like Docker and heroku, you
 
 This will start the static HTTP server and a WebSocket server. The server runs _synchronously_ (i.e. blocking call) on `http://[host]:[port]/`. Pluto notebooks can be started from the main menu in the web browser.
 """
-function run(host, port::Union{Missing,Integer}; launchbrowser::Bool=false, session=ServerSession())
+function run(host, port::Union{Missing,Integer}=missing; launchbrowser::Bool=false, session=ServerSession())
```
fredrikekre, line 38: Just a drive-by-comment: I think using `nothing` for "not-specified" is more "Julian".

fonsp: Thanks! I'll add that in. I think I should do that in some other places in the code as well

PR #318, opened 2020-08-25
Initialize global IOContext instance with proper displaysize

### LINKED ISSUE #292: Display DataFrame with more than 18 rows/columns
For larger DataFrames, are the first 18 rows and columns are displayed in Pluto.
Is there a way to increase this number?
I already tried the "standard methods" suggested by DataFrames.jl: setting the COLUMNS and LINES environment variables or using `show`, but without success.

```
using DataFrames

ENV["COLUMNS"]=1000
ENV["LINES"] = 50
# method that works in REPL / IJulia,  but seems to have no effect in Pluto
# https://juliadata.github.io/DataFrames.jl/stable/man/getting_started/#Installation

df = hcat([DataFrame(a=1:30, b=2:31) for _ in 1:10]..., makeunique=true)
# only 18 rows and 18 columns are displayed (the df size is 30x20)

show(df, allrows=true, allcols=true) # prints on the server side, but not in the notebook
```

Could you please help?

Sh4pe:
Make output respect the environment variables `LINES` and `COLUMNS`. 

Fixes #292

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -195,7 +195,11 @@ function html(io::IO, x::LaTeX)
 end
 
 "The `IOContext` used for converting arbitrary objects to pretty strings."
-iocontext = IOContext(stdout, :color => false, :compact => false, :limit => true, :displaysize => displaysize())
+iocontext = let
+    width = get(ENV, "COLUMNS", 88)
+    height = get(ENV, "LINES", 18)
+    IOContext(stdout, :color => false, :compact => false, :limit => true, :displaysize => (height, width))
+end
 iocontext_compact = IOContext(iocontext, :compact => true)
 
 const imagemimes = [MIME"image/svg+xml"(), MIME"image/png"(), MIME"image/jpg"(), MIME"image/jpeg"(), MIME"image/bmp"(), MIME"image/gif"()]
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -196,8 +196,19 @@ end
 
 "The `IOContext` used for converting arbitrary objects to pretty strings."
 iocontext = let
-    width = get(ENV, "COLUMNS", 88)
-    height = get(ENV, "LINES", 18)
+
+    width = try
+        parse(Int, get(ENV, "COLUMNS", "88"))
+    catch
+        88
+    end
+
+    height = try
+        parse(Int, get(ENV, "LINES", "18"))
+    catch
+        18
+    end
+
     IOContext(stdout, :color => false, :compact => false, :limit => true, :displaysize => (height, width))
 end
 iocontext_compact = IOContext(iocontext, :compact => true)
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -196,20 +196,8 @@ end
 
 "The `IOContext` used for converting arbitrary objects to pretty strings."
 iocontext = let
-
-    width = try
-        parse(Int, get(ENV, "COLUMNS", "88"))
-    catch
-        88
-    end
-
-    height = try
-        parse(Int, get(ENV, "LINES", "18"))
-    catch
-        18
-    end
-
-    IOContext(stdout, :color => false, :compact => false, :limit => true, :displaysize => (height, width))
+    displaysz = (parse(Int, get(ENV, "LINES", "18")), parse(Int, get(ENV, "COLUMNS", "88")))::Tuple{Int, Int}
+    IOContext(stdout, :color => false, :compact => false, :limit => true, :displaysize => displaysz)
 end
 iocontext_compact = IOContext(iocontext, :compact => true)
 
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -195,7 +195,7 @@ function html(io::IO, x::LaTeX)
 end
 
 "The `IOContext` used for converting arbitrary objects to pretty strings."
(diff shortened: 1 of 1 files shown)


Sh4pe:
One drawback is that the environment variables have to be set before `Pluto` is imported since the global `IOContext` instance is statically initialized [here](https://github.com/fonsp/Pluto.jl/blob/868a48b28b2989c82d9b85188d5ca9a0aff8e846/src/runner/PlutoRunner.jl#L198).

Can this be improved?

fonsp:
Aha, thanks for figuring out what those environment variables do.

But this parameter is not just there for table display - the display size `(18,120)` is the size of the cell output pane in Pluto's UI, measured in monospace characters. This is used by Julia to determine how much should be printed when displaying a 2D array, for example.

The value `120` has been wrong for a while (we use a larger font), and by coincidence, I was about to fix this but hadn't committed yet. It should be `88`:

image

fonsp:
Ah sorry I had a look at the docstring for `displaysize`.

It looks like this is fine, except we should manually use `(18,88)` as default, and update it before every format

Sh4pe:
Ok, I see. What do you think about 0c7ecbe547aa547509e2b14513709bbbf4e4087c?

fonsp:
`parse(Int,get(ENV,"LINES","18"))`

Sh4pe:
Good point. Added `try` ... `catch` ... `end` around it in 760536fc908fd8012a6adc1239fc0803d2bf94d7 as well.

fonsp:
About the try catch block, can you check the source code for `Base.displaysize()` and copy what they do?

fonsp:
Maybe you actually want to see an error instead of silently getting the default, not sure

Sh4pe:
Yes, thats probably less confusing. And this more similar to the behaviour of [Base.displaysize()](https://github.com/JuliaLang/julia/blob/v1.5.0/base/stream.jl#L454-L455) which does not catch the parse error as well.

Sh4pe:
Is this version ok, @fonsp?

fonsp:
This definition for `displaysize` is right, but now you need to set the ENVs before launching Pluto, right?

Sh4pe:
Yes, the value of the ENVs at the time [iocontext is defined](https://github.com/fonsp/Pluto.jl/blob/f7ec03dbac8d46208cba6e600fd6d513f82b71c3/src/runner/PlutoRunner.jl#L198) decides the size. This is something I'm not sure how to fix. Would it work to generate a new `IOContext` instance every time it is needed? Recreating it would allow for using e.g. `withenv(func, "COLUMNS"=>300)` or something like this.

fonsp:
Yes, it's already being recreated every batch of runs, https://github.com/fonsp/Pluto.jl/blob/master/src/runner/PlutoRunner.jl#L32

Maybe you could `@benchmark` how long it takes to recreate one?

Sh4pe:
I've tried to change the displaysize in `set_current_module` and tested it with [this](https://gist.github.com/Sh4pe/3b6c04f59f1e56b841cf1b26bb7ae8d5) notebook. Upon first execution, the DataFrame is displayed with 18 lines (which I don't expect). When I re-execute, the DataFrame is correctly printed with 20 lines.

It appears that the "first batch" does not yet know that one of its cells sets `ENV["LINES"]` and thus the display size is not correctly set in the second cell of my test notebook. 

It's getting complicated... :/ Is there another point where I could insert the change of the display size?

fonsp:
Yes it could just be inserted in the `show_richest` function but that would be bad for performance, since it gets called for every element of an array, for example

Sh4pe:
Ok. Then I'd suggest to stick with the initial implementation that reads the width and height from environment during initialization once. Then, we could not change the width/height on a "cell by cell basis".

I'm working extensively with DataFrames and I really like Pluto. But only seeing 18 rows of a DataFrame and very few columns makes it very hard to use Pluto for me.

What do you think, @fonsp?

fonsp:
Yep, it's an important problem that we need to solve. 

We should consider opening a PR a DataFrames - we are limited by the fact that they use the `LINES` and `COLUMNS` environment variables, which are meant to described the _terminal size_, to determine how many rows and columns to print in the _HTML output_. It should be possible to control these independently.

Sh4pe:
Hi @fonsp, sorry for not responding for quite some time. Meanwhile, I've come up with a viable workaround that works well enough for me. I won't have the time to work on this particular issue/PR for the forseeable future. But since I found a workaround, it would be fine for me to just close the PR.

dralletje:
@Sh4pe is your workaround something you could post in raw form? I'm fine putting in the time in the PR, but not having to do all the research would save me lots of time :)

Sh4pe:
My workaround is to convert the `DataFrame` to an `HTMLTable` from [BrowseTables.jl](https://github.com/tpapp/BrowseTables.jl). And if this table is too wide, I'll wrap it in a `<div>` element with vertical and horizontal scrollbars.

For wrapping the `<div>`, I use this function:

```julia
function wrap_in_scroll_div(table::HTMLTable; height="300px", title=nothing)
        buf = IOBuffer()
        BrowseTables.write_html_table(buf, table.table)
        html = buf |> take! |> String
        str_title = ifelse(isnothing(title), "", "<h4>$title</h4>")
        HTML("""
        $str_title
        <div style="height: $height; overflow-y: scroll;">
                $html
        </div>
        """)
end
```

fonsp:
After some fresh air I came up with this:

You can have an "IO Context properties wrapper", which does not affect the display of subsequent code

https://github.com/fonsp/disorganised-mess/blob/master/WithIOContext.jl

image

fonsp:
image

PR #349, opened 2020-09-04
💈 Cell queued state

shashankp:


fonsp:
Can you make it a draft PR?

fonsp:
Looks awesome by the way! Happy that you figured out how to get that information from Julia to display.

fonsp:
Have you experimented with using different animation speeds instead of colors?

shashankp:
> Have you experimented with using different animation speeds instead of colors?

not sure if its making the difference obvious

animation

fonsp:
I feel like this is much more clear than only showing a difference in colors. It's also because that particular choice of green and grey have the same luminosity, but in general i feel like giving it a color is semantically wrong. Green is "good", but that does not apply here.

Maybe we can do [fast]+[slightly dark gray] for the running cell, and [very slow]+[slightly lighter gray] for the rest?

shashankp:
> I feel like this is much more clear than only showing a difference in colors. It's also because that particular choice of green and grey have the same luminosity, but in general i feel like giving it a color is semantically wrong. Green is "good", but that does not apply here.
> 
> Maybe we can do [fast]+[slightly dark gray] for the running cell, and [very slow]+[slightly lighter gray] for the rest?

agree.. better?

grayed

fonsp:
Woa that gave me chills, very nice

shashankp:
Will browse the issues & cherry pick, there are quite a few.
When I get stuck at figuring out the core pieces, we could have a video call

welcome & thank you for the neat project 
already has visibility from the MIT course; needs to keep the momentum to make it great!

fonsp:
Great! Happy to help.

Tips for outreach/momentum are also welcome! Don't know much about this myself

Issue #344, opened 2020-09-03
Move FAQ back to GitHub Wiki


fonsp:
Right now it's here:
https://www.notion.so/3ce1c1cff62f4f97815891cdaa3daa7d?v=b5824fb6bc804d2c90d34c4d49a1c295

But this doesn't allow outside contributors. And SEO seems to be bad

fonsp:
If anyone wants to be a hero: we just need to revert the latest commits on the Wiki git of this repository - the FAQ used to be there

There were some small changes to the notion FAQ but it's fine if they are lost

shashankp:
@fonsp not being a hero here, but I dont think wiki repo is setup for contributions. I've tried to push [updated wiki home page](https://gist.github.com/shashankp/2ef4b4e79d83ea9d55f296157be69736) but got 403

fonsp:
Oof do you think there's a way to fix that?

shashankp:
@fonsp https://docs.github.com/en/github/building-a-strong-community/changing-access-permissions-for-wikis

fonsp:
Awesome I did it! Can you try again?

fonsp:
And can you make 4 subpages?

shashankp:
Done. 
Also, [Dev instructions](https://github.com/fonsp/Pluto.jl/blob/master/dev_instructions.md) will feel at home in the wiki.

fonsp:
Haha how did you do without being a co-owner of the repo?

fonsp:
I agree, but it's common for repositories to have a CONTRIBUTING.md:
https://github.com/JuliaLang/julia/blob/master/CONTRIBUTING.md

So let's actually rename it to that

shashankp:
> Haha how did you do without being a co-owner of the repo?

oops, maybe you would want to revert [permission](https://docs.github.com/en/github/building-a-strong-community/changing-access-permissions-for-wikis) once this issue is closed

fonsp:
Nah let's leave it and see what happens :)

fonsp:
Anyway, thanks a lot! 🥇

Feel free to write a PR for that rename since you brought it up

PR #353, opened 2020-09-04
rename contributing file

shashankp:
https://github.com/fonsp/Pluto.jl/issues/344#issuecomment-687061703


fonsp:
Do you feel like anything is missing in this guide?

shashankp:
I guess there will be devs new to Julia itself (like me), linking to an article which shows how to find help when needed (like using the ?) etc. would be great

PR #357, opened 2020-09-04
Change the 5-cut hint-show condition (add 1) (closes #356)

### LINKED ISSUE #356: sample notebooks: pizza 5 cuts typo?
This part near the end in sample notebook 2:

> To get an extra hint, figure out how many slices we can get from 5 cuts:

With 5 cuts we can get 16 slices, but in order to show the hint we need to put 15. 

The language of the hint itself noted that there should be one additional slice corresponding to the pizza with no cuts, but from wording of the question I am not sure if it is clear that we were supposed to subtract 1. In any case, it seems not consistent with the previous questions about smaller numbers of cuts and the last part with the slider, which agrees that 16 is the number of slices for 5 cuts.

zmoon92:
I don't think it is necessary to change the language in the hint, which acknowledges the extra 1 that should have been added.


fonsp:
Thanks a lot!

Issue #296, opened 2020-08-15
Use JuliaMono font 🔤


fonsp:
Thanks @cormullion for helping me out!

Read this discussion about loading the font:
https://github.com/cormullion/juliamono/issues/24#issuecomment-674382769

Pocket-titan:
Just my two cents: you might find the [css property `unicode-range`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range) useful, to be used inside the `@font-face` rule. This way, you could define two `@font-face` rules, one for the latin font & one for Latin + the other glyps (and have them use the same`font-family`identifier), where the former would have a `unicode-range` spanning the Latin alphabet & the latter spanning Latin + the rest.

PR #364, opened 2020-09-05
🔠 Replace Roboto Mono with JuliaMono

novoselrok:
Fixes: https://github.com/fonsp/Pluto.jl/issues/296

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -151,7 +151,7 @@ a {
 }
 
 pluto-output code {
-    font-family: JuliaMono, JuliaMono-Latin, monospace;
+    font-family: JuliaMono-Latin, JuliaMono, monospace;
     font-size: 0.75em;
 }
 
@@ -162,7 +162,7 @@ pluto-output pre {
     word-break: break-all;
     tab-size: 4;
     -moz-tab-size: 4; /* https://bugzilla.mozilla.org/show_bug.cgi?id=737785 */
-    font-family: JuliaMono, JuliaMono-Latin, monospace;
+    font-family: JuliaMono-Latin, JuliaMono, monospace;
 }
 
 pluto-output hr {
@@ -262,7 +262,7 @@ pluto-output div.footnote p.footnote-title::after {
 }
 pluto-output a.footnote,
 pluto-output div.footnote p.footnote-title {
-    font-family: JuliaMono, JuliaMono-Latin, monospace;
+    font-family: JuliaMono-Latin, JuliaMono, monospace;
     font-size: 0.75rem;
     font-weight: bold;
     margin-block-end: 0px;
@@ -297,7 +297,7 @@ pluto-output table > thead {
 
 pluto-output table > tbody td {
     font-size: 0.75rem;
-    font-family: JuliaMono, JuliaMono-Latin, monospace;
+    font-family: JuliaMono-Latin, JuliaMono, monospace;
 }
 
 pluto-output table td,
@@ -376,7 +376,7 @@ div.export_title {
     transform: rotate(180deg);
 
     margin-top: 20px;
-    font-family: JuliaMono, JuliaMono-Latin, monospace;
+    font-family: JuliaMono-Latin, JuliaMono, monospace;
 }
 a.export_card header {
     margin-block: 0px;
@@ -384,7 +384,7 @@ a.export_card header {
 }
 a.export_card section {
     font-size: 0.75rem;
-    font-family: JuliaMono, JuliaMono-Latin, monospace;
+    font-family: JuliaMono-Latin, JuliaMono, monospace;
     color: rgba(0, 0, 0, 0.5);
     font-weight: 500;
 }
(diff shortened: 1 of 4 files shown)


fonsp:
We should measure the performance impact.

Can you take measurements before and after the change using Chrome devtools and post the screenshots here?

The two different times are:
1. Load time on first load (with cache disabled - check the Network log)
2. Load time on second load, with the service worker cache enabled (default)

So 4 timings in total

novoselrok:
Measurements on the sample Tower of Hanoi notebook.

**Summary**

|                 |           | Load (ms) | DomContentLoaded (ms) |
|-----------------|-----------|-----------|-----------------------|
| Cache disabled  | Master    | 726       | 316                   |
|                 | JuliaMono | 1200      | 322                   |
| Cache/SW enabled   | Master    | 376       | 360                   |
|                 | JuliaMono | 391       | 362                   |

**Screenshots**

Master - cache disabled:

juliamono-cache-disabled

Master - cache enabled:

juliamono-cache-enabled

JuliaMono - cache disabled:

juliamono-cache-disabled

JuliaMono - cache enabled:

juliamono-cache-enabled

fonsp:
Oof that's a lot. 

But actually, maybe `load` is measured until all fonts are loaded? In that case it's what we expected, and why we deliberately moved the latin subset out of the rest. The goal is that there will be a long load time, but the "time-to-first-juliamono-latin" is roughly the same as the current "time-to-first-robotomono". 

What do you think?

novoselrok:
I think you are correct, `load` is until all fonts are downloaded. Additionally, I had to swap the order of Latin and regular fonts when specifying the `font-family`, so that Latin font gets downloaded first and then swapped if there are missing characters.

Now the "time-to-first-juliamono-latin" is roughly the same as the "time-to-first-robotomono".

fonsp:
Shouldn't it be a single font-face, with the latin subset defined only on a unicode subset?

Like @Pocket-titan suggested in https://github.com/fonsp/Pluto.jl/issues/296#issuecomment-675136703

Or does a single font-face need to be loaded entirely before it is applied to the page?

You can use this to simulate a super slow response for the non-latin font files:
http://slowwly.robertomurray.co.uk/

fonsp:
Another thing - I also use the Tower of Hanoi sample for my performance tests, but you need to test it on a page that contains non-latin characters inside code.

novoselrok:
It's not a problem, this is a really fun project and I hope I can contribute some more in the future :slightly_smiling_face: 

You are correct regarding the unicode range. I was able to merge the two font families into one and everything is working the same.

For testing: I'm using a slightly modified version of the Hanoi notebook with a couple of non-latin cells. And I'm using the Chrome DevTools network throttling to test the font loading on slower connections.

fonsp:
Great! So what are the test results for loading just the latin set?

novoselrok:
I re-ran the tests on master as well. As far as I can measure, the JuliaMono Latin-only times are essentially the same to the Roboto Mono loading times.

|                 |           | Load (ms) | DomContentLoaded (ms) |
|-----------------|-----------|-----------|-----------------------|
| Cache disabled  | Master    | 733       | 319                   |
|                 | JuliaMono (Latin only) | 784      | 332                   |
| Cache/SW enabled   | Master    | 386       | 330                   |
|                 | JuliaMono (Latin only) | 393       | 375                   |

fonsp:
🎉

fonsp:
@cormullion

fonsp:
Nice! I'm still not 100% satisfied with how the UI (top and bottom) looks with the new font, but I'll come back to it later

cormullion:
wow thanks a lot everyone! I’m very honoured to play a small part in Pluto’s early development! Definitely makes it worthwhile. 

You’re linking to a specific version (v0.013) for safety, I suppose? Currently I appear to be making one release a week, which obviously won’t last... v0.015 will have BoldLatin perhaps 😂

fonsp:
Okay! I will subscribe to your releases and update it here - I also release once a week :)

novoselrok:
@fonsp Do you want me to update anything else, or will you just finish up the UI tweaks?

fonsp:
Let's use Roboto Mono for UI (header+footer) and JuliaMono for the codemirrors

fonsp:
💕

fonsp:
This is awesome!

image

novoselrok:
🎉🔥

cormullion:
UnicodePlots is great - try `heatmap()` as well!

jtschneider:
Quick note about Roboto: The glyph for x and χ are the same in Roboto, which can be seen [here](https://fonts.google.com/specimen/Roboto+Mono?preview.text=%CF%87x&preview.text_type=custom&category=Monospace). This costed me some annoyance since I had code with x and χ , see also https://github.com/fonsp/Pluto.jl/issues/1071 .
I would therefore lobby for changing the font towards a font that has actually different glyphs for different (and often used) unicode characters, for example Fira Mono.

fonsp:
@jtschneider Do not cross-post to get more attention

KronosTheLate:
> UnicodePlots is great - try `heatmap()` as well!

That is what brought me here - an issue with heatmap unicode plots. The output in VSCode:
image

The output in Pluto:
image

Should I open a seperate issue for it, or is it enough to ping the relevant people by this comment?

cormullion:
This is a duplicate of https://github.com/fonsp/Pluto.jl/issues/894, I think.

KronosTheLate:
Looks like. Thanks for the quick reply - It looks like the issue is being properly handled over there.

Issue #342, opened 2020-09-03
do not change pwd after exiting Pluto


Roger-luo:
I find currently Pluto is using cd directly instead of use it as a context.

https://github.com/fonsp/Pluto.jl/blob/master/src/evaluation/WorkspaceManager.jl#L62

this changes the results of `pwd()` and can cause problems when someone tries to start pluto inside a script (e.g managing pluto instances)

fonsp:
It only `cd`s within the notebook process, right?

Roger-luo:
for this line, I think yes. But I think this somehow affects the main process, when you 

1. execute `Pluto.run`
2. then open a new notebook (don't save it)
3. then Ctrl+C in terminal
4. pwd() will be moved to the tempdir

![Screen Shot 2020-09-03 at 12 36 23 PM](https://user-images.githubusercontent.com/8445510/92142606-1a6d8b80-ede2-11ea-9ab2-0ea73abb76d7.png)

Roger-luo:
this should be fixed in #367

Issue #366, opened 2020-09-05
A better config system


fonsp:
@Roger-luo

Right now we have some `ENV` variables with a custom function to get default values - this was quickly put together for some flags in the automated testing, but I'd like to have something better.

First some use cases: (feel free to suggest more)
- all the current switches to server behavior: https://github.com/fonsp/Pluto.jl/blob/v0.11.12/src/Pluto.jl#L21-L29
- port, hostname and URL
- security configuration (tokens, cors, etc)
- launch with a couple of notebooks already running
- for binder: launch with a predetermined set of notebooks in the main menu, and no further options
- for binder: launch straight to the main menu
- custom themes, custom keyboard shortcuts
- 

Some requirements to make this easy:
- We want more than just strings
- We should use Julia code to set the config and nothing else. 
- I like that you can do 
```julia
with_env("PLUTO_WORKSPACE_USE_DISTRIBUTED" => "false") do
    Pluto.run(1239)
end
```

Ideas:
1. lots of keyword arguments to `Pluto.run`
2. the `ServerSession` will contain all of this. A config file is just a `.jl` script that defines a `ServerSession`.

Roger-luo:
I think we can consider use the same method I used in Comonicon's `Comonicon.BuildTools.install` for `Pluto.run` and `Pluto.open` (see also https://github.com/Roger-luo/Comonicon.jl/blob/master/src/tools/build.jl#L71)

1. the entry function of `Pluto` forward a set kwargs to a config handling function, e.g 

```julia
function Pluto.run(host, port;kwargs...)
  toml_config = read_toml(some_pluto_config_file)
  # merge config file configs and configs specified
  # by users using kwargs
  config = merge_config(toml_config, kwargs)
  # validate configs, see if there are anything incorrect
  validate_config(config)
  # actual run with all configs
  Pluto.run(configs) # This is dispatched on Pluto.run(::Dict)
end
```

2. user can choose to specify things in kwargs, some env variables, or a config file, e.g `Pluto.toml`
3. the priority takes in kwargs > Pluto.toml > env variables > default

As for config file, I actually suggest to just use a TOML, since you can limit what your config file can do in a TOML (given all we want to set up are PATHs, execution options etc. rather than arbitrary Julia code) but not an arbitrary Julia script. And note Pkg ships with a TOML and it will be a stdlib in 1.6.

fonsp:
[...]

PR #367, opened 2020-09-05
⚙ new config system without environment variables

Roger-luo:
This PR makes most of the path and env relocatable by runtime evaluating them (instead of compile time) and enable all Pluto environment variables to have a prefix `JULIA_` or not. (note this needs #341 to be merged first)


src/evaluation/WorkspaceManager.jl
```diff
@@ -82,7 +82,7 @@ function create_emptyworkspacemodule(pid::Integer)::Symbol
 end
 
 function default_environment_path(::Nothing)
-    get_pl_env("PLUTO_DEFAULT_ENVIRONMENT_PATH")
```
Roger-luo, line 85: @fonsp sorry I didn't bring this up in last PR. I'm wondering what's the reason to change `PLUTO_PROJECT` to `PLUTO_DEFAULT_ENVIRONMENT_PATH`? I think at least at CLI side, this variable is exactly the same as `JULIA_PROJECT`: it can be used to interact with some environment manager tool (e.g `direnv`) to control the environment variable, but when `pluto --project` is used, it gets overwrite by `--project`. I find `PLUTO_PROJECT` is more consistent with Pkg, which is easier to remember, what do you think?

Roger-luo:
There is a few things I will update if @fonsp agrees with that:

1. allow all environment variables to have a `JULIA_` prefix version. I think this can make things more explicit
2. trying to remove paths declared as `const`, this will prevent people from building distributable system image for Pluto

src/evaluation/WorkspaceManager.jl
```diff
@@ -97,6 +97,9 @@ function resolved_environment_path(notebook::Notebook)
     else
         if isabspath(notebook.environment_path)
             tamepath(notebook.environment_path)
+        elseif notebook.environment_path == "@."
```
Roger-luo, line 100: this fix the issue that when notebooks are specified with `@.` the project is using `notebook_dir/@.` path.

src/evaluation/WorkspaceManager.jl
Roger-luo: I think this name is short enough to be explicit. If you want to make the environment variable name actually matches the field name in the `struct`s, I think renaming the internal field names is probably better.

Roger-luo:
there a few other things I just realize that the workspace process is currently spawned by inheriting all main process flags - which is something we don't want. e.g if I start the main process in CLI with `compile=min`, the notebook process will also start with that. I guess it's OK to start a Pluto server instance using `compile=min` given there is little to compile anyway, this can reduce the start-up time of Pluto itself too. On the other hand, I think the notebook workspace shouldn't use the default startup file unless specified, since things like `OhMyREPL` etc. are not for notebooks and can result in a long notebook start-up time, so we probably want to have `--startup-file=no` too, unless it is specified by user explicitly.

I'm thinking about implementing a kwargs interface in `run` and `open` instead of this `PLUTO_PROJECT` environment variable, and let Pluto CLI handle these compiler option related environment variables, so we don't end up in too much environment variables here. What do you think? @fonsp

fonsp:
Thanks! I'll write a comment here: #366

Roger-luo:
I haven't make this branch runs, but maybe you want to have a look at the implementation I come up with for #366

Roger-luo:
to be short, I basically make all the configs go into a `Session` and share them across the internals. Now workspace actually takes a set of full compiler options instead of just project, so we can make it not inheriting the main process compiler flags.

For interface level, I haven't implement the `kwargs` interface, but user only needs to learn what is a `Session` and how to config it now.

src/webserver/Session.jl
```diff
@@ -1,5 +1,24 @@
...
+    launch_browser::Bool = true
+    single_notebook_mode::Union{Nothing,Notebook} = nothing
+    show_file_system::Bool = true
+end
+
+Base.@kwdef struct PlutoConfiguration <: AbstractPlutoConfiguration
+    working_directory::String = default_working_directory()
+    run_notebook_on_load::Bool = true
+    workspace_use_distributed::Bool = true
+    require_token_for_open_links::Bool = true
```
Roger-luo, line 16: put security struct back.

fonsp:
@Roger-luo what do you think?

Roger-luo:
perfect, now let me add:

- [ ] the kwargs interface for `run` and `open`
- [ ] add some tests as much as I can

then I think we are good to go!

Roger-luo:
I just added the kwargs interface for `run` @fonsp does this looks good to you? 

but somehow I cannot start a notebook session from this PR, something is broken... I need to check what's wrong.

fonsp:
What do you mean? Do you get an error?

Can you add a deprecation warning to the compat versions of Pluto.run?

fonsp:
Let's merge?

Roger-luo:
cool, it works now! I didn't realize there were other things that need to update. I tried to add some tests. I think it should be good to merge.

fonsp:
Can you check why the tests failed?

Roger-luo:
oops I wrote a wrong path in the test, let's see if it passes now.

Roger-luo:
@fonsp tests all passed now, guess it's good to merge.

Roger-luo:
OK I think this is because when we assume each project starts with `@.`: currently if we specify the project at `Pluto.run(;project="@.")`, this config will directly passed to the julia compiler, to let it decide what exactly `@.` means. 

However, the newly created notebooks are in a `tempdir` which does not have `Project.toml`, thus it fallbacks to the global, even the user probably wants the `Project.toml` in current directory. I think this is more a interface related issue, we might need to expand `@.` ourselves before we pass it to `addprocs`? or should we keep the current behaviour? @fonsp

Roger-luo:
umm, it seems before `create_workspace` is called, the `pwd()` is changed to the notebook folder already. Is this intended?

fonsp:
About `"@."`: you decide, I am too confused, and this will be different by `0.13` anyways. Until then, anything is fine as long as most people launch their notebook into the global env by default.

About `pwd`: I don't know what you mean, but the `pwd()` according to your notebook's code should always equal the place where the notebook is stored. Changing the path should `cd`.

Roger-luo:
I think I find out why, this is also the reason for #342 , basically, this is because Pluto will try to create a workspace in the server process to make UI loading faster. But it does not cd back when it finishes the job. This is causing

1. `pwd` is modified at server process, causing #342 
2. `@.` is expanded in the wrong path, since `pwd` is not the actual `pwd` of the server process anymore

src/evaluation/Run.jl
```diff
@@ -115,8 +116,12 @@ function update_save_run!(session::ServerSession, notebook::Notebook, cells::Arr
 	else
 		# this code block will run cells that only contain text offline, i.e. on the server process, before doing anything else
 		# this makes the notebook load a lot faster - the front-end does not have to wait for each output, and perform costly reflows whenever one updates
-		offline_workspace = WorkspaceManager.make_workspace(notebook, false)
-		
+		# "A Workspace on the main process, used to prerender markdown before starting a notebook process for speedy UI."
+		offline_workspace = WorkspaceManager.make_workspace(
```
Roger-luo, line 120: this workspace is created in the server process, which is causing the following workspace's compiler options pwd become incorrect.

fonsp:
Aha! good catch

Roger-luo:
I fixed #342 with a workaround for now in this PR, but I think it's not fully resolved:

 if a user config the server process to not use `Distributed` but always run in the server process, the current behaviour will effect the `pwd` between different workspace inside the server process. I think in principal what is preferred is to execute each cell using `cd(f, path)` when it's executing in the server process.

now, all the notebook environment behaviour, no matter it's new notebook or not should be the same. to summarize the current behaviour:

### Session Environment
1. `Pluto.run` set the environment to `@.` by default, which means:
  - if julia finds there is a `Project.toml` in parent directory, it will activate it
  - if julia finds no local `Project.toml`, it activates the global default environment (fallback to the current behaviour)
2. `Pluto.run` set by the user using relative path:
  - it will use `joinpath(pwd(), relative_path)`, which is the relative path to current `pwd`
3. `Pluto.run` set by the user using absolute path:
   - it will use the absolute path

### Notebook Environment
one can create a notebook with notebook environment, or open a single notebook via `SessionActions.open` with `compiler_options` that the `project` kwargs is set to desired project path. We always use notebook environment if it's not nothing.

PR #368, opened 2020-09-05
[Don't merge] set up formatter

Roger-luo:
This set up a formatter bot in github action, it will open PRs to format the codebase. Depends which workflow you prefer, there is also format checker in CI for each PR too (but whoever create the PR needs to run format locally first)


fonsp:
Could you run the formatter on a fork so that we see what it looks like?

PR #369, opened 2020-09-05
run JuliaFormmater

Roger-luo:
This runs JuliaFormatter using config

```julia
format("test", always_for_in=true, margin = 102)
```

Let me know if you want other code style.


Roger-luo:
ummm.... somehow the tests are broken by formatter I think it's something unexpected. I think I'll report this to JuliaFormatter later first.

fonsp:
I also don't quite see the use in this - is it just a fun gadget?

I think we should just set the formatter options in /.vscode/settings.json - this is not yet supported by the Julia extension by I have opened an issue about this

domluna:
I'm not familiar with VSCode settings but you could use a config file for the format settings https://domluna.github.io/JuliaFormatter.jl/dev/config/

fonsp:
Sorry, what I meant was - VS Code already has a formatter (which I disabled but I could reconsider). Why add another one as a CI step? It's not solving a problem that I have experienced

domluna:
It's just a way to automate code formatting/style so you don't have to worry about that while you write code. Typically the workflow is that it formats the file on save but as you mentioned for VSCode DocumentFormat.jl is the only supported formatter currently.

It might be beneficial in a CI workflow as a way to keep the formatting/style of a project in sync. Locally people may have different options based on their personal preferences.

Then again this all assumes automated formatting/style is something you want, which may not be the case, although it may grow on you.

fonsp:
Ah okay, I do like the idea of having a "canonical formatter", but I say we revisit this once we experience problems with not having a formatter.

domluna:
https://github.com/julia-vscode/LanguageServer.jl/pull/759 may also be relevant

domluna:
@fonsp as an aside would it possible to format code cells in Pluto as a plugin?

fonsp:
Yeah maybe that would be nice, but also more of a long-term goal. The short term goal is to make computing more accessible to beginners.

Roger-luo:
> Sorry, what I meant was - VS Code already has a formatter (which I disabled but I could reconsider). Why add another one as a CI step? It's not solving a problem that I have experienced

@fonsp I think for development others might not enabled it (consider someone using Vim or other editors), as a result the submitted PR will not in the same format. As a result you will need to manully format it using vscode yourself locally then commit the changes. But this can be automated in CI. Or at least check if someone's PR match the format requirement in CI (like what the JuliaLang/julia does)

fonsp:
I understand the use case, but I don't have strong opinions about Julia formatting - it's all fine. And the development is mostly javascript from this point on.

PR #372, opened 2020-09-06
Highlight selected cell & minor UI tweaks

marius311:
Hi, I decided to mess with some things and get acquainted with the code by trying to smooth over some UI things which I thought were sharp edges as well as adding some simple missing stuff. For now I put everything in one PR here but happy to discuss individual changes and/or take them out. Summary:

1. Highlight selected cell. I often lose track of where in the code I am, this helped. 

selected_cell

2. Remove queued trafficlight animation. The recent work in https://github.com/fonsp/Pluto.jl/pull/349 great, but I think has just a few minor issues. First, I think animating the queued trafficlight is problematic because 1) its visually disorienting to have that many things moving on screen, especially at two different speeds 2) if the animation represents computation, it doesn't make logical sense that the queued cell is animating at all, since its not running yet and 3) because of the way the CSS animation works, there's often a visually jarring "jump" when switching from the slow speed to the fast speed since the fast speed always starts from an offset of zero. For these reasons, I took it out, but kept the repeating gradient on the queued cells, which I think nicely indicates the cells are about to run but currently are not running. I also removed the tiny differences in gradient colors because too many subtly different colors on-screen feels messy. End product: 

queued_cell

(I couldn't figure out how to get rid of the small flash after running is finished which sometimes happens, maybe this is just a CSS deficiency)

3. I removed linewrap, which no code editor that I know of has on by default. 

4. I made the cursor wider (same width as Jupyter's cursor on my screen fwiw), as Pluto's was really thin and I kept losing track of it. Would be good to double check this looks OK on other resolutions too. 

I did find all of these things improved the overall feel for me, but again, happy to discuss. Thanks!

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -167,7 +167,6 @@ export const CellInput = ({
         }
 
         cm.setOption("extraKeys", keys)
-        cm.setOption("lineWrapping", false)
 
         cm.on("cursorActivity", () => {
             if (cm.somethingSelected()) {
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -167,6 +167,7 @@ export const CellInput = ({
         }
 
         cm.setOption("extraKeys", keys)
+        cm.setOption("lineWrapping", false)
 
         cm.on("cursorActivity", () => {
             if (cm.somethingSelected()) {
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -911,8 +911,8 @@ pluto-cell.running.errored > pluto-trafficlight {
     background-size: 4px 22.62741699797px;
     /* 16 * sqrt(2) */
     transition: background 0.05s ease-in;
-    animation: 500s linear 0s infinite running scrollbackground;
-}
+    animation: 1000s linear 0s infinite running scrollbackground;
+
 
 @keyframes scrollbackground {
     0% {
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -1403,7 +1403,7 @@ pluto-cell.code_differs .cm-s-default .CodeMirror-gutters {
 }
 
 .cm-s-default .CodeMirror-cursor {
-    border-left: 1px solid #776e71;
+    border-left: 1.5px solid #776e71;
 }
 
 .cm-s-default span.cm-comment {
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -807,6 +807,7 @@ pluto-trafficlight {
     pointer-events: none;
     border-top-left-radius: 4px;
     border-bottom-left-radius: 4px;
+    border-left-color: rgba(0, 0, 0, 0.1);
     background: rgba(0, 0, 0, 0.1);
     transition: background 0.05s ease-in;
 }
@@ -825,7 +826,8 @@ pluto-cell.code_folded > pluto-trafficlight {
 }
 
 pluto-cell:focus-within > pluto-trafficlight {
-    border-left: 4px solid rgba(0, 0, 0, 0.2);
+    border-left-width: 4px;
+    border-left-style: solid;
     margin-left: -4px;
     background-clip: content-box;
     transition: background 0.05s ease-in;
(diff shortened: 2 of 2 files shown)


fonsp:
Very nice, thanks a lot!

The only thing is 3. - I've already thought hard about line overflows and decided upon the current design: line wrap with small dots before lines to indicate it. I don't want the cell editor to be scrollable content - this makes Pluto feel 'less robust'. If you want, you can open a separate PR about it, but I'd rather let this wait for #2

fonsp:
I'll look into the trafficlight flicker again sometime

marius311:
Great, glad you like it! Reverted 3, that sounds fair. I can imagine at some point in the future it'd be customizable which would be nice too.

fonsp:
I tried it out and I have some comments:

### 2

Queued cells _should_ have a slowly moving traffic light. Right now, when you open a large notebook and the running cell is too far down to see, it just feels like the notebook has crashed. You're right that "moving" means that the cell means "running", but in web UI it often also means "things are not broken, give it a second". Progress bars create a more friendly experience: https://en.wikipedia.org/wiki/Progress_bar#Perception

The original PR #349 had a larger difference in animation speeds - let's try that again.

(I can live with the rotation jump when the animation speed changes.)

### 1

We should fix the positioning for small displays:
image

The 👁 clips with the window edge.

Also: instead of the focus bar being a darker color, maybe we can use the same color as the traffic light? Then the focused cell just has a traffic light that's twice as wide.

marius311:
> Queued cells should have a slowly moving traffic light... it just feels like the notebook has crashed. 

Yea, I can see that. My only comment would be it seems cleaner to do slower speed or dimmed color for queued cells, but not both, mainly because of the too many subtle differences things. Between those two, I think I do prefer the slower speed. 

> Then the focused cell just has a traffic light that's twice as wide.

You're right, I think I broke my own "too many subtly different colors" thing here actually. 

> The eye clips with the window edge.

Good catch, I can try and fix that.

fonsp:
Yeah thanks for pointing out that we don't need to add another color!

About the window edge - we can increase the margin on the left, but do so carefully - this is valuable screen space! Maybe we can squeeze the 👁 a little tighter?

shashankp:
> > Queued cells should have a slowly moving traffic light... it just feels like the notebook has crashed.
> 
> Yea, I can see that. My only comment would be it seems cleaner to do slower speed or dimmed color for queued cells, but not both, mainly because of the too many subtle differences thing. Between those two, I think I do prefer the slower speed.
> 
> > Then the focused cell just has a traffic light that's twice as wide.
> 
> You're right, I think I broke my own "too many subtly different colors" thing here actually.
> 
> > The eye clips with the window edge.
> 
> Good catch, I can try and fix that.

thank you for the inputs 
This needed some direction & feedback to see which animation works best.

fonsp:
Ping me when you're done!

marius311:
After some more thought and testing, I don't think the two speeds are that good. The goal should be that when quickly glancing at the screen, you can easily identify what's running and what's queued, and trying to identify two different speeds in the end I don't think works. Instead, I came up with this:

queued_cell

I also tweaked the colors so the error-red matches the percieved alpha of the gray (red was really bright before), and such that the hatches always contain one color that matches the cell/selection highlight color so selecting things while they're running/queued still looks good. Ready to review from my end.

fonsp:
This looks very very nice :)

fonsp:
(I squashed the eyeball a little to make the left page margin smaller, and made it jump left when the cell is in focus.)

fonsp:
Yayo thanks so much for your tasteful PR!

marius311:
Great, thanks for the comments, they definitely improved the initial attempt! 

> (I squashed the eyeball a little to make the left page margin smaller, and made it jump left when the cell is in focus.)

FWIW I will say I don't think I like this. Every time you click a cell to select it, the little eyeball appears then immediately moves, and kind of draws your eye away unnecessarily. I'd say its worth the 4px to remove the distraction. I'll leave it up to you though!

fonsp:
You’re right, and I’m not stuck to the current thing

Let’s leave it for a little while and fix it later? Also I’m still not sure how relevant the 200px wide screen layout is

PR #373, opened 2020-09-06
add closebrackets CodeMirror addon

marius311:
This (or its equivalent) are enabled by default in Jupyter/Juno/VSCode from what I can tell. It always surprised me when I select some text and hit `(` expecting to wrap the current selection in parenthesis and it doesn't work in Pluto. This fixes that.

(sorry for the barrage of simple-ish PRs, just trying to share some things that have improved my productivity, always happy to discuss)

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -167,6 +167,7 @@ export const CellInput = ({
         }
 
         cm.setOption("extraKeys", keys)
+        cm.setOption("autoCloseBrackets", true)
 
         cm.on("cursorActivity", () => {
             if (cm.somethingSelected()) {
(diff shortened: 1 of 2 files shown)


fonsp:
This came up before - I like these kinds of cyborg addons, but I'd like to try this out before putting it in. 

I'm assuming that it just completes brackets, and not `begin end`, `let end`, and that's a missed opportunity! Julia is a blocky language, and especially `end` should be auto-inserted, just like how you get `}` in C-style languages.

But maybe this is nice on its own (if it isn't too aggressive)

dralletje:
Sorry I superseded this pull request with a commit oops

PR #378, opened 2020-09-07
Changed button opacity for pointer:coarse to be 0.25 permanently

### LINKED ISSUE #348: Buttons clickable but not visible on touch screen
Recently, I often find myself hide/show cells which is a really cool feature. However, it is hard to find the cell when it is hidden and has no output. It still takes one line of space, but the hide/show symbol is not shown when a cell is not selected and it cannot be selected when hidden so that the symbol appears. 🐔 🥚 

Luckily it is still clickable even though invisible.

A similar thing applies to the delete button which is not shown when a cell is not selected but it is still clickable, I sometimes deleted cells which I only wanted to select because I clicked in the top right where the invisible delete button is. 

I had the impression that this was working as expected some weeks ago.

This change was probably implemented to have more clutter-free view on the notebook. 
So maybe these symbols can appear on mouse hover over the cell or so?

sravanth-chebrolu:
This PR fixes #348. 

Modified CSS `opacity` for `@media screen and (any-pointer: coarse)`. On `coarse`, all buttons will always be displayed.

image

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -934,9 +934,14 @@ pluto-input > button > span {
 
 @media screen and (any-pointer: coarse) {
     pluto-cell > button,
+    pluto-cell > pluto-runarea
+    {
+        opacity: 0;
+        /* to make it feel smooth: */
+        transition: opacity 0.25s ease-in-out;
+    }
     pluto-input > button,
-    pluto-shoulder > button,
-    pluto-cell > pluto-runarea {
+    pluto-shoulder > button {
         opacity: 0.25;
     }
     pluto-cell:focus-within > button,
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -937,9 +937,7 @@ pluto-input > button > span {
     pluto-input > button,
     pluto-shoulder > button,
     pluto-cell > pluto-runarea {
-        opacity: 0;
-        /* to make it feel smooth: */
-        transition: opacity 0.25s ease-in-out;
+        opacity: 0.25;
     }
     pluto-cell:focus-within > button,
     pluto-cell:focus-within > pluto-input > button,


fonsp:
Does it look the same for `pointer: fine`?

I notice that I used `any-pointer: coarse` instead of `pointer: coarse` - should that be changed?

fonsp:
And thanks a lot!!

sravanth-chebrolu:
> Does it look the same for `pointer: fine`?

I think `pointer:fine` has no issues to begin with, the problem arose because touch displays have a `pointer:coarse`, right?

> I notice that I used `any-pointer: coarse` instead of `pointer: coarse` - should that be changed?

That bit confused me as well, from what I gather( based on the MDN docs [this](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/any-pointer) and [this](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer)):

Using `any-pointer: coarse`, will target the less accurate pointer, and `any-pointer: fine` will target the more accurate one, if both exist on the device. However using `pointer: coarse/fine` will only target the primary pointer, so I'm guessing `any-pointer` is the way to go?

fonsp:
Yep! Thanks

sravanth-chebrolu:
P.S: This was my first PR!

fonsp:
Exciting! What do you want to work on next?

If you like CSS or React, have a look here:
https://github.com/fonsp/PlutoUI.jl/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22good+first+issue%22
and here:
https://github.com/fonsp/Pluto.jl/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22good+first+issue%22+label%3Afrontend

sravanth-chebrolu:
Will look around and contribute to any issue that strikes a chord with me, thanks!

Issue #359, opened 2020-09-05
LaTeXStrings repr with mixed text/math


zmoon:
The current approach (https://github.com/fonsp/Pluto.jl/commit/7b5dd1d8aa4720d28b6a2b0a665a32d6ae7f634c), stripping `$` from beginning and end before passing to `Markdown`'s `LaTeX`, does not work in the case of mixed text/math-mode, which LaTeXStrings [does support](https://github.com/stevengj/LaTeXStrings.jl#usage). 

![image](https://user-images.githubusercontent.com/15079414/92292627-36f5ea80-eeec-11ea-93bf-910826960f89.png)

Since it ends up getting rendered with MathJax, one solution would be to wrap the text (non-math) parts with `\text{}` before passing to `LaTeX`. If that seems like a good idea, I would be interested in trying to implement it.

fonsp:
Aha - is that why the package exists? I'm always confused why we don't just create a macro `L"..."` that goes to `Markdown.LaTeX`. With your solution, you could have that feature and still just use `Markdown.LaTeX`. 

Anyway, yes your solution sounds good - you should play around with it inside a Pluto notebook, instead of modifying the Pluto source and rebooting everything while you work on it. Let me know if you need help with that

fonsp:
I guess the other reason why `LaTeXStrings` exists is that `LaTeXString <: AbstractString`?

fonsp:
Also why can I not do this?
```julia
equation = L"\sqrt{100}"
md"I am $(equation) years old."
```
This would not be a problem if they used the same underlying type

fonsp:
Sorry, all of that is not really relevant here

zmoon92:
> Also why can I not do this?
> 
> ```julia
> equation = L"\sqrt{100}"
> md"I am $(equation) years old."
> ```

I tried this in Pluto and seemed to work fine?
![image](https://user-images.githubusercontent.com/15079414/92307108-fc2e9980-ef61-11ea-96b2-8ed1382a80b5.png)

zmoon92:
<details>
<summary>Here is what I came up with so far...</summary>

```julia
begin

    # I'm sure there is a better way to do this...
    # e.g., `length(findall(...` ? with positive look-behind regex
    # or a comprehension
    function count_char_nonescaped(s::AbstractString, c)
        count = 0
        for i = 1:length(s)
            if i == 1
                if s[i] == c
                    count += 1
                end
            else
[...]

PR #381, opened 2020-09-07
Wrap LaTeXString with `text{}`

zmoon:
Wrap `LaTeXString` with `text{}` in order to preserve mixed text/math in the rendered repr using MathJax.

https://github.com/fonsp/Pluto.jl/issues/359

### DIFF
--- a/frontend/editor.html
+++ b/frontend/editor.html
@@ -41,7 +41,7 @@
     <script src="warn_old_browsers.js"></script>
 
     <script src="common/SetupMathJax.js"></script>
-    <script type="text/javascript" id="MathJax-script" src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-svg-full.js" async></script>
+    <script type="text/javascript" id="MathJax-script" src="https://cdn.jsdelivr.net/npm/mathjax@3.1.2/es5/tex-svg-full.js" async></script>
 </head>
 
 <body class="loading no-MαθJax">
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -344,10 +344,9 @@ function show_richest(io::IO, @nospecialize(x); onlyhtml::Bool=false)::MIME
                     htmlesc(io, repr(mime, x; context=iocontext_compact))
                 end
             elseif mime isa MIME"text/latex"
-                # LaTeXStrings prints $ at the start and end.
-                # We strip those, since Markdown.LaTeX only contains the math content
+                # Wrapping with `\text{}` allows for LaTeXStrings with mixed text/math
                 texed = repr(mime, x)
-                html(io, Markdown.LaTeX(strip(texed, '$')))
+                html(io, Markdown.LaTeX("\\text{$texed}"))
             else                
                 show(io, mime, x)
             end


zmoon:
Is the `texed = repr(mime, x)` line still necessary? What else is needed?

fonsp:
How did the MathJax issue go?

zmoon:
Good, they said the fix should be in the next patch release.

fonsp:
@zmoon MathJax released an update, can you check if it works with their latest version? Can you update the version that Pluto is using?

zmoon:
It seems like there are still issues after the initial fix: https://github.com/mathjax/MathJax/issues/2524#issuecomment-691698767

Pluto specifies `mathjax@3` so it should use the latest npm version without changing anything.

https://github.com/fonsp/Pluto.jl/blob/18468239340c956b5e1c71f42757b597a6e4b50c/frontend/editor.html#L43-L45

fonsp:
Oh that should be a full version instead, can you set it to `3.1.2`?

The new bug looks fine to ignore

zmoon:
Seems to work.

image

fonsp:
Thanks!

stillyslalom:
MathJax isn't meant for text-first rendering, and treating it otherwise seems to be causing more problems than it fixes. Better to use native Markdown constructs to handle text/inline math:
image

zmoon:
The goal here was just to make `LaTeXString`s show nicely in Pluto. They are supposed to represent text-first LaTeX (consistent with the MIME type `text/latex`, afaict). Writing in the Pluto notebook it is certainly more convenient to use Markdown, though `$...$` is more common than ``` ``...`` ``` I would think. 

For the short term it seems like it might be better for `LaTeXStrings` display to be special-cased...?

stillyslalom:
But the display for `MIME"text/latex"` is done by MathJax, which isn't meant for text-first rendering - due to the lack of line-wrapping, it can't display more than a few words of text, and certainly not a paragraph.

image

fonsp:
LaTeXStrings is such a mess oof

zmoon:
> But the display for `MIME"text/latex"` is done by MathJax, which isn't meant for text-first rendering - due to the lack of line-wrapping, it can't display more than a few words of text, and certainly not a paragraph.

Yeah, it's not optimal, but I think it works fine for the main use case of `LaTeXStrings` - constructing labels and annotations for plots. But that shouldn't come at the expense of other packages' displays.

Maybe in the future the `text/latex` MIME type could be treated in a more complex way, leaving normal text as normal text and giving the LaTeX environments and math text to MathJax. This seems to be what IPython display does in Jupyter.

fonsp:
That doesn't work for `\begin{equation}`

zmoon:
Doesn't seem to be a problem
image

`$$...$$` is not recognized the same way that it is in `md"..."` though.

fonsp:
I am very confused, can you write it without a L_str inside Markdown.parse?

fonsp:
It would be really really helpful if you write a PR for this an create a notebook showing off all the things that we want to support. If not then maybe it's best to drop support for LaTeXStrings again

zmoon:
> I am very confused, can you write it without a L_str inside Markdown.parse? Or was that the point?

I used `L_str` partially for convenience and partially to check it would work for `LaTeXString` type. But I can confirm it works the same for normal string too as long as things are escaped properly.

> It would be really really helpful if you write a PR for this an create a notebook showing off all the things that we want to support.

I will see what I can do!

stillyslalom:
I'd recommend using `@raw_str` for this sort of thing:

image

fonsp:
@zmoon Great! Just the testing notebook would already be a big help

zmoon:
@fonsp I made a notebook testing some options [here](https://gist.github.com/zmoon/3c44fc091352a37037c41acac56a618f#file-pluto-textlatex-test-jl)

What I discovered is that although my simple example with `Markdown.parse` above worked without proper math delimiting, in general that is not the case (e.g., if the equations have multiple `_` or `*` in them). That is (https://github.com/JuliaLang/julia/issues/37334), either `$...$` starting on same line or the Julia Markdown math delimiters should be used. Not everything that should be rendered by MathJax comes in with the proper delimiters to mark it so (e.g., `\begin{equation}...`, since it is valid LaTeX without math delimiters), so some parsing/processing would be necessary (e.g., with mixed text and `\begin{equation}...` to add math delimiters to the equation part).

The issue is that while `text/latex` MIME type may not be, `Markdown.LaTeX` and MathJax are math-first. But most things that use `text/latex` in the notebook context seem to be math-only or at least math-first. So I guess some options are
1. revert to the previous usage of `Markdown.LaTeX` without `\text` and either
   a. special-case `LaTeXString`s
   b. or not
2. implement parsing `text/latex` and adding math delimiters where necessary to support mixed text/math (like LaTeX does), showing it in the notebook as mixed normal text and MathJax similar to Jupyter (in Jupyter notebook at least, [seems to](https://github.com/jupyter/notebook/blob/master/notebook/static/base/js/mathjaxutils.js#L59) be done in the js part)

Seems like 1. is the way to go for now?

Issue #380, opened 2020-09-07
Images dont show in exported HTML


fonsp:
MWE: https://github.com/mitmath/18S191/blob/master/homework/homework1/hw1.jl

fonsp:
We need to convert those `blob`s into `data` base 64 URLs

PR #384, opened 2020-09-08
blob images to base64 data on html export

shashankp:
https://github.com/fonsp/Pluto.jl/issues/380


fonsp:
Nice, thanks!

The function is async now, does that mean that if you are fast enough, the download link does not work?

fonsp:
Or maybe we can just assume that no one is that fast

fonsp (changes requested):
https://github.com/fonsp/Pluto.jl/pull/384#issuecomment-688728734

frontend/components/Editor.js
```diff
@@ -765,16 +765,23 @@ export class Editor extends Component {
...
+                                    download_link.href = URL.createObjectURL(
+                                        new Blob([html], {
                                             type: "text/html",
                                         })
                                     )
+                                    document.body.appendChild(download_link);               
+                                    download_link.click();                                   
+                                    document.body.removeChild(download_link); 
+                                }).finally(() => {
+                                    export_card.classList.remove("export_in_progress");
```
shashankp, line 783: For large files / online notebooks, user can be given visual feedback

fonsp: Thanks for this addition! 

Even if it is online, the blobs are already in memory (right?), so the overhead should be tiny.

I will get a big notebook to test this one

frontend/common/OfflineHTMLExport.js
```diff
@@ -1,27 +1,60 @@
...
-    <html lang="en">
-    <head>
-        <meta name="viewport" content="width=device-width" />
-        <title>⚡ Pluto.jl ⚡</title>
-        <meta charset="utf-8" />
+    var blob_to_base64_promises = [];
+    Array.from(body.querySelectorAll("img")).forEach((img) => {
+        if (img.src.match(/^blob:/)) {            
+            blob_to_base64_promises.push(new Promise((resolve) => {
+                var xhr = new XMLHttpRequest();
```
fonsp, line 11: I found this SO answer, looks like we can use `fetch` instead - that would be easier on the eyes

https://stackoverflow.com/questions/11876175/how-to-get-a-file-or-blob-from-an-object-url

I'll try this myself

fonsp:
Sorry I clicked on "Request changes" but I will look it this myself

shashankp:
> Sorry I clicked on "Request changes" but I will look it this myself

Please check the added commits to get around the raised issue.

fonsp:
Thanks, it's merged! I'm a git noob so i made a new PR with your commits: https://github.com/fonsp/Pluto.jl/pull/457

PR #389, opened 2020-09-08
Move welcome message to end of code loading

ianshmean:
I noticed that for me, the welcome message appears well-before code loading has finished and the repl is ready. This might make for nicer UX?

### DIFF
--- a/src/Pluto.jl
+++ b/src/Pluto.jl
@@ -27,18 +27,6 @@ const ENV_DEFAULTS = Dict(
 )
 get_pl_env(key::String) = haskey(ENV, key) ? ENV[key] : ENV_DEFAULTS[key]
 
-if get(ENV, "PLUTO_SHOW_BANNER", "true") == "true"
-@info """\n
-    Welcome to Pluto $(PLUTO_VERSION_STR) 🎈
-    Start a notebook server using:
-
-  julia> Pluto.run()
-
-    Have a look at the FAQ:
-    https://github.com/fonsp/Pluto.jl/wiki
-\n"""
-end
-
 include("./evaluation/Tokens.jl")
 include("./runner/PlutoRunner.jl")
 include("./analysis/ExpressionExplorer.jl")
@@ -64,5 +52,17 @@ include("./webserver/Static.jl")
 include("./webserver/Dynamic.jl")
 include("./webserver/REPLTools.jl")
 include("./webserver/WebServer.jl")
+
+if get(ENV, "PLUTO_SHOW_BANNER", "true") == "true"
+@info """\n
+    Welcome to Pluto $(PLUTO_VERSION_STR) 🎈
+    Start a notebook server using:
+
+  julia> Pluto.run()
+
+    Have a look at the FAQ:
+    https://github.com/fonsp/Pluto.jl/wiki
+\n"""
+end
 
 end


fonsp:
Right now it only prints it once, when Pluto is precompiling, but not when you launch it for the second time. It this is nice (accidental) feature.

Is it the same with `__init__`?

ianshmean:
Ah, no. With this it would print each time (I assumed that was what was wanted)

Perhaps the `@info` could just be moved to the end of the `include` statements?

fonsp:
Oh yeah that would be perfect! Good iea

ianshmean:
@fonsp updated! 👍

ianshmean:
I just realized that if the user precompiles via `Pkg.precompile()` or `pkg> precompile`, then this message will print during that process, which doesn't make sense, and then won't print again.
I think it needs to be moved into `__init__()`, but I'm not sure how to make it print only the first time after compilation.

fonsp:
This is fine, it's only there to say hello and make the software seem less intimidating, if you are precompiling then you're probably comfortable enough to do without the message.

kalmarek:
it makes sense to print banner in interactive mode (you may add `isinteractive()` to the boolean check).
tbh. I don't think it's great idea to display welcome messages. I couldn't find a way to circumvent these as well, despite efforts:
https://discourse.julialang.org/t/redirecting-stdout-to-avoid-banners-while-import-ing/37633/8

any new ideas are welcome though!

fonsp:
I tried `isinteractive()` but it is always false inside package code, it can also give `true` when called from an interactive env

kalmarek:
not sure what do you mean by "always `false` inside package code"?
it will certainly be `true` if you start REPL, or run a script with `julia -i script.jl`, no matter whether inside a module or not

fonsp:
Try it out yourself: make a package that prints `isinteractive()`, and load it using `import`

ianshmean:
From within `__init__()` or just directly in the module?

kalmarek:
@fonsp this is what I get:
```julia
module MyPackage

__init__() = @show isinteractive()

end
```

```julia
$ julia --project # starting repl
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.1 (2020-08-25)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using MyPackage
[ Info: Precompiling MyPackage [c2604015-7b3d-4a30-8a26-9074551ec60a]
isinteractive() = true

```
```
$ julia --project -e "using MyPackage" # just executing a script
isinteractive() = false

```
```
$ julia --project -i -e "using KnuthBendix" # executing in interactive mode
isinteractive() = true
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.1 (2020-08-25)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> 
```
how did you do?

kalmarek:
I guess you tried to store the value in a variable ;) then indeed it's assigned during precompilation, where `isinteractive()== false` and never changed later. e.g.
```julia
module MyModule
    using Dates
    const date_loaded = now()
    get_date_loaded() = date_loaded
end
```
precompilation of this package will fix `date_loaded` to the time and date when the package was precompiled, not, when it's loaded ;)

fonsp:
Oh that explains it, thanks!

But still, I am happy with the current accidental behaviour.

PR #390, opened 2020-09-08
📝 Add Ctr+M shortcut for toggling Markdown mode

ekzhang:
Hey there, thanks for making this awesome project! This notebook server has been really useful to me.

One thing I noticed is that it gets repetitive to keep typing `md"` every time I want to start a Markdown block. This PR aims to try to make this easier by adding a keyboard shortcut to do the same. Right now I picked Ctrl+M/Cmd+M, but feel free to change this if you prefer.

Hope this aligns with the direction you're taking the project!

### Changes

- Added a shortcut (Ctrl+M/Cmd+M) to change a code cell to Markdown, or a Markdown cell to code
- Robust to leading/trailing whitespace
- Retains your cursor position in the cell

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -108,6 +108,42 @@ export const CellInput = ({
                 cm.execCommand("selectAll")
             }
         }
+        keys[mac_keyboard ? "Cmd-M" : "Ctrl-M"] = () => {
+            const value = cm.getValue()
+            const trimmed = value.trim()
+            const offset = value.length - value.trimStart().length
+            if (trimmed.startsWith('md"') && trimmed.endsWith('"')) {
+                // Markdown cell, change to code
+                let start, end
+                if (trimmed.startsWith('md"""') && trimmed.endsWith('"""')) {
+                    // Block markdown
+                    start = 5
+                    end = trimmed.length - 3
+                } else {
+                    // Inline markdown
+                    start = 3
+                    end = trimmed.length - 1
+                }
+                if (start >= end || trimmed.substring(start, end).trim() == "") {
+                    // Corner case: block is empty after removing markdown
+                    cm.setValue("")
+                } else {
+                    while (/\s/.test(trimmed[start])) {
+                        ++start
+                    }
+                    while (/\s/.test(trimmed[end - 1])) {
+                        --end
+                    }
+                    // Keep the selection from [start, end) while maintaining cursor position
+                    cm.replaceRange("", cm.posFromIndex(end + offset), { line: cm.lineCount() })
+                    cm.replaceRange("", { line: 0, ch: 0 }, cm.posFromIndex(start + offset))
+                }
+            } else {
+                // Code cell, change to markdown
+                cm.replaceRange(`\n"""`, { line: cm.lineCount() })
+                cm.replaceRange('md"""\n', { line: 0, ch: 0 })
+            }
+        }
         const swap = (a, i, j) => {
             ;[a[i], a[j]] = [a[j], a[i]]
         }


fonsp:
Woa it even keeps the cursor position! Fantastic!

fonsp:
About the keyboard shortcut - VS Code has a semi-undocumented notebook mode, and I want Pluto to use the same default keyboard shortcuts wherever possible. Do they have one?

ekzhang:
Thanks! The keyboard shortcut in VS Code is just "M" when you're selecting the cell in command mode.

fonsp:
Awesome, thanks Eric! It will be released in today's update :)

What do you want to work on next? I'm always happy to have a video call if you want to discuss open issues/different approaches/future ideas/the weather: fonsvdplas@gmail.com

ekzhang:
Amazing! In terms of low-hanging fruit, I wish there was a way to copy/paste cells, and I also wish there was a simpler way to run Pluto from the terminal, like a `pluto` executable in the path. Maybe this is an unconventional way to do things though, not sure if I can say much on that since I'm inexperienced in Julia.

Anyway, great stuff! It was super easy to set up the dev environment, will definitely send another PR if I find something in the future... (using Pluto for a course this semester because I hate Mathematica 😅).

fonsp:
Good to hear! Look forward to hear more from you!

The Pluto CLI exists 😊 - https://github.com/fonsp/PlutoUtils.jl

fonsp:
Copying + pasting cells would be great! I hope you discovered the cell selection feature? I assume you did, coming from Mathematica :)

It needs to work among different notebooks! And if possible, different notebooks from different sessions (e.g. one on binder, one running locally).

Maybe you could think about the contents of the clipboard? Some options:
1. It doesn't matter, you put the data in localStorage. But then it only works within the same session
2. The copy data is exactly like it is in the file (i.e. `# |==\ncode\n\n` ish). And when you paste onto the page it recognizes that format (and generates new cell IDs to avoid duplicates). This is cool, because people often post notebooks in that format on github.
3. The copy data also contains formatted output? Like what happens right now if you text-select and copy+paste to Google Docs. 


Another thing: for this you would want to place a superbig text cursor after a cell or between two cells, before you hit Ctrl+V. I think this could be really cool, if you also allow text input to create a new cell.

fonsp:
Let me make that a new issue

fonsp:
#399

lyon-fnal:
Thanks for adding this nice feature. I'll note that by default, Cmd+M minimizes the current window. To get your feature to work, I had to reassign the Minimize action to a different key (see https://apple.stackexchange.com/questions/115562/how-do-i-disable-the-minimize-command-m-shortcut-in-mavericks). Reassigning is fine with me, as I sometimes hit Cmd+M by mistake ... the minimizing window can be somewhat jarring. Now that's no longer a problem and I can use your feature!

fonsp:
@lyon-fnal As a mac user, can you suggest a solution?

lyon-fnal:
Yes - I did in my previous post - sorry that wasn't clear. One can reassign the Minimize function to a different key (e.g. not Cmd+M). See  https://apple.stackexchange.com/questions/115562/how-do-i-disable-the-minimize-command-m-shortcut-in-mavericks (works for Catalina) for instructions.

ekzhang:
@lyon-fnal I also use a Macbook Pro and haven't experienced this being an issue when I was developing. What browser are you using?

Cmd+M does indeed map to "minimize" for me, but it's not doing that when my cursor is in the CodeMirror editor.

fonsp:
Ctrl+M on an empty cell places the cursor at the end of the cell, instead of inside the markdown literal. Could you have a look?

image

lyon-fnal:
It does that for me too on Safari.

ekzhang:
Good point @fonsp. I had some spare time today and was planning to work on #399 , so this fits well into my schedule - will look at it.

ekzhang:
Okay, I don't have a good fix for this unfortunately due to some limitations of CodeMirror's API. The best solution I could find is to add a hack when `getCursor()` returns the last index of the document, to manually call `setCursor()` and move it back to before the inserted `"""`. This doesn't work with multiple cursors, selection regions, or multiple selection regions though.

I'll send in a PR anyway, since it seems like Ctrl-M on an empty cell is a fairly common case...

Issue #196, opened 2020-06-26
Macros and DSLs


zenon:
I try out Soss.jl, a tool for probabilistic programming.

I can't attach .jl files, so I changed them to txt.

This (non Pluto) works fine:
[sosstest01.txt](https://github.com/fonsp/Pluto.jl/files/4839201/sosstest01.txt)

This (Pluto) doesn't:
[sosstest02.txt](https://github.com/fonsp/Pluto.jl/files/4839208/sosstest02.txt)

Error message:
Cyclic references among m, y and forwardModel:
Combine all definitions into a single reactive cell using a `begin ... end` block.

Am I doing sth wrong, or has Pluto an issue with macros?

Kind greetings, z.

fonsp:
Pluto doesn't understand macros, it just looks at the code as if the macro wasn't there.

This is a difficult problem: the macro is not known to Pluto - it is only defined inside the notebook process. To understand (i.e. _expand_) macros, there needs to be an analysis step _after_ normal analysis, but _before_ runtime (because the analysis determines which cells will run).

So _Domain Specific Languages_ like Soss.jl might not work well.

zenon:
Hello Fons, thank you for the fast answer. 

Hm. I was afraid it would come out this way. :-)

Kind greetings, z.

fonsp:
About DSLs: I am thinking that maybe Pluto can include a "mock version" for a bunch of popular packages with DSLs. 

So not the original macro, but a macro that transforms the code into something that has the intended _references_ and _definitions_. For example, Replace=with~.jl could transform
```julia
@replaceit x ~ y + 1
```
(Pluto thinks that `x`, `y` and `+` are the references, with no assignments)
to
```julia
x = y + 1
```

fonsp:
Although, looking at this issue again - the problem is not the DSL, it's a bug in Pluto's syntax analysis! #225

fonsp:
Your original issue should be fixed now! But the DSL problem is still there

zenon:
Yes! It now works (even if I understand that there are no guarantees :-) )

Thank you for the hint.

aplavin:
Another important case when this issue happens in with the `Query.jl` package: `df |> @map({a = _.b})` is treated as assigment to `a` by `Pluto`.

fonsp:
See a workaround for some macros here: #371

fonsp:
[...]

Issue #142, opened 2020-06-03
Package environment


zenon:
Hi,

I try and use many different libraries, and only some of them I keep. So, to start a project that uses a certain set of libraries, I enter a new directory, and in Julia I say

`] activate "."`

And now, I have a clean environment.

Pluto, on the other hand, starts its Notebooks in the standard environment (thus, for me, doesn't find any library).

So I added some cells before my "using XXX" lines:

using Pkg
Pkg.activate("myPath")

Side note: This isn't very portable. If there'd be a better way, I'm happy.

When writing the notebook, everything went fine. When reloading it, the ordering by Pluto kicked in, and all packages were loaded up front, and thus not found (or so I interpret what I saw).
This reordering seeems sensible to me in all cases except when working with environments :-)

I have no good idea where to put such meta information. Somehow I either
- need to be able to tell Pluto that some code lines must get priority (Initializer code)
- directly have a syntax for activating an environment.

Both makes Pluto more complex, I'm afraid.

As "using XXX" has to be on the top level, I don't see a way to put this into functions, like, I mean

     # not syntactically correct julia 
     function initialize()
        using Pkg
        Pkg.activate("myDir")
    end

    x = do
       initilize()
       using XXX
       using YYY
    end

Any ideas?

Kind greetings, z.

fonsp:
Hi z,

Very good point! I have been thinking a lot about this lately - I think that every Pluto notebook should start in its own package environment, and there should be a GUI for loading packages, or this can be done with `using XXX` directly.

The other thing here is that a Pluto notebook should contain that package info! That way a `.jl` is completely reproducible (as far as packages go). Before going too much into detail - what do you think about this idea?

fonsp:
Also, you can use 
```julia
let
    import Pkg
    Pkg.activate(".")
    Pkg.add("XXX")
end
```
as you first cell, and it _should_ also be the first cell to run. Does that work?

zenon:
Ah!

Thank you Fons!, I gave up whan reading the error message that 'using' must be top level; didn't expect to find that there a re more top levels than I thought. :-)

Yes, a let/end or begin/end block seems to do it. (Assuming that my single experiment showed characteristic behavior.) Thank you!

I'll coment with thoughts on a general solution in the next hours.

Kind greetings, z

zenon:
Some thoughts.

Creating a directory (to put the environment in), and installing packages is potentially dangerous.
I wouldn't do it without asking.

[...]

PR #396, opened 2020-09-09
Adding macro for activating parent git repo environment

cadojo:
This is a macro for code I use at the top of my notebooks. It searches for a Git repository in a directory above the notebook working directory. If one is found, then it runs `Pkg.activate` on that top-level path.

It partially addresses one non-default Pluto use case - storing notebooks in Git repositories. This has been discussed in [#142].

The usage is simply `@activate_repo_env`.

Despite Pluto not fully understanding macros (#196), I think this macro should work. It doesn't use any outside variables, and all variables in the macro are locally scoped within a `let` block.

I don't think this is an ideal fix - maybe a better solution would be to add arguments for SVN, Git, or the current working directory? Or maybe switch this from a macro to a function, or put it in a `Pluto.run` argument?

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -14,8 +14,9 @@ import Base64
 import REPL.REPLCompletions: completions, complete_path, completion_text
 import Base: show, istextmime
 import UUIDs: UUID
+import Pkg
 
-export @bind
+export @bind, @activate_repo_env
 
 MimedOutput = Tuple{Union{String,Vector{UInt8}}, MIME}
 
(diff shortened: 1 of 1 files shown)


Roger-luo:
I feel this might not necessary to be in the frontend but can just serve as an option when one runs `Pluto.run`?

cadojo:
> I feel this might not necessary to be in the frontend but can just serve as an option when one runs `Pluto.run`?

I agree that could work too. One thought I had against that was the case where you're running multiple Pluto notebooks off of one server. If you're running two notebooks that don't use the same environment, then you'd need to run multiple server instances, right?

Either way, I agree the front end isn't the best place for this. The best case scenario is each notebook using its own workspace, so the environment is entirely contained in the notebook.

fonsp:
This should not be part of Pluto - it should be a small code snippet that you share with others.

Adding syntax to every notebook is a very serious decision that cannot be reversed without breaking old notebooks. We should avoid it whenever possible.

In this case, we will want to remove this macro in a couple of months when #142 is fixed.

fonsp:
But @Roger-luo it sounds like the requested feature is to run your notebook with [`julia --project`](https://discourse.julialang.org/t/pattern-for-activating-the-current-project/15379/2?u=fonsp). We support this, right? Since we use `nothing` instead of `""`?

Roger-luo:
> If you're running two notebooks that don't use the same environment, then you'd need to run multiple server instances, right?

No, we have per notebook environment now. You can specify an environment for each notebook directly, it's a PR just get merged #341 

A more convenient interface is implemented in https://github.com/fonsp/PlutoUtils.jl/pull/9/files#diff-0df7c7b3a0fa9ac7a22c08596327a096R23 if you are familiar with CLIs.

You can choose to start pluto manually with that command, or just use the same code to do it.

PR #403, opened 2020-09-10
Replace hyphen with Em dash

Immortalin:
The default dash looks like unary minus, especially with the default notebook color scheme. It is quite confusing for beginners. `cans_in_stock = consumption -` looks like a mistyped `cans_in_stock = consumption--`


fonsp:
Is that the only dash in the notebook?

PR #405, opened 2020-09-10
Dont launch browser when inside docker

shashankp:
https://github.com/fonsp/Pluto.jl/pull/190

https://github.com/fonsp/PlutoUtils.jl/pull/13#issuecomment-690056330

### DIFF
--- a/src/webserver/WebServer.jl
+++ b/src/webserver/WebServer.jl
@@ -32,7 +32,9 @@ end
 
 function open_in_default_browser(url::AbstractString)::Bool
     try
-        if Sys.isapple()
+        if isfile(".dockerenv")
+            false
+        elseif Sys.isapple()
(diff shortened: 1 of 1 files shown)


fonsp:
On docker, Pluto should be launched with this option disabled (and in the future, with some extra security features disabled):
```julia
Pluto.run(; configuration=ServerConfiguration(launch_browser=false), security=ServerSecurity(something...))
```
I like this more than checking for docker. What do you think?

shashankp:
> On docker, Pluto should be launched with this option disabled (and in the future, with some extra security features disabled):
> 
> ```julia
> Pluto.run(; configuration=ServerConfiguration(launch_browser=false), security=ServerSecurity(something...))
> ```
> 
> I like this more than checking for docker. What do you think?

yup, more sensible 👍

Issue #383, opened 2020-09-08
@test macro with rtol


cnaak:
In the following, `rtol` is just a parameter of the infix operator `≈` (`approx`), but `Pluto` thinks an `rtol` variable has been created:

`@test A ≈ B rtol = eps(Float64)`

Side effects include not being able to create an `rtol` variable in another cell, which errors with: "**Multiple definitions for rtol:**"

fonsp:
#196 #407

jcmkk3:
I'm not sure if it would be useful, but the [presentation](https://youtu.be/0uzrH2Ee494) for Salsa.jl was referencing some issues very similar to this.

fonsp:
Fixed in https://github.com/fonsp/Pluto.jl/commit/c6fa171b0ee7cbffcd151a982ec92a4b1b22c8f4

Issue #330, opened 2020-08-31
Plugins/Extensions ?


onetonfoot:
Thanks for this it's really great. I've wanted something like Observable in Julia land for a while now. 

Do you have any plans for extensions/plugins of some kind?  I'm addicted to using Vim bindings and, it would be great to have them in this environment also?

evadknarf:
surfingkeys?  a browser extension though...

fonsp:
There will be plugin support once we have a better idea of the scope of which things the plugins should control, but if it's just a front-end change (like keybindings), a browser extension is already a good method.

You could also wait for #56 

So please post more ideas of what plugins should be able to do!

PR #407, opened 2020-09-10
Add MacroZoo -- pre-processing to help ExpressionExplorer understand

mcabbott:
This treats expressions like `a[i] := ...` appearing within macros as the notation of `@einsum` and friends, in which this is assignment `a`, and `i` is a dummy index not a variable. 

As suggested in #196, because `:=` is not allowed in ordinary code. If there are any macros which use `:=` without appearing to indexing something, then this should change nothing. 

Probably the function `einsum_strip` is in the wrong place. Not sure that `push!(res.funccalls, funcname)` is the right way to keep the macro's name. One test may want fixing.

Example notebook: https://gist.github.com/mcabbott/2f9342705226619cd8c47e24b5fb2e4d


fonsp:
Thanks for the PR! Do you know which macros are affected, besides the ones you wrote tests for?

I appreciate your work, but it is a lot of new, complicated code, that I will need to maintain. Doing this for every DSL macro out there is not an option - the project needs to remain something that enjoy working on. (I don't mean to be negative about your code style - it looks great and matches the rest of `ExpressionExplorer`.)

On the other hand, your addition is well contained in its own functions - maybe I need to accept that Pluto will have chunks of code that I don't need to touch?

Let me continue this comment in #196

mcabbott:
I don't know of any macros using `:=` to mean something else, but am not sure how to search exhaustively (and someone could make a package tomorrow). OMEinsum is the other package which uses it the same way. 

The branch for `:=` should be inside `elseif ex.head == :macrocall`, but perhaps I've misunderstood the logic? (It's also inside ` if (length(funcname) == 1 || (length(funcname) >= 2 && funcname[1] == :Base))` which might be wrong, `Einsum.@einsum` is not caught.)

I was sure this was going to be like 5 lines until I started! Maybe things like this should be quarantined somewhere outside the main file, and tested to death? Xref plugins #330 perhaps, maybe there's some other mechanism.

fonsp:
Could you make it more general, so that other macros from other packages can easily be added?

mcabbott:
I made a list of what Query.jl uses, as it was mentioned in #196. Without thinking too hard yet, or adding tests.

fonsp:
Looks like the tests broke?

mcabbott:
The pass locally, on 1.5, but the CI does look related:
```
Test Summary:          | Pass  Total
100
Communication protocol |   21     21
101
┌ Error: Expression explorer failed on: 
102
│   ex = :($(Expr(:toplevel, :(#= /tmp/Important blueprint.jl#==#cb26907e-f608-11ea-1455-553f5bc666d0:1 =#), :(html"a"))))
103
└ @ Pluto.ExpressionExplorer ~/work/Pluto.jl/Pluto.jl/src/analysis/ExpressionExplorer.jl:668
104
UndefVarError: MacroZoo not defined
105
Stacktrace:
106
 [1] explore!(::Expr, ::Pluto.ExpressionExplorer.ScopeState) at /home/runner/work/Pluto.jl/Pluto.jl/src/analysis/ExpressionExplorer.jl:341
107
```
That said, this should perhaps be WIP a little longer -- the list of Query.jl macros might be over-zealous, and has no tests. And not sure whether you want standard libraries like Tests mixed up in the zoo?

fonsp:
The error is:
```
 UndefVarError: MacroZoo not defined
```

The macros from `Test` all work already - most macros don't change code in a relevant way. The same for Query.jl - are you sure that we need mock versions of those? Feel free to write tests, but be sure to run those tests without any mock versions first. Less code is always better.

mcabbott:
But why the error on CI only, is the question? Am always confused by importing sub-modules, should it be `.MacroZoo` in more places? 

Re `@test`, this fixes #383, a keyword argument that previously read as an assignment. 

Re Query.jl, in #196 I pinged the guy who requested this for comment, I haven't used the package much.

fonsp:
Maybe you are testing the master branch by accident?

mcabbott:
It certainly fails locally when I edit the new tests & get it wrong. But is there anything non-standard about this package which I need to know? Like some other repository it needs for some tests, or something?

fonsp:
Instead of testing whether the mock macro creates the 'right' mock version, could you test whether the end result is recognized correctly? So do the same as all the other ExpressionExplorer tests, but for expressions that contain these special macros.

We don't care about _what_ the mock version is, we care about the end result.

mcabbott:
Like this?
```
@test testee(:(@test π ≈ 3.14 atol=0.01), [:π], [], [[Symbol("@test")], [:≈]], [])
```
I left some tests of just the mocking functions too, though, which might help future-me understand what those functions were aiming to produce.

fonsp:
Exactly

mcabbott:
I should probably have said, but that's the first line of `@testset "Macro Zoo"` right now. Should there be more?

mcabbott:
Bump.

Does this need something changed? Or do you think it's a hopeless approach & you'd like to wait for something else entirely?

fonsp:
It's just a shame to compromise, but I guess that this is a good temporary solution.

fonsp:
One comment: could you add a try catch block so that it doesn't fail if the mock macro didn't dispatch?

mcabbott:
Sure, something like this? 

Am not sure if `@warn` is what you want, I believe this will go to the terminal for debugging.

Issue #352, opened 2020-09-04
Vector(undef, 3) does not show


fonsp:
![image](https://user-images.githubusercontent.com/6933510/92229707-2de42980-eeaa-11ea-83ff-ff3ed14e5ea4.png)

but in the REPL

![image](https://user-images.githubusercontent.com/6933510/92229751-418f9000-eeaa-11ea-8ea8-b18d8c0f13f7.png)

fonsp:
Someone needs to figure out how Base Julia does it, i.e. go through what happens when you call

```julia
show(stdout, MIME"text/plain"(), Vector(undef, 3))
```

shashankp:
displaying data does a loop, which is failing; MVE:

![image](https://user-images.githubusercontent.com/498813/92242468-6fda9300-eedd-11ea-9a22-33c31c1ad1d4.png)

fonsp:
I see why Pluto errors, but why wonder why `show(stdout, MIME"text/plain"(), Vector(undef, 3))` _does not error_.

shashankp:
works fine cos it goes to [base implementation](https://github.com/JuliaLang/julia/blob/f896006621414725ac1355fe23a779c50995091d/base/show.jl) which has checks for isassigned & isdefined. As expected, works fine in REPL too
![image](https://user-images.githubusercontent.com/498813/92304411-3ff1c500-ef9b-11ea-865a-8b08ed0e71d2.png)

where as
![image](https://user-images.githubusercontent.com/498813/92305000-b513c900-efa0-11ea-8ac7-7b8526897b88.png)
goes to
`show(io::IO, ::MIME"application/vnd.pluto.tree+xml", x::AbstractArray{<:Any, 1})` [ref](https://github.com/fonsp/Pluto.jl/blob/d943b3863a611e3a64db4534488dc63924504022/src/runner/PlutoRunner.jl#L394)

May be it needs isassigned/isdefined checks?

fonsp:
That sounds good!

shashankp:
this should help with https://github.com/fonsp/Pluto.jl/issues/354 too

fonsp:
No that's unrelated I think

fonsp:
Fixed by @Moelf ! Yayo

Testset is here:
https://github.com/fonsp/Pluto.jl/pull/411/files#diff-57577b94ab5dc9eb946ef0662a875f1dR24

fonsp:
@Moelf 

[...]

PR #411, opened 2020-09-11
fix undef showing

Moelf:
main issue was that error happens at assign:
https://github.com/fonsp/Pluto.jl/blob/b53d282a132752e34563593b8c7c294866042efb/src/runner/PlutoRunner.jl#L372

so instead of broadcast over `zip` of (index, elements), a specialized function for array is implemented.

Also, I think a lot can be re-used from `Base`, for example:
```
julia> Base.print_matrix(stdout, vvv)
 #undef
 #undef
 #undef
```
Maybe for the next re-factor :construction_worker:

Also I switched to `@view` when possible, since users can have huge array (maybe each element is big or something), so slicing with copy was not as good of an idea (I think).

After this patch:
image

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -375,6 +375,19 @@ function show_array_row(io::IO, pair::Tuple)
     print(io, "</v></r>")
 end
 
+function show_array_row(io::IO, idxs::AbstractVector{<:Integer}, ary::AbstractArray{<:Any, 1})
+    for i in idxs
+        if isassigned(ary, i)
+            show_array_row(io, (i, ary[i]))
+        else
+            # such that #undef is not confused with "#undef"
+            print(io, "<r><k>", i, "</k><v>")
+            print(io, Base.undef_ref_str)
+            print(io, "</v></r>")
+        end
+    end
+end
+
 function show_dict_row(io::IO, pair::Union{Pair,Tuple})
     k, el = pair
     print(io, "<r><k>")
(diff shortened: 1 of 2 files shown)


Moelf:
fix #352

fonsp:
Nice! Could you add a testset?

https://github.com/fonsp/Pluto.jl/blob/master/test/RichOutput.jl

Moelf:
@fonsp done

src/runner/PlutoRunner.jl
```diff
@@ -375,6 +375,19 @@ function show_array_row(io::IO, pair::Tuple)
     print(io, "</v></r>")
 end
 
+function show_array_row(io::IO, idxs::AbstractVector{<:Integer}, ary::AbstractArray{<:Any, 1})
```
fonsp, line 378: This should be a new function: it does not show a `row`.

src/runner/PlutoRunner.jl
```diff
@@ -375,6 +375,19 @@ function show_array_row(io::IO, pair::Tuple)
...
+    for i in idxs
+        if isassigned(ary, i)
+            show_array_row(io, (i, ary[i]))
+        else
+            # such that #undef is not confused with "#undef"
+            print(io, "<r><k>", i, "</k><v>")
+            print(io, Base.undef_ref_str)
+            print(io, "</v></r>")
```
fonsp, line 386:
```julia
show_array_row(io, (i, Text(Base.undef_ref_str)))
```

fonsp (changes requested):

src/runner/PlutoRunner.jl
Moelf: I can rename it to "rows" but that's too subtle of a difference?

fonsp:
Can you remove abbreviations? idx, ary, els

Moelf:
> Can you remove abbreviations? idx, ary, els

yes, where and why? I mean we used `i, el = pair` before without a problem right?

fonsp:
Awesome, thanks!

PR #419, opened 2020-09-12
allow up/down arrow between cells

marius311:
Two things here: 

* When you're at the top/bottom of a cell, if you hit up/down it takes you to the previous/next cell, at the same column, exactly as if the whole notebook was just one single big document. This is basically the current behavior of Del/Backspace, so I think this fits nicely into that mental model, and also pretty convenient for keyboard-only use.

* Moving up/down by a cell using Pg Up/Down is switched to always putting the cursor at the top of the new cell. This was already the PgDown behavior, but PgUp used to put it at the end of the last line. I think the current behavior is better since it actually scrolls the top of the cell into view and if you're browsing across many cells, which I think is a fair assumption if you've moved your hand to Pg Up/Down, you are probably looking to read them from the beginning.

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -78,10 +78,10 @@ export const CellInput = ({
         // these should be fn+Up and fn+Down on recent apple keyboards
         // please confirm and change this comment <3
         keys["PageUp"] = () => {
-            on_focus_neighbor(cell_id, -1)
+            on_focus_neighbor(cell_id, -1, 0, 0)
         }
         keys["PageDown"] = () => {
-            on_focus_neighbor(cell_id, +1)
+            on_focus_neighbor(cell_id, +1, 0, 0)
         }
         keys["Shift-Tab"] = "indentLess"
         keys["Tab"] = on_tab_key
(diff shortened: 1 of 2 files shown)


fonsp:
Thanks! This is the direction that we were going in.

On the other hand, I am noticing some problems with backspace delete. I can elaborate, but how do you feel about the feature?

fonsp:
Also would you like to video call sometime? You seem to have good ideas :)
fonsvdplas@gmail.com

marius311:
> how do you feel about the feature?

About backspace/delete? I like it alot actually. What's the issue?

And happy to chat if you think it'd be useful, tbh I really don't have much free time to contribute much here, but the little PR here/there has been fun, will email you!

fonsp:
Here's a conversation from last week on the Pluto slack about the pros/cons/future of the backspace feature:
https://gist.github.com/fonsp/0911943f27baab81b45c53b1c95044d2

marius311:
Thanks, helpful to follow the discussion, I may comment there. This PR seems independent of what happens to backspace/del and could be merged independently though?

PR #425, opened 2020-09-14
On binder, put the token in the url so it can be shared

fonsp:


fonsp:
💚💖💞💟

fonsp:
@dralletje

dralletje:
Beautiful

fonsp:
#429

fonsp:
oepsie

PR #434, opened 2020-09-14
Consider cmd key for shortcuts on mac

dralletje:
Keyboard shortcuts are harder than I imagined at the start actually.
Not everything just works if we change all "Ctrl" appearances to "Cmd", because of some colliding shortcuts.
So I guess it's best to allow both Ctrl **and** Cmd in shortcuts on mac 🤷‍♀️


fonsp:
smart

PR #437, opened 2020-09-15
Logging in notebook

fonsp:
https://user-images.githubusercontent.com/6933510/113442696-d3378f80-93f0-11eb-85ad-c2c01e96b413.mov

Adds support for [Julia logging](https://docs.julialang.org/en/v1/stdlib/Logging/): `@info`, `@warn`, `@debug`, `@error` and custom log messages.

(This does _not_ relay the terminal - that is a different subject.)

# TODO

- [x] ~The log should "pop out" of the `overflow:hidden` box when you click on it.~
- [X] Virtualized logs (only load a viewport worth of logs into the DOM)
- [x] layout might not work if the cell has many lines but cell input is hidden
- [x] ~Selecting cells is difficult if there are also logs~
- [X] Option to hide logs (Future TODO: save that propery in a comment @dralletje 
- [x] Narrow-screen layout still needs improvement
- [x] show kwargs
- [x] rich output
- [x] ~use a lighter background for the kwargs~
- [x] State diffing becomes slow when there are lots of logs
- [x] Don't also print logs to terminal?
- [ ] Limit a crazy number of logs?
- [x] Dark mode

## Test notebook

https://github.com/fonsp/disorganised-mess/blob/main/logging.jl

# Try it out!

```julia
julia> ]
pkg> activate --temp
pkg> add Pluto#logging
```


fonsp:
image

fonsp:
image

src/runner/PlutoRunner.jl
```diff
@@ -683,4 +684,36 @@ const fake_bind = """macro bind(def, element)
...
+        # id=id,
+        file=file,
+        line=line,
+        kwargs=Dict((k=>to_pluto_html(v) for (k,v) in kwargs)...),
+    ))
+end
+
+# we put this in __init__ to fix a world age problem
+function __init__()
+    Logging.global_logger(PlutoLogger())
```
c42f, line 716: It may be better just to use `Logging.with_logger()` in some context which surrounds the user's code?

c42f:
This is super cool!

Let me know if there's any help you'd like in integrating this, or subtle questions about the logging system. I'm also happy to do a call if that's helpful? (Note I am in AEST=UTC+10)

I can help out with the Julia side of the code if necessary. Figuring out the most correct way to format the log messages might be the hard bit.

fonsp:
Right now the messages only appear if you had the notebook open while the logs happened, so the logs are lost when you refresh. Instead, logs need to be stored in an array inside the cell's _state_.

Our current state management is not very flexible - implementing this means adding a couple of new message types, handlers and state updates, and some careful thinking to ensure that the state is never corrupt. We were planning to update the state management system soon, so I think it would be nicer to wait for that.

c42f:
Sure, no problem at all.

What are your thoughts for UI when log records arise from code defined outside Pluto, or from functions defined in other cells?

I can think of a couple of options:
* Dynamically assign them to the line from which they originate (Note that this requires taking a stack trace for every message which will appear in the UI, which can be done by the logging backend. Could be expensive; though UI elements are also expensive so we might be ok :-) )
* Plot them on a line separate from the code lines of the cell. This can be done with the `file` and `line` keys that every log record comes with.

I kinda favor trying the dynamic option and seeing how it pans out.

fonsp:
The dynamical approach is a good idea! Could you show me how to get the stacktrace from a log?

Before your idea, I was thinking that every non-script log message could appear as a vertical line, since we know it's temporal position, but not the line number. Something like this:
image

The log would show in the cell that was most recently running.

Finding the backtraced line number is better of course, but we might still need this as a fallback? For async external code for example? Or can those also be backtraced?

c42f:
> Could you show me how to get the stacktrace from a log?

The idea would be to add a call to `backtrace()` from within `handle_message(::PlutoLogger, ...)` — this runs as a child function of the user's `@info` so the backtrace is directly accessible.  From that backtrace, extract any info necessary and attach it to the log record as metadata before passing it to the frontend.

For async tasks, note that loggers are inherited from the parent task if `Logging.with_logger` was used. So as long as you install `PlutoLogger` before creating those tasks, the same considerations will apply. (Or if you're using `global_logger` that will be the fallback for any tasks which don't have a local logger installed.)

LinusSch:
Quick thought on the UI: I would prefer the dots from the same iteration of the loop to be vertically aligned. That way there is a little bit more information, directly visible.

I guess this idea breaks the case where multiple logging events happen on a single line of a single iteration of the loop, but this could be solved by having two or three sectors with different colors as needed. Exactly how many logging events are happening at a single point in the code is rarely as useful as quickly realizing whether several events happening in a loop are happening on the same or on different iterations (in my experience, which is more limited than yours, so feel free to correct me if I'm wrong).

If this idea can work well or not probably depends on how you show the details of each event, that view may become unwieldy in cases where many events get bundled into a single dot. If so then the usefulness may not be worth the drawback.

fonsp:
image

fonsp:
Support for variable line heights:

https://user-images.githubusercontent.com/6933510/114543864-15b36480-9c5a-11eb-88c6-d1bd04712e4c.mov

fonsp:
Schermafbeelding 2021-04-13 om 14 25 21

https://user-images.githubusercontent.com/6933510/114552046-6fb92780-9c64-11eb-9a7c-24d9ca8cae7d.mov

lungben:
I think this would be a very valuable addition!

It would be good to show "normal" console output (e.g. by `print`) in the same way.

fonsp:
@pankgeorg Any idea how we can get this, or a simplified version of this, merged soon?

pankgeorg:
> @pankgeorg Any idea how we can get this, or a simplified version of this, merged soon?

In terms of the CM6 migration, this looks cool now. Let's make a small mergelist!

fonsp:
@dralletje Can you take a look at the state diffing performance problem in this PR? You had a solution right?

frontend/editor.css
```diff
@@ -2149,6 +2170,213 @@ nav#undo_delete.hidden {
...
+    position: absolute;
+    z-index: 25;
+    right: min(-100px, -100vw + (700px + 25px));
+    top: 1px;
+    overflow-x: auto;
+    /* overflow-y: hidden; */
+    background: hsl(0deg 0% 100% / 81%);
+}
+
+/* @media screen and (max-width: calc(700px + 25px + 6px + 150px)) { */
```
pankgeorg, line 2195: Is this commented on purpose? @fonsp

fonsp:
@dralletje Your recent commits (0e5f8ce and cc045d3) did the trick! 

<details>
<summary>Benchmarks</summary>


Before:
Schermafbeelding 2022-01-17 om 20 36 50
After:
Schermafbeelding 2022-01-17 om 20 41 15

</details>


Script:

https://gist.github.com/fonsp/9551c8688f05b4baf21d3f242696dfcc

PR #440, opened 2020-09-15
🥫 Fix binder connection issues

fonsp:


fonsp:
@dralletje i think this fixes the screenshot that i sent you. It also now pings manually to keep the connection alive in binder/heroku

Issue #229, opened 2020-07-24
Cells with resource heavy code 


ixzh:
It is more of a personal problem: 

I have a cell which reads 400+ of text files[10k lines per file on average] into array of strings, and another cell parsing them into array of dataframes. These two cells will take quite some time to run (1 mins to read, 10+ mins to parse into tables). Usually I will just wait for them [and the compiling overheads] to complete before I can start work.

The problem is the cells depend on functions in other cells [Reference Cell?]. When I need to change the functions, or sometimes just the comments inside functions, it will trigger these two cells to re-run.

I only need to read and parse the strings once. No change to the read data required.

What is the right way to prevent such cells from rerunning but generating the same data? Thanks!

fonsp:
Hey! We will answer this question next week during our JuliaCon talk, is it okay to wait until then?

You can register for JuliaCon for free (do it!) or watch the talks on youtube after the week.

fonsp:
See also #13 and #116

evadknarf:
haha,  can't wait! See you there!

evadknarf:
Back to this issue again, as I've been haunted by a few cells which took too long to run...

I tried to @save (from JLD2) all variables to prevent parsing strings into DataFrames, but the macro is ignored... I have to read in the data as strings and parse them again each time restarting julia. (as I move around a lot)

Any suggestion to resolve resource heavy cells in a notebook?

fonsp:
Did you trigger a re-run without knowing about it, or do you want to disable reactivity?

evadknarf:
I know the cell will be triggle by some action like adjusting output format of its dependant cells. Even first time loading is unavoidable.  would appreciate some intelligence in limiting rerun the same results,  and macro support to load variables from JLD2 file to eradicate the problem?

karlwessel:
Would it be possible for Pluto to detect if only comments in a cell have changed and recognise that it doesn't have to rerun the code? That could reduce unnecessary reruns in some cases.

Otherwise I can only recommend to thoroughly test resource heavy cells with small datasets in the beginning until you are pretty sure that you don't have to change them to often later when using the real dataset.

evadknarf:
> test resource heavy cells with small datasets in the beginning until you are pretty sure that you don't have to change them

not only in the current cell, but also upstream dependant cells. It's hard to make such avoidance upfront.

fonsp:
#298

Issue #116, opened 2020-05-17
Automatic timeout for cells to enforce reactivity


karlwessel:
This is an idea I would like to discuss.

The intention of Pluto, as far as I understood it, is to be an reactive notebook where you get more or less instant feedback to changes so you immediately see the  consequences of them.

If you have a cell that takes a long time to execute it really breaks the whole reactivity and collides with automatic reevaluation of cells. And often the only solution for long running cells seems to be to stop/kill Pluto and restart it (and julia).

I wonder therefore if it would be beneficial to consider long running cells (which are not compiling) an implementation error against which to guard the notebook per default, since most times they really are. For example when one underestimated the complexity of a method or messed up some loop ending condition. 

A possible guard against that could be a really low (like 0.5 to 5 seconds) automatic timeout for each cell after which execution is stopped and an error is thrown telling the user to either reduce the load for that method or maybe set an explicit longer timeout for that cell in case it really needs to take that long.

A possible Problem could be compilation times, they should not count into the timeout.

Even cooler would be a mechanism by which the user can implement a cell so that it can return a partial result after the timeout instead of throwing an error.

fonsp:
Cool idea, long-running cells is something that we should do something about! 

I'm not sure about automatically interrupting, this would lead to bad things when importing a package, writing to a file, downloading a file, and other code with side effects. 

So I would say that Pluto should prompt the user whether they want to stop a cell that is taking too long. Would you agree? And have you used the stop button that's currently implemented? You shouldn't have to shut down Pluto completely.

A common worry about Pluto is that (intentionally) long-running code will "not work in Pluto" or will be re-evaluated when they don't want to. I'm not sure what to say about the first one, but the second one could be solved with more informative UI.

My idea for this was to warn users when one of the downstream cells is _expected_ to have a long runtime, based on its previous runtime, and prompt whether they still want to run a cell. This would be useful in cases where you are changing a parameter, without realising that your long-running code depends on that parameter. In that case, you might want to do something different, like define a new parameter.

karlwessel:
You are right, automatic interruption would probably be to risky (also in the sense that it might make users just angry :-). Some prompt or warning might be good, but could lead also to the problem that users get tired of having to click away warnings or prompts all the time...

[...]

PR #442, opened 2020-09-16
Fix 'undefineds' time unit for long running cells

isofew:
Hi, I found that when a cell runs for more than 1000 seconds, the displayed time unit will become 'undefineds'. (See the following screenshot)

image

It seems to be due to an index error in ` frontend/components/RunArea.js` I fixed the loop index bound so it won't further increase when already at the largest time unit (seconds)

### DIFF
--- a/frontend/components/RunArea.js
+++ b/frontend/components/RunArea.js
@@ -15,7 +15,7 @@ const prettytime = (time_ns) => {
     }
     const prefices = ["n", "μ", "m", ""]
     let i = 0
-    while (i < prefices.length && time_ns >= 1000.0) {
+    while (i < prefices.length - 1 && time_ns >= 1000.0) {
         i += 1
         time_ns /= 1000
     }


fonsp:
Yay! Thank you for debugging and fixing this yourself! I guess it was trying show `1.2 kilosecond` :)

And cool to hear that you are doing long running computations with Pluto! Maybe you would like to comment on your experience here? #116 or #229

isofew:
> Yay! Thank you for debugging and fixing this yourself! I guess it was trying show `1.2 kilosecond` :)
> 
> And cool to hear that you are doing long running computations with Pluto! Maybe you would like to comment on your experience here? #116 or #229

Thanks for the reference, I'll check out the threads and see how can I contribute there.

By the way I also noticed another bug related to long running cells: when the cell finished, the bar on its left is still in a running state and the time measurement won't display. I know it's finished because all dependent cells returned and I can interact normally, see example below:

Screenshot 2020-09-16 at 18 58 29

Do you have some idea what the problem might be?

fonsp:
Is this after your fix?

isofew:
> Is this after your fix?

No it's before the fix.

Issue #307, opened 2020-08-18
Table of contents / Automatic enumeration of Markdown headers?


Datseris:
Hi there,

I use Jupyter (specifically Jupyter Lab) to make project reports every 3 months or so. As a result, the reports get quite large. I structure them in detail with markdown headers and sections. Jupyter provides an excellent extension, called simply "table of contents" that does two things:

1. Enables a table of contents sidebar that is **clickable** and brings you to the markdown header that you clicked (very useful)
2. Optionally enumerates all markdown headers with 1., 1.1, 1.2.1, etc. (nice to have, but not as crucial as the above).

Example: 

![image](https://user-images.githubusercontent.com/19669089/90527488-e2274580-e171-11ea-9df5-cd5fa23e3b9a.png)

Is this possible to be done in Pluto.jl? Is it planned?

fonsp:
For anyone who wants to take this on:

This can be done entirely inside a Pluto notebook - we don't need to modify Pluto itself. I think this would should be a `PlutoUI.TableOfContents()`.

Try this:

```julia
md"# Hello"
```

```julia
md"# World!"
```

```julia
html"""
<script>
const headers = Array.from(
	document.querySelectorAll(
		"pluto-notebook pluto-output h1"
	)
)

return html`<ul>${headers.map(h1 => 
			html`<li>${h1.innerText}</li>`
	)}</ul>`

</script>
"""
```

Up to you to figure out how `MutationObserver` works. If needed we can add a custom `cellOutputChanged` event to `document`.

Datseris:
Notice that in my example headers of any level get linked, not just `h1` (just in case h1 means header level)

shashankp:
pluto-cell id can be used to link

```
html"""
<script>
const headers = Array.from(
	document.querySelectorAll(
		"pluto-notebook pluto-output h1"
	)
)

var parentCellId = function(el) {
    while (el.nodeName != 'PLUTO-CELL') {
        el = el.parentNode;
        if (!el) return null;
    }
    return el.id;
}

return html`<ul>${headers.map(h1 => html`<li>
	<a style="text-decoration: none" href="#${parentCellId(h1)}">${h1.innerText}</a>	 </li>`
	)}</ul>`

</script>
"""
```

![image](https://user-images.githubusercontent.com/498813/92567042-bd1e8200-f29a-11ea-8bdd-ef65532ee483.png)


But I guess this is not reactive as its html
[...]

PR #445, opened 2020-09-16
Cell output changed event

shashankp:
Companion to https://github.com/fonsp/PlutoUI.jl/pull/30
Needed for https://github.com/fonsp/Pluto.jl/issues/307
Uses window level CustomEvents
to trigger rendering of TableOfContentsElement after html elements are rendered.

_Note: doesnt work for cell delete scenario_

### DIFF
--- a/frontend/components/Cell.js
+++ b/frontend/components/Cell.js
@@ -76,8 +76,7 @@ export const Cell = ({
     errored,
     output,
     selected,
-    on_change,
-    on_update,    
+    on_change,  
     on_update_doc_query,
     on_focus_neighbor,
     disable_input,
@@ -90,6 +89,18 @@ export const Cell = ({
 }) => {
     // cm_forced_focus is null, except when a line needs to be highlighted because it is part of a stack trace
     const [cm_forced_focus, set_cm_forced_focus] = useState(null)
+    
+    const cell_ref = useRef(cell_id);
+    const on_output_changed = () => {
+        cell_ref.current.dispatchEvent(
+            new CustomEvent("cell_output_changed", {
+                bubbles: true,
+                detail: {
+                    "cell_id": cell_id
+                }
+            })
+        )
+    }
 
     useEffect(() => {
         const focusListener = (e) => {
@@ -122,6 +133,7 @@ export const Cell = ({
                 code_folded: code_folded && cm_forced_focus == null,
             })}
             id=${cell_id}
+            ref=${cell_ref}
         >
             <pluto-shoulder draggable="true" title="Drag to move cell">
                 <button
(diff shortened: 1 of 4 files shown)


fonsp:
We don't need to pass callbacks from the Editor to the RawHTMLContainer, the event can be dispatched directly by the RwHTMLContainer. Also, it looks like the event is never fired?

dralletje:
I agree with @fonsp.
I do understand the way that the callback has to travel up the tree to get a reference to `cell`, but I like the idea of going with vanilla javascript fully.

I'd say we do the dispatching from the Cell component, so we can still add the `cell` as detail, but it's not in the oh-so-big Editor.js file.

Also then we can dispatch the event from the actually pluto html node 
```
let cell_element_ref = useRef();
<pluto-cell
  ref=${cell_element_ref}
  on_render=${() => cell_element_ref.dispatchEvent( ... )}
```
this feels a lot cleaner and allows for a lot more control (the handler can get the actual cell element, stuff with bubbling?)

fonsp:
Giving the entire `cell` as event detail is nice, but on the other hand, we don't want people to depend on its structure if we want to change it later. (I do)

dralletje:
I'm working on a small rewrite of the html rendering stuff, and I'll put  this in as well. Is there a way to add someone as a collaborator to a commit? Feels bad to take part of your code and erase your name from it 😂 @shashankp

shashankp:
@dralletje can do git gymnastics with multiple remotes, but I'd say don't worry: more commits coming from me 😁

fonsp:
The `cell_output_changed` event is never fired, right?

I think we should not have an `on_delete` event, because it is only useful if the JS code that depends on that event is keeping an internal notebook state. We should design our API such that this is not necessary.

A better option is to fire the same `cell_output_changed` event after setting the `innerHTML` to `""`. We can think about custom events for notebook state updates later.

fonsp:
Another comment: do we need to fire this event ourselves? Is the built-in [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) not right for this?

shashankp:
> The `cell_output_changed` event is never fired, right?

It does get fired in RawHtmlContainer > execute_scripttags 

> I think we should not have an `on_delete` event, because it is only useful if the JS code that depends on that event is keeping an internal notebook state. We should design our API such that this is not necessary.

> Another comment: do we need to fire this event ourselves? Is the built-in [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) not right for this?

Haven't looked into this, will check it 👍
hopefully it will help with cell rearrange

dralletje:
MutationObserver is too low level for this stuff, also don't think you can get this behavior specifically, or it would break really quick when we change anything about the html structure.

shashankp:
I see if MutationObserver is used, then the TableOfContents element needs to listen to changes on notebook (or only markdown cells?) & this PR isn't needed. But having all these events being sent to it might effect performance in cases of live data changes like a slider.

dralletje:
I made it work with only a mutation observer. Let's see if that works first 😁
I'll close this PR in the meantime.

PR #446, opened 2020-09-17
Make sure cells don't use top-level return

dralletje:
A tiny wrapper in the timed_expr/trycatch_expr - wanted to make it more sound logicy (separate functions composed) but that felt like it was going to make it more complex.
So I find this suffices for now :D


fonsp:
Can you add a test?

dralletje:
@fonsp tests added

dralletje:
Lekker

dralletje:
Aaait

PR #462, opened 2020-09-20
add unamed parameter type to references, fixes #423

### LINKED ISSUE #423: Methods with unnamed arguments do not re-evaluate
When we define a method with unnamed arguments like this
```julia
f(::Example) = 100
```
if the `Example` structure is modified, the cell defining `f(::Example)` does not re-evaluate.

Yet, if we define the method like this
```julia
f(var::Example) = 100
```
it works properly and this cell will update when `Example` is modified.

(Also, I have to say, using Pluto for teaching has been incredible—amazing work with this package 💕)

Pangoraw:
Hi,

This is a fix for issue #423 where methods with unnamed arguments where not added as referencers to the parameters' type. I fixed it by adding a special case in the `explore_funcdef` function.

This is my first try at touching a serious julia codebase so I will be happy to fix any comments

Thanks :balloon:

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -576,6 +576,13 @@ function explore_funcdef!(ex::Expr, scopestate::ScopeState)::Tuple{FuncName,Symb
         return mapfoldl(a -> explore_funcdef!(a, scopestate), union!, ex.args[2:end], init=(name, symstate))
 
     elseif ex.head == :(::) || ex.head == :kw || ex.head == :(=)
+        # account for unnamed params, like in f(::Example) = 1
+        if ex.head == :(::) && length(ex.args) == 1
+            symstate = explore!(ex.args[1], scopestate)
+
+            return Symbol[], symstate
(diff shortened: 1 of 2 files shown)


fonsp:
Very nice! Thank you!

fonsp:
I've added some tricky tests, let's see if they pass!

Pangoraw:
One is failing, I will take a look

fonsp:
Thanks! If it's not simple to fix, then we will just merge this PR, and add the failing test as an issue for later.

Pangoraw:
Alright, it did not seem related. The `...` params were send back to `explore!` (as a default for unknown expressions) were they would be treated as traditional reference to the variable. I added a case to "unwrap" the `...` and recurse on it.

Should i make another commit or keep it in the same ?

fonsp:
Awesome, thanks again! You're right that this was a separate issue, so I'm happy that you fixed this too 😊 

I added some more tests for the splatting operator (`...`), they already passed locally so it's probably all good!

PR #466, opened 2020-09-21
📋 Copy/paste functionality for cells

### LINKED ISSUE #399: Copy paste entire cells
@ekzhang

Right now you can select multiple/single cells, but you can't copy paste them:

selectcells

It needs to work among different notebooks! And if possible, different notebooks from different sessions (e.g. one on binder, one running locally).

Maybe you could think about the contents of the clipboard? Some options:

1. It doesn't matter, you put the data in localStorage. But then it only works within the same session
1. The copy data is exactly like it is in the file (i.e. # |==\ncode\n\n ish). And when you paste onto the page it recognizes that format (and generates new cell IDs to avoid duplicates). This is cool, because people often post notebooks in that format on github.
1. The copy data also contains formatted output? Like what happens right now if you text-select and copy+paste to Google Docs.

Another thing: for this you would want to place a superbig text cursor after a cell or between two cells, before you hit Ctrl+V. I think this could be really cool, if you also allow text input to create a new cell.

ekzhang:
# Copy-Paste Functionality

This closes #399 by adding copy/paste functionality for entire cells.

The implementation is **unfinished**. Current progress:

- [x] Copy-paste multiple cells / entire selection
- [x] Enable copy-paste between multiple notebooks, including on different servers
- [x] Enable copy-paste from Julia (.jl) file to notebook
- [x] Enable copy-paste from notebook to Julia (.jl) file
- ~Add "big cursors" to paste in any location in the notebook~
- ~Include outputs, paste to word processor with images~
- [x] Scroll pasted cells into view
- [x] Ctrl-X functionality (not sure if this is possible)
- [x] Drag-and-drop between multiple notebooks (also not sure if this is possible)

## Demo

Would appreciate any feedback about how this might work better!


ekzhang:
Just added a commit that scrolls the document when you're making a selection, and your mouse reaches the upper or lower 30% of the page. This is consistent with the scrolling behavior when you're dragging!

Next up I'll implement drag-and-drop between multiple notebooks (should be possible after this refactoring).

ekzhang:
### Drag and Drop



@fonsp Happy to receive your comments now!

ekzhang:
(Cutting multiple cells also is affected by #482)

fonsp:
I made the selectionarea only scroll in the direction that you are selecting in, and clicking clears the selection.

fonsp:
Could you try to fix this?

91f968d4-4135-4e45-8a05-5503c008d482

I did it on purpose for the GIF but I think that this is common when dragging cells around, since the codemirror is a large hitbox

ekzhang:
What would be your suggestion for the intended behavior in this case?

karlwessel:
> What would be your suggestion for the intended behavior in this case?

At the moment Pluto indicates that it will drop the cell after the one where the drag ended. If that is the same cell being dragged, like in fonsp example above, the drag and drop doesn't change anything. Which seems reasonable to me.

ekzhang:
Oh, gotcha, so the point is that we shouldn't be able to drag cells as text into CodeMirror.

I'll try giving the cells a different MIME type than text/plain.

ekzhang:
Changing the MIME type to `text/pluto-cell` did the trick, just as expected.

fonsp:
Are cell uuids still globally unique after copying and dragging between notebooks?

ekzhang:
Yes, cell UUIDs are regenerated.

PR #467, opened 2020-09-21
handle try catch in different scopes, closes #464

### LINKED ISSUE #464: `try` block is not scoped correctly
```julia
try
	x = 1
catch
end
```
```julia
try
	x = 1
catch
end
```
gives a mulitple defs error.

Pangoraw:
Hi,

I added a case to handle `try`/`catch` expressions as two blocks with their own scopes. This prevents the error happening in #464.

We start by exploring the `catch` block to deal with the case where a global reference could be shadowed with the symbol for the catched error:
```julia
a = 1
```
```julia
try
   do_something()
catch a
   report(a)
end
```

We then explore the `try` block as usual.

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -112,6 +112,14 @@ Base.push!(x::Set) = x
 # HELPER FUNCTIONS
 ###
 
+function explore_inner_scoped(ex::Expr, scopestate::ScopeState)::SymbolsState
+    # Because we are entering a new scope, we create a copy of the current scope state, and run it through the expressions.
+    innerscopestate = deepcopy(scopestate)
+    innerscopestate.inglobalscope = false
+
+    return mapfoldl(a -> explore!(a, innerscopestate), union!, ex.args, init=SymbolsState())
+end
+
 # from the source code: https://github.com/JuliaLang/julia/blob/master/src/julia-parser.scm#L9
 const modifiers = [:(+=), :(-=), :(*=), :(/=), :(//=), :(^=), :(÷=), :(%=), :(<<=), :(>>=), :(>>>=), :(&=), :(⊻=), :(≔), :(⩴), :(≕)]
 const modifiers_dotprefixed = [Symbol('.' * String(m)) for m in modifiers]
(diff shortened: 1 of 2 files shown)


fonsp:
Great, thanks again! Can you add tests with `finally`? And one with `finally` but without `catch`?

Pangoraw:
I had not accounted for `finally` :facepalm: It should be all set now.

> That's allowed, right?

It look like [it is](https://docs.julialang.org/en/v1/manual/control-flow/#finally-Clauses), I added tests for this cases. 

Since I had a bit of code repetition, I added a helper function to explore the code in a new scope.

src/analysis/ExpressionExplorer.jl
```diff
@@ -112,6 +112,14 @@ Base.push!(x::Set) = x
 # HELPER FUNCTIONS
 ###
 
+function explore_new_scope(ex::Expr, scopestate::ScopeState)::SymbolsState
```
fonsp, line 115: rename to explore_inner_scoped

src/analysis/ExpressionExplorer.jl
```diff
@@ -112,6 +112,14 @@ Base.push!(x::Set) = x
...
+    # Because we are entering a new scope, we create a copy of the current scope state, and run it through the expressions.
+    innerscopestate = deepcopy(scopestate)
+    innerscopestate.inglobalscope = false
+
+    return explore!(Expr(:block, ex), innerscopestate)
```
fonsp, line 120: why create a `:block`?

src/analysis/ExpressionExplorer.jl
```diff
@@ -112,6 +112,14 @@ Base.push!(x::Set) = x
...
+end
```
fonsp, line 121: This pattern is used in lots of other places. Maybe you could also replace those with a call to `explore_inner_scoped`?

fonsp (changes requested):

src/analysis/ExpressionExplorer.jl
Pangoraw: I replaced it for `let`, `for` and `while` :+1: However, I could not use it on the [function declaration](https://github.com/fonsp/Pluto.jl/blob/master/src/analysis/ExpressionExplorer.jl#L386-L391) where it seemed different enough.

fonsp:
Awesome, thanks!!

PR #469, opened 2020-09-21
Comments using [ctrl-/] match the line indentation #️⃣

mossr:
Instead of commenting like this:
```julia
function example()
    for i in 1:10
#       some_code = i^10
    end
end
```
we'd get this:
```julia
function example()
    for i in 1:10
        # some_code = i^10
    end
end
```

For other `toggleComment` options see: https://github.com/ficristo/codemirror-addon-toggle-comment


dralletje:
I like this a lot hahaha, was wondering why it wasn't doing this

Issue #177, opened 2020-06-10
Support defining methods of a function in multiple cells


karlwessel:
The other solution is much harder to implement consistently - i.e. being able to detect multiple methods with the same argument types (instead of just detecting multiple functions with the same name) and deleting only the methods from one cell, while leaving the others. 

All of this is possible within the scope of Pluto, but not something for the next couple of months :)

_Originally posted by @fonsp in https://github.com/fonsp/Pluto.jl/issues/109#issuecomment-628060038_

karlwessel:
done, although maybe now the comments seem a bit weird :)

fonsp:
(split into two issues: #178)

fonsp:
It would be helpful if someone with an eye for edge cases (@karlwessel) would write the test cases for this _before_ thinking up a solution.

([Writing test cases is fairly easy,](https://github.com/fonsp/Pluto.jl/blob/master/test/React.jl#L202-L253) and they don't need to be perfect yet.)

fonsp:
Strictly speaking, two cells `f(x::Int) = 1` & `f(x::Int) = 2` should not be allowed. _However_, there are some very difficult situations like:

`f(x::Int) = 1` & `f(x::Int64) = 2`

`A = Number` & `f(x::A) = 1` & `f(x::Number) = 2`

`A = rand(Bool) ? Number : String` & `f(x::A) = 1` & `f(x::Number) = 2`

If all of these situations are simply ignored (we probably will), then a notebook like the first situation would be _ambiguous_  (ignoring the SECRET fact that the _order in which cells are placed_ is used as a tie-breaker when two cells are mutually independent.)

karlwessel:
Is it possible to detect those cases during execution of a notebook by comparing the `methodswith` count for each function for which there are cells defining methods for? And throw a runtime exception if one cell that should increase the count doesn't because it replaced another method?

fonsp:
That's going in the right direction! (Although my 5th example would not get detected with that method)

karlwessel:
Well, maybe we don't have to catch _all_ corner cases...

fonsp:
Building on your counting methods idea:

It's very nice if we can keep all of the syntax analysis in one place, _before_ running the code, and then run cells in a precomputed order. Checking the number of defined methods after running cells (in the middle of a reactive run), and then taking that information back to do something special makes that tricky. But there are three reasons to do this kind of "dynamic" code analysis instead of static anaylsis:

- this problem
[...]

PR #471, opened 2020-09-22
implemented tests towards solving issue #177

karlwessel:
Implementation of the tests from [fons' comment](https://github.com/fonsp/Pluto.jl/issues/177#issuecomment-645039993) towards implementing issue #177 .

All of them pass and ideally should still pass when issue #177 is implemented.

### DIFF
--- a/test/React.jl
+++ b/test/React.jl
@@ -280,6 +280,64 @@ import JSON
             @test topo_order.errable == Dict()
         end
     end
+    
+    @testset "Extended multiple assignments" begin
+        function testformultipleassignmenterror(def1, def2, prep=Cell(""))
+            notebook = Notebook([prep, def1, def2])
+            fakeclient.connected_notebook = notebook
+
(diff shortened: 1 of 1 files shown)


fonsp:
Why do they pass?

karlwessel:
Because they actually are true positive cases of `multiple definitions of the same method` and Pluto currently detects every method with the same name as a case of `multiple definitions of the same method`.

So Plutos current tactic is to find all true positives by just finding all positives, including the false ones :).

karlwessel (commented):
Why? When one of the definitions is deleted it is correct that there is no multiple definitions error anymore.

fonsp:
Thanks @karlwessel !

I merged your branch into a development branch: [`ab13217` (#538)](https://github.com/fonsp/Pluto.jl/pull/538/commits/ab13217d8e090d39745ea8aed2b61833ef43ef61)

and your tests are now inside Pluto!

PR #472, opened 2020-09-22
Automatically use Revise.jl in notebooks if available

lungben:
It is a common workflow to include Revise.jl in the `startup.jl` script to have Revise always available in the Julia REPL.
I think in Pluto it would be good to have Revise automatically activated if it is installed.
However, adding Revise as a direct dependency of Pluto is probably too much - many Pluto use cases do not benefit from Revise.

This little code change checks if Revise is available in the LOAD_PATH (e.g. in the global Julia environment). If yes, it adds `using Revise` to the commands automatically executed on creation of each workspace.

Tested with Revise v3.1.2.

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -97,7 +97,7 @@ export const CellInput = ({
         }
         keys["Ctrl-/"] = () => {
             const old_value = cm.getValue()
-            cm.toggleComment()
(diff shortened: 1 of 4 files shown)


fonsp:
Although this is a nice productivity hack, we will have to reverse this to make the built-in package manager (#142) work.

PR #478, opened 2020-09-23
handle macro definitions in ExpressionExplorer.jl, closes #477

### LINKED ISSUE #477: user defined macro are not reactive
If I write my own macro into a cell, cells calling it won't be noted as dependent. 

#### Example

```julia
macro a()
  return :(1 + 1)
end
```
```julia
@a()
```
:arrow_up: Here updating the definition for `a` wont update the bottom cell

Pangoraw:
This adds the handling of macro definitions almost like functions. The only difference is that it changes the macro name to be `Symbol("@name")` instead of just `:name` because calls to the macro will refer to `Symbol("@name")`.

#### Note
I am wondering if you can define macro with a name more complex than a single symbol :thinking: In this case, I need to consider a more robust solution.

Closes #477

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -379,7 +379,7 @@ function explore!(ex::Expr, scopestate::ScopeState)::SymbolsState
 
         # This is not strictly the normal form of a `for` but that's okay
         return explore!(Expr(:for, ex.args[2:end]..., ex.args[1]), scopestate)
-    elseif ex.head == :function || ex.head == :abstract
+    elseif ex.head == :function || ex.head == :macro || ex.head == :abstract
         symstate = SymbolsState()
         # Creates local scope
 
@@ -390,6 +390,12 @@ function explore!(ex::Expr, scopestate::ScopeState)::SymbolsState
         innerscopestate.inglobalscope = false
 
         funcname, innersymstate = explore_funcdef!(funcroot, innerscopestate)
+        # Macro are called using @funcname, but defined with funcname. We need to change that in our scopestate
+        if ex.head == :macro
+            setdiff!(innerscopestate.hiddenglobals, funcname)
+            funcname = Symbol[Symbol("@$(funcname[1])")]
+            push!(innerscopestate.hiddenglobals, funcname...)
+        end
 
         union!(innersymstate, explore!(Expr(:block, ex.args[2:end]...), innerscopestate))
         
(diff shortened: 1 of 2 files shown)


fonsp:
We need both names, `Symbol("@name")` and `:name`, because `@name` calls the macro, and `name` references the macro.

Pangoraw:
> We need both names, `Symbol("@name")` and `:name`, because `@name` calls the macro, and `name` references the macro.

I am not sure I understand, how can the symbol alone can be used ?
```julia
julia> macro a(); return :(1) end
@a (macro with 1 method)

julia> a
ERROR: UndefVarError: a not defined
```
Declaring `:name` would not allow the user to declare a macro and function with the same name (possible in the REPL)

```julia
julia> macro a(); return :("macro") end
@a (macro with 1 method)

julia> function a(); return "function" end
a (generic function with 1 method)

julia> a()
"function"

julia> @a
"macro"
```

fonsp:
Ah oops! Never mind then

fonsp:
About more complex names: I found this silliness:

```julia
julia> macro var"#  helloo (("(e)
           e
       end
@#  helloo (( (macro with 1 method)

julia> var"@#  helloo ((" 123
ERROR: syntax: extra token "123" after end of expression
Stacktrace:
 [1] top-level scope at none:1

julia> @var"#  helloo ((" 123
123
```

Pangoraw:
> I found this silliness

Good catch! After investigating, I think Julia has our back here:
```julia
julia> :(var" ")
Symbol(" ")

julia> :(@var" ").args[1]
Symbol("@ ")
```
I added tests for those cases as well :+1:

fonsp:
Thanks!

There were already some test cases about macros, can you put those in the same category? And can you test for argument types in macros?

Pangoraw:
Oops, fixed it :wrench: I wrote tests for:
- Simple macro def ([source](https://github.com/fonsp/Pluto.jl/pull/478/files#diff-2089b2016831147491e94b40cf551faaR297))
- Param types ([source](https://github.com/fonsp/Pluto.jl/pull/478/files#diff-2089b2016831147491e94b40cf551faaR300))
- Params types and default value ([source](https://github.com/fonsp/Pluto.jl/pull/478/files#diff-2089b2016831147491e94b40cf551faaR303))
- Global variable use in inner scope ([source](https://github.com/fonsp/Pluto.jl/pull/478/files#diff-2089b2016831147491e94b40cf551faaR306))

fonsp:
🌈🌈

Issue #291, opened 2020-08-14
Performance - delay creating codemirrors


fonsp:
When you load a notebook, it creates a codemirror for every cell. We could optimise this by only creating codemirrors for cells that are _not_ folded. 

Cells that are initially folded should only create the codemirror when it is first unfolded.

Maybe we can even delete codemirrors when they are not visible? Instead of the current `display: none`

novoselrok:
Additional enhancement would be to initialize code mirrors only after they come into the viewport (using IntersectionObserver API).

fonsp:
Solved by #484.

Strangely, the performance benefit is not noticeable, so let's not do the viewport thingy

PR #484, opened 2020-09-24
Adding and removing CodeMirror instance on-demand, i.e. depending on …

pupuis:
…the visibility of a cell - #291.

Hi all,

I took on the enhancement in #291 by only creating the CodeMirror instance if a cell is hidden (folded) and deleting the instance upon folding/hiding the cell. 

I introduced a new boolean called cm_enabled that is basically !is_hidden, but allows to later add more conditions (e.g. "is cell visible on screen" - that was probably meant by viewport, right?).

The general idea is then to, instead of as before running the effect hook that set up the CodeMirror only with initial values, re-run the effect hook everytime cm_enabled changes and to wrap everything else that requires the instance into an IF that checks for cm_enabled.

This is my first contribution to this project and I hope you like it - if you have any suggestions or concerns, I'd be happy to discuss. 

If this first step is done, I'd be happy to also take on the viewport-idea.

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -285,13 +285,6 @@ export const CellInput = ({
         }
     }, [cm_forced_focus])
 
-    // if the CodeMirror initialized/changed while it was hidden, and it suddely became unhidden, we need to refresh it to fix a layout bug where the gutter takes no horizontal space.
-    useLayoutEffect(() => {
-        if (!is_hidden) {
-            cm_ref.current && cm_ref.current.refresh()
-        }
-    }, [is_hidden])
-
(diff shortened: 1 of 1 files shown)


pupuis:
Question: Should I maybe not link the pull request, as there would still be the viewport enhancement and thus, the pull request does not quite close the issue yet or would we open a new issue?

Thanks in advance.

fonsp:
Strangely I am measuring no performance benefit... I tested on: 
https://github.com/fonsp/disorganised-mess/blob/master/big.jl


It does however fix a codemirror glitch when unfolding a cell, so I'm happy 🙃

fonsp:
Thank you! 🌈

Issue #492, opened 2020-09-25
Opening Github /raw links doesn't work


musabkilic:
Hi! I was working on this issue on the [18S191 repository](https://github.com/mitmath/18S191/issues/23) and I realized that the links in the form of github.com / *author* / *repo* / raw / ..... , doesn't open. 

I believe the problem can be solved by removing the "/raw" part from the URL path similar to what Pluto already does to /blob links. 
https://github.com/fonsp/Pluto.jl/blob/df1fc957370027f3b459535893798eee4f9637a2/frontend/components/Welcome.js#L55-L58

fonsp:
What is a `/raw` link?

musabkilic:
/blob https://github.com/fonsp/Pluto.jl/blob/master/sample/Basic.jl
/raw https://github.com/fonsp/Pluto.jl/raw/master/sample/Basic.jl

I'm referring the normal Github interface webpage as a /blob link, and the raw file as a /raw link. A /raw link redirects to https://raw.githubusercontent.com. Sorry if my wording/naming is confusing.

fonsp:
If this is already a raw link, then it already works, right? You can paste any link to a `.jl` file, there are just 2 special cases to turn github links into raw links.

musabkilic:
That's what I and I think presumably @zmoon thought, too. And I started changing the links in the 18S191 website to sent a pull request and before sending one I tried and I saw that Pluto doesn't work the way I thought it would.

Here is how Pluto handles https://github.com/fonsp/Pluto.jl/raw/master/sample/Basic.jl .

1- It detects that this is a github.com link
2- Replaces github.com part with raw.githubusercontent.com
3- There is no /blob part so we are done

Then Pluto tries to open https://raw.githubusercontent.com/fonsp/Pluto.jl/raw/master/sample/Basic.jl but Github returns 404 error because this is an invalid link. The correct link is at https://raw.githubusercontent.com/fonsp/Pluto.jl/master/sample/Basic.jl, notice the /raw part missing.

https://github.com/fonsp/Pluto.jl/blob/df1fc957370027f3b459535893798eee4f9637a2/frontend/components/Welcome.js#L55-L58

PR #493, opened 2020-09-25
Add Github /raw link support

musabkilic:
Explained here: #492

### DIFF
--- a/frontend/components/Welcome.js
+++ b/frontend/components/Welcome.js
@@ -55,6 +55,7 @@ export const process_path_or_url = async (path_or_url) => {
         } else if (u.host === "github.com") {
             u.host = "raw.githubusercontent.com"
             u.pathname = u.pathname.replace("/blob", "")
+            u.pathname = u.pathname.replace("/raw", "")
         }
         return {
             type: "url",
--- a/frontend/components/Welcome.js
+++ b/frontend/components/Welcome.js
@@ -53,9 +53,7 @@ export const process_path_or_url = async (path_or_url) => {
                 path_or_url: files[0].raw_url,
             }
         } else if (u.host === "github.com") {
-            u.host = "raw.githubusercontent.com"
-            u.pathname = u.pathname.replace("/blob", "")
-            u.pathname = u.pathname.replace("/raw", "")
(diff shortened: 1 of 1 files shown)


zmoon:
Seems like it would be safer to replace `"/raw/"`? Because if it is a `/blob` link and the filename starts with `raw` that would mess it up.

musabkilic:
Yes, I get your concern and I was just thinking about the same thing. Another thing to mention is that if the author or the repository name starts with "blob", current code replaces that part. This might sound unlikely but it can cause problems in the future.

I think the best/safest way to handle this is to *not* change github.com to raw.githubusercontent.com. Instead, adding "?raw=true" query string to URL returns the corresponding raw.githubusercontent.com page, i.e we can let Github redirect the raw file link.

Example:
https://github.com/fonsp/Pluto.jl/blob/master/sample/Basic.jl => https://github.com/fonsp/Pluto.jl/blob/master/sample/Basic.jl?raw=true

It's also easy to implement with just adding `u.searchParams.set("raw", true)` and removing the "replace /blob with this github with that" part. If @fonsp is happy with this approach, I can update the code.

zmoon:
That's cool! Then any of the GitHub link types can be used, no `pathname` modification necessary.

fonsp:
There's a reason why I didn't use that, but I forgot what it was 😬 Let me think...

musabkilic:
https://github.com/fonsp/Pluto.jl/commit/b26b0f1df4e131f9f2b285c622048f72b22b47f9#diff-d5648022db67f8ce30e167edf519aa1eR47

https://github.com/fonsp/Pluto.jl/commit/9fdc353e487ce097481b86edb5745a4ca4597fce#diff-d5648022db67f8ce30e167edf519aa1e

These might refresh some memories. Maybe setting the search/query part(not adding raw=true, just replacing all the query part) caused some problems.

fonsp:
Haha thanks, now I remember:

I changed it here: https://glitch.com/edit/#!/pluto-on-binder

because I had an escaping issue with the `?`, but I later found that the issue is solved by URL encoding twice :)

Let me try...

fonsp:
Resolves #492

PR #498, opened 2020-09-27
📜 Logging tweaks and fixes

c42f:
* Prevent frontend from crashing with unknown log levels
  (eg ProgressLogging.@-progress)
* Avoid spamming the user with infinite log relay errors once the remote
  log channel is closed
* Only relay info-and-above-level logs from third party modules, to
  prevent excessive debug level log spam.

Fixes a few problems I found while testing #496

### DIFF
--- a/frontend/common/Logging.js
+++ b/frontend/common/Logging.js
@@ -6,7 +6,7 @@ const log_functions = {
 }
 
 export const handle_log = ({ level, msg, file, line, kwargs }, filename) => {
-    const f = log_functions[level]
+    const f = log_functions[level] || console.log
     const args = [`%c┌ ${level}:\n`, `font-weight: bold`, msg]
     if (Object.keys(kwargs).length !== 0) {
         args.push(kwargs)
--- a/frontend/common/Logging.js
+++ b/frontend/common/Logging.js
@@ -11,21 +11,11 @@ export const handle_log = ({ level, msg, file, line, kwargs }, filename) => {
     if (Object.keys(kwargs).length !== 0) {
         args.push(kwargs)
     }
-
     if (file.startsWith(filename)) {
         const cell_id = file.substring(file.indexOf("#==#") + 4)
         const cell_node = document.getElementById(cell_id)
 
-        // hacky hackkkkkk
-        const cm_internal_node = cell_node.querySelector(".CodeMirror-code")
-        const origin_line = cm_internal_node.children[line - 1]
-
-        args.push(`\n\nfrom`)
-        if (origin_line == null) {
-            args.push(cell_node)
-        } else {
-            args.push(origin_line)
-        }
+        args.push(`\n\nfrom`, cell_node)
     }
     f(...args)
 }
--- a/src/evaluation/WorkspaceManager.jl
+++ b/src/evaluation/WorkspaceManager.jl
@@ -69,6 +69,9 @@ function start_relaying_logs((session, notebook)::Tuple{ServerSession, Notebook}
             next_log = take!(log_channel)
             putnotebookupdates!(session, notebook, UpdateMessage(:log, next_log, notebook))
         catch e
+            if !isopen(log_channel)
+                break
+            end
             @error "Failed to relay log" exception=(e, catch_backtrace())
         end
     end
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -704,7 +704,12 @@ const old_logger = Ref{Any}(nothing)
 
 struct PlutoLogger <: Logging.AbstractLogger end
 
-Logging.shouldlog(::PlutoLogger, args...) = true
+function Logging.shouldlog(::PlutoLogger, level, _module, _...)
+    # Accept logs
+    # - From the user's workspace module
+    # - Info level and above for other modules
+    _module === current_module || convert(Logging.LogLevel, level) >= Logging.Info
+end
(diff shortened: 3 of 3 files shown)


fonsp:
Thanks! I rushed this feature a little 😬

About relaying Debug level messages: the browser console actually has a log level filter built in!

Everything enabled:

image

Debug disabled (button on top of the console):

image

We'll have to think about how all of this should be documented, but maybe we can leave in the Debug logs and ask users to filter there? On the other hand, the use case here is to enable logging for setups like github codespaces, binder, SSH, etc, and maybe Debug is only useful for package development, which you wouldn't do on those setups?

fonsp:
Oh I just saw that you relay Debug from inside the notebook, that's smart

src/evaluation/WorkspaceManager.jl
```diff
@@ -69,6 +69,9 @@ function start_relaying_logs((session, notebook)::Tuple{ServerSession, Notebook}
             next_log = take!(log_channel)
             putnotebookupdates!(session, notebook, UpdateMessage(:log, next_log, notebook))
         catch e
+            if !isopen(log_channel)
+                break
+            end
```
fonsp, line 74: `take!(log_channel)` is a blocking call, and we only call the `start_relaying_logs` functions once per workspace (async), so we should never `break` the loop, right?

fonsp (changes requested):

src/evaluation/WorkspaceManager.jl
c42f: I noticed that when the user presses CTRL-C in the REPL to kill Pluto, the REPL is immediately flooded with an infinite stream of log messages:

```
┌ Error: Failed to relay log
│   exception =
│    Distributed.ProcessExitedException(2)
│    Stacktrace:
│     [1] worker_from_id(::Distributed.ProcessGroup, ::Int64) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/cluster.jl:1074
│     [2] worker_from_id at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/cluster.jl:1071 [inlined]
│     [3] #remotecall_fetch#146 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:421 [inlined]
│     [4] remotecall_fetch at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:421 [inlined]
│     [5] call_on_owner at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:494 [inlined]
│     [6] take! at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:626 [inlined]
│     [7] start_relaying_logs(::Tuple{Pluto.ServerSession,Pluto.Notebook}, ::Distributed.RemoteChannel{Channel{Any}}) at /home/chris/.julia/dev/Pluto/src/evaluation/WorkspaceManager.jl:69
│     [8] (::Pluto.WorkspaceManager.var"#1#2"{Pluto.ServerSession,Pluto.Notebook,Distributed.RemoteChannel{Channel{Any}}})() at ./task.jl:356
└                            @ Pluto.WorkspaceManager /home/chris/.julia/dev/Pluto/src/evaluation/WorkspaceManager.jl:72
```

This seemed to fix that case — I assumed because the remote process had been killed, and the `RemoteChannel` was now closed.  In this case, the async `Task` which is running `start_relaying_logs` is now an orphan, and we should allow it to exit (otherwise it'll run forever and never be cleaned up).

frontend/common/Logging.js
```diff
@@ -6,26 +6,16 @@ const log_functions = {
 }
 
 export const handle_log = ({ level, msg, file, line, kwargs }, filename) => {
-    const f = log_functions[level]
+    const f = log_functions[level] || console.log
```
c42f, line 9: BTW, we could try to infer more closely whether it should be one of `debug`, `info`, `warn` and `error` should be used for nonstandard log levels (on the Julia side).  But I figured that that console log isn't the UI we want anyway, so just having a crude fallback to bare `console.log` would be simple for now.

c42f:
> Oh I just saw that you relay Debug from inside the notebook, that's smart

I'm not quite sure what you mean by relay?

All I've done is try to infer what the user cares about: they care about debugging their own notebook code, but not debugging some library which they happen to be `using`. So I suppressed `@debug` messages from third party Julia modules. This is partly for user convenience, and partly for performance.  Performance-wise I don't think we want the backend sending 10_000s of likely-irrelevant debug log messages, even if they're hidden from the user.

src/evaluation/WorkspaceManager.jl
c42f: As a side note - whenever you start code with `@async` it's worth thinking about how the `Task` will exit. Julia doesn't have a task cancellation API, so you need to do it manually by checking whether the resources needed by the task are still open. Typically these "resources" would be a channel of some kind.

See also https://github.com/JuliaLang/julia/issues/33248 for a lot of discussion about task lifetime.

fonsp (approved):

fonsp:
Thanks! While I have you here, would you like to call today?

c42f:
Ah I missed that, sorry :)

PR #502, opened 2020-09-27
Fixes condition for is_hidden for CellInput

### LINKED ISSUE #500: [dev] Cell.js - is_hidden is always false
Hi,

while working on Cell.js I realised that it seems that is_hidden for CellInput always evaluates to false as it contains the condition ```class_code_folded && !class_code_folded```. I suppose that this is a typo? However, I wouldn't know how to fix it, as I cannot recognise what was originally intended. I think this was introduced in a recent commit.

Thank you in advance for looking into it.

Best

pupuis:
Hi @fonsp ,

I checked out your commit and fixed the condition based on what I think was your intention. Was I right?

(Closes #500)


fonsp:
❤

PR #505, opened 2020-09-28
Forces non-empty notebooks even when deleting multiple c…

### LINKED ISSUE #444: Notebook without cells
Pluto generally prevents you from creating notebooks with no cells, by making a new cell when you delete the last one. This makes sense, since there is no reason to have a notebook with no cells.

However, I noticed that to create an empty notebook by holding backspace in the only cell. After the cell disappears, my browser interprets the backspace as a shortcut to go to the previous page (the pluto menu). So I leave the notebook before the new cell is created. This is something that a user (like me) can do on accident.

Opening an empty notebook while it is still running gives an unusable notebook, since there is no button to create new cells. Opening after shutting it down results in a MethodError. Either way, not great.

It might be good to make sure that empty notebooks are loaded properly (by adding an empty cell to the notebook on loading). However, a more robust way of preventing empty notebooks would also solve the problem.

pupuis:
…ells at once.

I think this should solve the bug described by @ekzhang. I'm not sure if this actually solves the bug described by the author of the issue, I was unable to reproduce it. I actually came to the same conclusion as @ekzhang: Deleting one works, deleting multiple does not. Maybe that is what the author intended as well.

On implementation: I moved inserting a new cell if the last remaining one is deleted out of ```delete_cell``` to generalise the behaviour to multiple cells as well (i.e. new function ```delete_multiple_cells``` works in case of one or multiple deleted cells. I based this on the assumption that ```delete_cell``` is only called by ```confirm_delete_multiple```. If that is not the case I could make the check optional via an optional parameter. If anyone knows another function that calls ```delete_cell```please let me know.

Edit: (As mentioned in issue #444).


fonsp:
I don't know if it's called elsewhere but you can search the entire project using VS Code (magnifying glass on the left). If not, can you delete the action entirel?

pupuis:
As far as I see, the first three are a css-class for buttons, so not related to the Editor function.
The stuff in Editor.js are my changes and the last two seem to be the backend functions that are called when we do a ```this.client.send``` in the Editor to delete the cell in the backend.

Screenshot 2020-09-29 at 21 15 10

So seems to me that it's unused otherwise. However, I would not want to delete function as it encapsulates the behaviour for single cells and pretty much describes perfectly what it does: Initiating that a single cell can be deleted. So in the light of coding style I would say we should keep the function, but I can also paste its code into the ```forEach``` in the new function ```delete_multiple_cells```, if you'd prefer it that way.

ekzhang:
How about writing the logic inline? Not speaking for Fons here, but I think one potential downside of keeping the `delete_cell` function in the `actions` object is that we expose it to all of the children, but we might not want strangers to call the function directly.

pupuis:
Turns out I was painfully unaware that those weren't class-methods but part of the requests 'interface'. Having understood this, I suggest we turn ```delete_cell``` into a class-method but not part of the requests. Then the encapsulation works, but we don't 'expose' it. Thanks for clarifying that!

pupuis:
Okay, I fixed that! More review welcome 👍

fonsp:
Just to confirm, did you check that this solves #444, i.e. holding backspace?

pupuis:
After checking again I realised that my fix only solves the problem described by @ekzhang. I remember not being able to reproduce the issue described by the author. By browser did not react in the same way.

Maybe we can split the issue, s.t. the problem described by @ekzhang can be solved by this PR and the original issue might first require someone to reproduce it.

But yeah, this PR does not solve the original issue. Sorry for that. I can also simply remove the link to the issue for this to be able to merge. What do you suggest?

fonsp:
Which problem is that?

pupuis:
Selecting all remaining cells and deleting them yields an empty notebook

pupuis:
commented under the issue #444

pupuis:
Quick update: I removed the linkt to #444, as it does not close the issue but solves a different bug that was only mentioned there. It is probably no need to open a separate issue just to have this PR link to something.

So I am still happy about reviews or feedback but also just to merge this one. Let me just quickly resolve conflicts.

fonsp:
@dralletje was the issue fixed?

dralletje:
> @dralletje was the issue fixed?

Yes

PR #521, opened 2020-10-01
Handle parametric type assignments, closes #517

### LINKED ISSUE #517: Union with Parameter Type throws UndefVarError
I got UndefVarError with below

```julia
Option{T} = Union{T, Nothing}
```

output
```
UndefVarError: T not defined

top-level scope@Local: 1
```

in Terminal

```
┌ Warning: unknow use of `=`. Assignee is unrecognised.
│   ex = :(Option{T})
└ @ Pluto.ExpressionExplorer ~/.julia/packages/Pluto/upwxH/src/analysis/ExpressionExplorer.jl:160
```

Same code can function on julia repl.

## Environment:
- julia 1.5.0
- Pluto v0.11.14

Pangoraw:
Hi! This PR fixes the assignments of parametric types as disclosed in #517. For these, I did two things:
 1. The warning was fixed by handling these types of assignments and by constructing the references (https://github.com/Pangoraw/Pluto.jl/commit/d29df8d3f2271e5745cdbe0cba35296c4da05aba)
 2. The error was due to the fact that `julia` is trying to use the generic type as a normal type when we use an expression like this: `local result = $(expr)` as in Pluto's runner. The solution was to wrap the assignment in a `const` (a `begin` would also work) expression (https://github.com/Pangoraw/Pluto.jl/commit/fab4715b0abfc8ff99986c28f775d51fff3b4a5d).
###### Without `const`
```julia
julia> local result = (Nullable{T} = Union{T, Nothing})
ERROR: UndefVarError: T not defined
Stacktrace:
 [1] top-level scope at REPL[1]:1
 [2] run_repl(::REPL.AbstractREPL, ::Any) at /build/julia/src/julia-1.5.1/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:288
```
###### With `const`
```julia
julia> local result = (const Nullable{T} = Union{T, Nothing})
Union{Nothing, T} where T
```

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -282,12 +282,19 @@ function explore!(ex::Expr, scopestate::ScopeState)::SymbolsState
             # function f(x, y) x + y end
             return explore!(Expr(:function, ex.args...), scopestate)
         end
-        assignees = get_assignees(ex.args[1])
+
         val = ex.args[2]
+        # Handle generic types assignments A{B} = C{B, Int}
+        if ex.args[1] isa Expr && ex.args[1].head == :curly
+            assignees, symstate = explore_funcdef!(ex.args[1], scopestate)
+            innersymstate = union!(symstate, explore!(val, scopestate))
+        else
+            assignees = get_assignees(ex.args[1])
+            symstate = innersymstate = explore!(val, scopestate)
+        end
 
         global_assignees = get_global_assignees(assignees, scopestate)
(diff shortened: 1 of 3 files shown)


fonsp:
Another one!

fonsp:
Maybe you could have a look at a big one: #177

Pangoraw:
> Maybe you could have a look at a big one: # 177

I will take a look, it seems like an interesting challenge :eyes:

Issue #302, opened 2020-08-17
Better autocomplete


fonsp:
Sorting autocomplete results:

_Right now results are presented in alphabetical (i.e. code index) order. The dream is to have full AI power like kite.com, but perhaps there are some simple improvements to be made first._

- [x] lowercase first, then uppercase?
- [x] for `Module.<TAB>`, first show names that are `export`ed by the module, then the rest. The set of exported names should be everything that you need, and the other names are probably not meant to be used.
- [x] Show basic type info: is it a function, submodule, ...?
- [x] Fuzzy search
- [x] Autocomplete all words in the same CM - https://codemirror.net/demo/anywordhint.html

### Please post other ideas and suggestions!

fonsp:
Those already work inside Pluto - is it broken for you?

fonsp:
Great to hear that you want to contribute! If you want to work with Julia instead of JS, then writing sample notebooks might be the most fun. Ideas [here](https://github.com/lukavdplas/pluto-notebooks/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) and [here](https://github.com/fonsp/Pluto.jl/projects/2#column-9124852), I'd be happy to have a video call if you have an idea: fonsvdplas@gmail.com

cadojo:
> Those already work inside Pluto - is it broken for you? You might need to press TAB twice.

Ah I must have missed it before, it is working for me! Thanks!

> writing sample notebooks might be the most fun

I'm about to start my last year of grad school, I'll be taking a few aerospace dynamics courses this semester that might offer opportunities for interesting functional programming examples. I'll see what my classes look like - if an assignment fits, two birds with one stone. If not, I'll add an example or two this semester. I'll send an email with ideas once they're more thought out!

fonsp:
Nice! i think a functional programming sample could be really useful - Pluto makes more sense when you program in the functional style

fonsp:
https://github.com/fonsp/Pluto.jl/issues/111#issuecomment-683432200

pbouffard:
I'm not sure what it's called (fuzzy searching maybe?) but the kind of autocomplete where you are shown results that include not just those that begin with the characters you've typed so far.

I.e. like:
<img width="447" alt="image" src="https://user-images.githubusercontent.com/231346/96934671-7a45f180-1477-11eb-8290-620755b613c7.png">

I use this constantly in various IDE's, including VS Code.

fonsp:
[...]

PR #526, opened 2020-10-02
Sort autocomplete by showing lowercase first

iboss-ptk:
Resolve part of #302:
- lowercase first, then uppercase


fonsp:
If it sorts in reverse, does it now give `z` before `a`?

iboss-ptk:
@fonsp It is. Should we go like
- lowercase
- @stuffs (for macro)
- uppercase
- the rest
sorted alphabetically for each bucket instead?

iboss-ptk:
@fonsp It is. Should we go like

lowercase
uppercase
the rest
sorted alphabetically for each bucket instead?

fonsp:
Yep! Just a style comment: you can use a [stable sort algorithm](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability) like MergeSort, and sort by a function that just assigns categories:
```julia
heuristic(s::String) = if islowercase(firstchar(s))
	1
elseif isuppercase(firstchar(s)
	2
else
	3
end
```

iboss-ptk:
@fonsp cool! I will look into that :D

PR #530, opened 2020-10-03
Find and replace feature

### LINKED ISSUE #96: Find (and Replace)
It would be super useful to have a 'Find' and 'Find and Replace' utility in Pluto! :)

Related to this, I think it would be useful if one could toggle between showing/collapsing all code cells. (I'm imaging the workflow: show all cells, find something, fix it and collapse the cells back to their original state.) Cheers!

pupuis:
...closes #96 

I am working on a feature for find and replace in Pluto. Currently only draft. 

Features:

* Collapsible find/replace pane with Ctrl+F
* Highlighting matching words while typing
* Skipping through individual matches (i.e. focusing matches)
* Replace currently focused match

I want to add:

* Auto-paste of selected word when pressing Ctrl+F
* Replace all
* Convenient behaviours (e.g. first replace is also focusing).

If someone has ideas/wishes, I'd be happy to hear.

# Try it out!

```julia
julia> ]
pkg> activate --temp
pkg> add https://github.com/pupuis/Pluto.jl#find-and-replace
```


pupuis:
Hi all, I finished a first working prototype now and I would very much appreciate feedback in any form. I found space for the find-replace pane in the header, which I think fits in quite nicely.

## Features:

* Ctrl+F opens the pane. If one selects a word in a cell first, the pane opens with the selected word already entered
* All matching words in all cells are highlighted (red colour)
* Pressing "Next" selects the next match. Selection: red colour + grey background
* Replace button replaces the currently selected match with the text entered or behaves like "Next" if nothing is selected yet (other editors replace the first match instantly but I don't know if I like that)
* "All" button replaces all matches (regardless of selections)
* Hidden cells are ignored
* Highlights while typing the matching word

## Demo:

findreplace

(Sorry for the low-quality GIF by the way, I had to get it under 10 MB :D But it should give an overall feeling for the behaviour)

### Here's a demo of the on-the-fly highlighting (i.e. as you type):

findreplace2

fonsp:
This is really great, thanks so much! Some quick comments just looking at what you wrote:
- placing the UI inside the header is well found, but i think it would be better in its own container, because:
  - this is not a "meta" action, all other things in the header are
  - i am worried that this will not work well on small screens
- we should also search inside hidden cells, especially when replacing. This is made a bit more difficult by recent "optimizations" that delete the entire codemirror when the cell is hidden. I dont like these optz so we might undo them again.
- could you try making the buttons gray/blue/🌈 instead of brown?

fonsp:
Also, have you tried to use CodeMirror's built-in search-replace functionality?

pupuis:
> 
> 
> Also, have you tried to use CodeMirror's [built-in search-replace functionality](https://codemirror.net/doc/manual.html#addon_searchcursor)?

Yes, I use the addon when searching each instance respectively. You can find the implementation using the codemirror method ```getSearchCursor``` in ```common/FindReplace.js``` in the function ```find_all_markers```, which implements the search for matches in an individual instance.

pupuis:
> 
>     placing the UI inside the header is well found, but i think it would be better in its own container, because:
>
I moved the pane to the bottom-left and made it sticky (similar to the live docs on the right).

> 
>     we should also search inside hidden cells, especially when replacing. This is made a bit more difficult by recent "optimizations" that delete the entire codemirror when the cell is hidden. I dont like these optz so we might undo them again. #628
>
This requires a bit more thinking I suppose. I imagine something like: If we navigate to a match found in a hidden cell, we temporarily make it visible but fold it again once the navigation leaves that code mirror. For now, I have only removed all checks for the cell being hidden when searching. The temporarily-unfolding-thing has yet to be implemented.

>     * could you try making the buttons gray/blue/🌈 instead of brown?

I made them purple for now, but I don't know if it looks alright. Maybe someone with more talent for color schemes than me can find a better color? :D

pupuis:
Still working on the following problems:

* Page should focus the selected match when navigating (should be easy (cm.focus())
* Some frontend-tests do not pass (not really sure why yet)
* Temporarily unfolding cells when navigating

pupuis:
@fonsp @dralletje: Quick question to either of you: I was experimenting with the JS method ```Element.scrollIntoView``` to make sure that upon navigating through the matches, if a match is not in the viewport, the page is scrolled to put the corresponding code mirror into the viewport's center. This works pretty neat on my Firefox and Chrome installations, however, [this source](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) states that the feature is experimental still.

Now to my question: Do we want to use features that are experiemental (such as ```scrollIntoView```) in Pluto? How's your opinion on that?

fonsp:
Clicking on a stack trace in Pluto already has this feature: you peek into hidden cells, and the line is scrolled into view. Have a look at how this works

pupuis:
Ok, well, as far as I see, in the stack trace a custom event ```cell_focus``` is dispatched, which makes the corresponding codemirror focus. This has two drawbacks with regards to my scenario:

1. Focusing means also that the cursor is placed in the code mirror instance (which is not necessarily what I want when navigating through the matches (see below)).

2. Focusing scrolls the instance to the bottom of the window (if it is not visible) - maybe I'm wrong at this point, but I kind of remember it that way

**Why focusing is an issue**
I see two ways to go here: One would be that the user navigates through the matches using another shortcut (e.g. ctrl + arrows or something) and/or with the "next" button. Then we could focus the code mirror and select the matched word so that the user can change it right away - might be useful. The other way would be (and this is how I implemented it here) that the user navigates by pressing enter while having focus on the find-and-replace-pane. This is e.g. the behaviour of Atom.

Anyways, if we want the latter way, it still raises the question if we would want the experimental feature to stay in the code. If you have concerns, I will look for an alternative (there surely is one 👍)

fonsp:
The experimental feature is not a problem, but definitely try to reuse the same mechanism as stack traces. Feel free to expand its functionality if you need to

fonsp:
(Join us at https://julialang.zulipchat.com/#narrow/stream/257776-Pluto.2Ejl-development !)

pupuis:
>
>but definitely try to reuse the same mechanism as stack traces
>
Ok, I checked that mechanism again, but here is what essentially differs to my use-case: The stack-trace click makes the cell focus (event ```cell_focus```), which means that the codemirror is scrolled to the bottom of the view and the code causing the error (I assume) is selected. But this sets the cursor into the codemirror, whereas I simply want the user to see the current match (of the find-and-replace process) and possibly in the center of the screen - where I hope the user's attention most likely should be. 

Bottom-line is, I don't really know what to reuse from the stack-trace, I think the desired effect of the two use-cases is just different. Am I missing something?

dralletje:
You don't have to use it 1:1, but you can use whatever scrolling it uses.

Also, we do want to select the matching text, right? VSCode and any editor I've ever used does that 😁

pupuis:
By selecting I mean: Set the cursor on or around a match.

My VSCode does not do that. If I type after pressing enter, it types in the find field. So the cursor was set there. Of course we are marking the match.

Does it behave differently for you?

dralletje:
You are completely right, I confused "being highlighted" with "being selected" 🤦🏽‍♀️

pupuis:
Good to have clarified the terminology though ✌🏼☺️

>
> You don't have to use it 1:1, but you can use whatever scrolling it uses

What I wanted to get at before is that, as far as I understand, it uses the built-in scrolling used by CodeMirror, when calling focus().

pankgeorg:
I've updated this PR with the latest `master` and I've refactor the code a bit to be more structured (moved most stuff into `FindReplace.js` and removed the stuff that worked with the previous state type). I'll add more stuff soon to make it even more robust.

todo list:

- [ ] highlight newly typed words
- [ ] escape close
- [ ] search in hidden cells (kind-of-happens - will remove the codemirror unmounting)
- [ ] test?

fonsp:
Right now we delete a codemirror when we hide a cell instead of just `display: none` -- this was a mistake, there's no performance benefit (i tested), and it means that the Ctrl+Z history of a codemirror is lost when you hide it. So feel free to undo! https://github.com/fonsp/Pluto.jl/pull/484

pupuis:
@fonsp Seems to me that in the current master branch this was already undone. The if clause is gone. It seems that the codemirror instance is created once on the first render of the CellInput and then this effect

```js
useEffect(() => {
      cm_ref.current.options.disableInput = disable_input
}, [disable_input])
```
controls ```disableInput```.

fonsp:
Ah I see, it got rewritten to an if statement in Cell.js: https://github.com/fonsp/Pluto.jl/blob/37ca5533febe7635150d2a80ab65184af89788e8/frontend/components/Cell.js#L113-L114

pankgeorg:
@fonsp I changed these lines (113-114) to always have the CodeMirror there, with display: none in css.

@pupuis the `disableInput` still shows the CodeMirror, it just doesn't allow text input.

I'm also sending a custom event so the `CellInput` component updates the Cell's class to unhide it (classes should exist in React state). Still have a bug where the textmarkers in newly shown cells don't show up.

pankgeorg:
Hey @pupuis !! I've updated the code with the latest and restructured it a bit - can you take a look if you have some time and give me feedback on what you think?
Thanks!

pupuis:
Hi @pankgeorg, thank you so much, I'm happy you could extend/refactor my initial code. I'm trying to find time this week to check it out! Quick question: Did you implement, or do you think it is possible to add, some unit tests? I think it would help to improve the robustness. 

Tell me what you think.

Best

pankgeorg:
I didn't implement any test yet but I will definitely do !! Ping me for any questions!

pupuis:
Hi, I will probably not find time to look into it until the second week of March, but am afterwards relatively free, so maybe we can then finalize the draft. Here are two things I'd love to get done:
- [ ] Add unit tests
- [ ] Add regex search functionality (if not already implemented). I think this should be relatively easy, but its a very nice feature, in my opinion.

pupuis:
If you find time, we can also coordinate and work on it together!

pankgeorg:
>Add regex search functionality (if not already implemented). I think this should be relatively easy, but its a very nice feature, in my opinion.

great idea, not implemented yet! All yours!!

> If you find time, we can also coordinate and work on it together!

yeap, we can coordinate on [Zulip](https://julialang.zulipchat.com/#narrow/stream/257776-Pluto.2Ejl-development)!!

pupuis:
Hi @pankgeorg, I finally got to work on the PR, March was too busy. However, upon activating Pluto in the find-and-replace branch with

```
(@v1.4) pkg> activate .
(pluto-tests) pkg> dev ../Pluto.jl/
```
I encountered the following error:
```
ERROR: Unsatisfiable requirements detected for package TableIOInterface [d1efa939]:
 TableIOInterface [d1efa939] log:
 ├─TableIOInterface [d1efa939] has no known versions!
 └─restricted to versions 0.1 by Pluto [c3e4b0f8] — no versions left
   └─Pluto [c3e4b0f8] log:
     ├─possible versions are: 0.12.20 or uninstalled
     └─Pluto [c3e4b0f8] is fixed to version 0.12.20
``` 
I tried to find ```TableIOInterface.js``` and saw that you are also a contributor. Can you reproduce this error or know how I can fix it? Maybe I am missing something... Thank you in advance.

fonsp:
Try: `Pkg.update(); Pkg.resolve(); Pkg.instantiate();`, this problem is unrelated to this PR.

pupuis:
Sorry, you're right, nothing to do with the PR.

Your hint helped, thanks.

pupuis:
Hi,

I would like to get this feature to a state in which it can get merged. First of all, thank you @pankgeorg for refactoring and moving things into the joint FindReplace.js and for keeping this PR maintained! After testing the functionality, I have the following four additions/fixes in mind:

- [ ] replacing does (always) not reset selections [BUG]
- [ ] pressing next after changing the search-string should trigger re-selection, starting with the first match
- [ ] quitting should unselect everything
- [ ] regex functionality

This in summary is what I will turn to now.

pupuis:
## Quick overview over the regex feature
A regex can be indicated via ```/regExp/```. If valid:
grafik

If invalid:
grafik

pupuis:
Hi @fonsp, @dralletje: I forgot to write it down explicitly last time but from my side, this feature is (at least in its first version) done. I will mark it as ready for review, so you can now test it and give comments. Any feedback is highly appreciated.

pankgeorg:
That's great! I'll have a look soon!!

pankgeorg:
This is amazing work @pupuis!

This is soo close - it needs a couple of things:
- [ ] update highlighting as you write into a codemirror (working on it)
- [ ] 'softer'/incremental updates of the textmarker array -- it feels a _little_ slow on big notebooks.
- [ ] our menu doesn't popup in non-english search combination (the native does popup) - `control + φ` in greek for example

pupuis:
Hi @pankgeorg,

as for the incremental updates of the textmarker array: I suppose you mean searching the entire notebook and creating a textmarker instance for each match is too slow?

I mean optimally we only instantiate once a match comes into view, but we should still search over the entire notebook, so we can tell the user how many matches we found. 
If you had something in mind for this already, please let me know.

As for the popup on non-english keyboards: We can add greek, but for other keyboards we might want to wait for issues, right? Or should be support a couple of keyboards by default. How is this handled with other keyboard shortcuts in Pluto?

fonsp:
Hi @pupuis !

Thanks again for your work! Sorry for the long process!

@pankgeorg wrote the last comment more as a todo list for ourselves, p and I discussed these last todo items and we already have some ideas for solutions. Sorry for the confusion!

CarloLucibello:
Any further progress on this? It is a very basic and useful feature for a coding environment

fonsp:
Closing this because it needs to be redone after #1159

PR #532, opened 2020-10-04
Bugfix: Delete button not working on feedback field etc. (inputs)

pupuis:
Hi,

I've noticed that due to a recent commit the backspace/delete key would not delete text in html-inputs (e.g. instant feedback). I removed the ```e.preventDefault()```that caused this.

Commit that caused it: https://github.com/fonsp/Pluto.jl/commit/1180226ff997001105200cad151f4cdb18795cdf by @ekzhang @fonsp.

My solution is possibly not the best way to do this, but at least it would fix the bug for now.

Do we need the ```preventDefault```altogether?

Greetings ☺️

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -677,6 +677,7 @@ export class Editor extends Component {
             const selected = this.state.notebook.cells.filter((c) => c.selected)
             if (selected.length > 0) {
                 this.requests.confirm_delete_multiple(verb, selected)
+                return true
             }
         }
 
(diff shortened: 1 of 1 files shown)


fonsp:
I don't think that we need `preventDefault`. Did you try if deleting multiple cells still works?

pupuis:
I can check again and inform you. However, aren't some browsers going to the previous page on a backspace/delete? If yes, then we need the preventDefault to make sure users dont accidentaly skip back upon backspace/delete.

fonsp:
Ah that's right! This is why we need it.

PR #538, opened 2020-10-07
👒 Include function signature in reactive edge - fix #177

### LINKED ISSUE #177: Support defining methods of a function in multiple cells

fonsp:
If you want to try it out:
```julia
pkg> activate --temp
pkg> add Pluto#canonical-function-root-as-reactive-link
```

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -22,13 +22,11 @@ jobs:
             matrix:
                 # We test quite a lot of versions because we do some OS and version specific things unfortunately
                 julia-version: ["1.0", "1.1", "1.2", "1.3", "1.4", "1.5"]
-                os: [ubuntu-latest, windows-latest]
-                include:
-                    # macOS - only the extremes, linux and windows cover pretty much everything already i think
-                    - julia-version: "1.0"
-                      os: macOS-latest
-                    - julia-version: "1.5"
-                      os: macOS-latest
+                os: [ubuntu-latest, windows-latest, macOS-latest]
+                exclude:
+                    # macOS and windows - only the extremes, linux covers pretty much everything already i think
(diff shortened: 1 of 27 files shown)


fonsp:
Important bits are:

## New type to represent nodes in the reactive graph, with more detailed front and back edges
https://github.com/fonsp/Pluto.jl/blob/496d5b5c2f5dcdee348a6a9ad1972e1585151325/src/analysis/ReactiveNode.jl#L3-L59

https://github.com/fonsp/Pluto.jl/blob/496d5b5c2f5dcdee348a6a9ad1972e1585151325/src/analysis/Topology.jl#L63-L88

## We canonalize+hash the expression containing the method signature
https://github.com/fonsp/Pluto.jl/blob/496d5b5c2f5dcdee348a6a9ad1972e1585151325/src/analysis/ExpressionExplorer.jl#L701-L763

https://github.com/fonsp/Pluto.jl/blob/496d5b5c2f5dcdee348a6a9ad1972e1585151325/test/MethodSignatures.jl

fonsp:
Haha the rewrite is 100x faster WHAT on the **first run**, because i found some code that causes lots of allocations.


On subsequent runs, there is a 1.5x - 2x runtime speedup, nice! No difference in allocs.

See results here:
https://github.com/fonsp/disorganised-mess/blob/master/pluto%20analysis%20benchmark.jl

(run in VS Code with Alt+Enter)

PR #543, opened 2020-10-07
add duplicate methods detection using signatures instead of names

fonsp:


fonsp:
Hi @Pangoraw!

In the end I decided to try it in a different direction, by using the function signature as the reactive edge, and without evaluating a dummy function. It works really well, and I think it is good enough to keep people happy for a while, but we might want to revisit a more complicated method later.

Anyway, just wanted to say thanks for experimenting and contributing to the issue! I look forward to hear what you think about the current approach.

-fonsi

Pangoraw:
Nice! Adding the feature while maintaining simplicity and performance is really impressive ! :rocket:

PR #547, opened 2020-10-08
Live doc search

### LINKED ISSUE #111: Better live docs
Some improvements:

- [x] Show what is currently being searched for
- [x] Support cross-reference links between docs
- [ ] Show that it is searching
- [ ] Title of the docs should be a textbox where you can enter a query directly
- [x] Handle dots, like `Module.funky`
- [ ] Function arguments
- [ ] Show when the name is found but it has no docs
- [ ] To the right of a cell might make more sense
- [ ] Heuristics for which function is dispatched on - can VS Code do this nowadays? I guess that we can do this dynamically, since some of the types are known

Extras:

- [ ] `apropos(str)` might be useful. Is resource intensive though, maybe write our own version that just returns the first hit? Although that might not be the most relevant one
- [ ] Show filename and line number of defintion. Double cool if it is a clickable link to GitHub! (Like in methodtables shown as `text/html`) I found this useful when the docs are unclear or when you want to know how something works. 

What is Julia's doc system missing?
- [ ] Given an object, **tell you what you can do with it**. In OOP languages this is `obj.<TAB>` to autocomplete.

### Please post other ideas and suggestions!

iboss-ptk:
resolve #111

- Make live doc title a searchable text box
- refactor live search doc into functional component


fonsp:
This is awesome!!!!!!

fonsp:
Thank you! ❤

PR #551, opened 2020-10-09
add css style for code inside markdown

Moelf:


Moelf:
after:
image

fonsp:
Can you make it prettier? Experiment with padding, border, rounded corners, an accent color

Moelf:
before:
image

after:
image

frontend/editor.css
```diff
@@ -167,6 +167,12 @@ pluto-output code {
     font-size: 0.75rem;
 }
 
+pluto-output .markdown code {
```
fonsp, line 170: What happens if you remove the `.markdown` selector?

Moelf: it will influence the normal output, say cell:
```
x = 3
```
the `3` will be displayed as if it's in markdown.

fonsp:
I will reformat this so that we don't need the `.markdown` selector:
```julia
html"""
<p>Hello <code>world</code></p>
"""
```
should look identical to 
```julia
md"""
Hello `world`
"""
```

fonsp:
I think it's `:not(.inline)` but i'm not sure

Moelf:
close, turns out it's `.inline_output` ;)
image

pankgeorg:
Why is the `border-radius` so big?
Github seems to use `6px`  (both `<code>` and `<pre>`)
Slack seems to use `3px` for `<code>` and `4px` for `<pre>`
[Documenter.jl](https://juliadocs.github.io/Documenter.jl/stable/) seems to use `0` with a `0.1em`padding and a bigger (`.875em vs .75em`) font size.

fonsp:
Because it's pretty 🌈

There's no way to discuss these things objectively, but note that we are not trying to match the style of either documenter, slack or github. (Especially not slack, yuck!)

pankgeorg:
That is not what I'm suggesting (to match something) -- but I'm pretty sure they've had teams of UX designers that decided against big border-radiuses (same is true for Jupyter btw - `2px`).
Even though it _is_ pretty, being pretty is not it's role - I think big/pretty --> distracting, that's why I would suggest a `6px` or less for both cases. Up to you, of course.

Moelf:
while I more or less agree with you, can we just get this merged first so I can at least SEE the code instead of NOT SEE them?

fonsp:
As Oscar Wilde put it:

> A fashion is merely a form of ugliness so absolutely unbearable that we have to alter it every six months!

PR #566, opened 2020-10-12
update msg

Moelf:
since this msg doesn't show up after `using` (by default just at precompile?)


fonsp:
Hmmmmmm let's try to make it print at `import` instead

Moelf:
I thought the current behavior is intentional? https://github.com/fonsp/Pluto.jl/pull/389

Issue #584, opened 2020-10-18
Pluto frontend doesn't display ligature fonts despite using JuliaMono


AnselmJeong:
To my understanding, current Pluto uses `JuliaMono` fonts.
However, the frontend of Pluto does not display ligature fonts properly. May I ask whether there any settings or ways that may enable ligature fonts?

I have been trying to modify several css settings like `font-variant-ligatures` but to no avail.

Always thanks to your contributions. I'm happily using this library.

fonsp:
Thank you for the suggestion! I like ligatures, especially the `|>` is pretty, but they make Julia slightly more difficult for beginners, because you can look at Julia code and not know how to type it yourself.

So we won't enable ligatures by default, subscribe to #2 if you want to enable them.

dralletje:
(I do however have a fix in the works that will allow you to set `font-variant-ligatures`)

fonsp:
https://github.com/fonsp/Pluto.jl/pull/590

PR #590, opened 2020-10-19
Ligatures

pankgeorg:
This is a fix for [Pluto frontend doesn't display ligature fonts despite using JuliaMono #584
](https://github.com/fonsp/Pluto.jl/issues/584)

The font loading order load the main font first, then loads the minifed (latin only) font that seems to override the first one (that has the ligatures).

First off, reading the css, I understand the _intent_ of the code is to prodive ligatures, since `.Codemirror.css`'s setting of `font-variant-ligatures: contextual;` is not overriden. 

Both the [author ](https://cormullion.github.io/pages/2020-07-26-JuliaMono/#fnref:nottheonlyone) of the font and [this guy](https://practicaltypography.com/ligatures-in-programming-fonts-hell-no.html) disagree with ligatures in general.

This fix enables the contextual ligatures
image

but not the alternative ligatures
image
as it just fixes the loading order.

Personal opinion, the slashed zero and the simple star look really cool.
It can be enabed with the `font-feature-settings: "zero", "ss05";` property (which is not part of this PR)

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -1622,4 +1622,3 @@ pluto-input:focus-within .cm-s-default .CodeMirror-matchingbracket {
     color: rgba(0, 0, 0, 0.2);
     font-style: italic;
 }
-}
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -65,15 +65,6 @@ pluto-notebook {
     display: block;
 }
 
-pluto-notebook .CodeMirror pre.CodeMirror-line,
-pluto-notebook .CodeMirror pre.CodeMirror-line-like {
-  font-variant-ligatures: none;
-}
-
-.cm-operator {
-    font-variant-ligatures: var(--pluto-operator-ligatures);
-}
-
 /* STANDARD HTML ELEMENTS */
 
 /* (can be overriden by custom style) */
@@ -1616,6 +1607,15 @@ pluto-input:focus-within .cm-s-default .CodeMirror-matchingbracket {
     font-weight: 700;
     background-color: #1b4bbb21;
     border-radius: 2px;
+}
+
+pluto-notebook .CodeMirror pre.CodeMirror-line,
+pluto-notebook .CodeMirror pre.CodeMirror-line-like {
+  font-variant-ligatures: none;
+}
+
+.cm-operator {
+    font-variant-ligatures: var(--pluto-operator-ligatures);
 }
 
 .cm-s-default.CodeMirror-empty {
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -8,12 +8,9 @@
 
 :root {
     --pluto-cell-spacing: 17px;
-    /* use the value "contextual" to enable contextual ligatures
-        for julia mono see here: https://cormullion.github.io/pages/2020-07-26-JuliaMono/#contextual_and_stylistic_alternates_and_ligatures
-    */
-    --font-variant-ligatures: none;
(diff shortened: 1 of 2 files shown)


fonsp:
We decided not to enable ligatures by default, sorry!

fonsp (changes requested):
-

pankgeorg:
I added the overriding css using a variable so a single cell can override this - aligned to how dark mode seems to be developed:

```html"<style>pluto-notebook {--font-variant-ligatures: contextual;}</style>"```
(maybe sth like this can be in PlutoUI as recipeLigature? )

Let me know of your thoughts of this. 

Thanks,
Panagiotis

**Comment by dralletje** [comment]
Okay last small thing, can you prefix the variable with `--pluto`, and make it only work on operators?
This is what I have locally:

```
.CodeMirror pre.CodeMirror-line,
.CodeMirror pre.CodeMirror-line-like {
    font-variant-ligatures: none;
}
.cm-operator {
    font-variant-ligatures: var(--pluto-operator-ligatures);
}
```

**Comment by pankgeorg** [comment]
Of course -- I updated the text to match yours.

quick question: you've removed the `pluto-notebook` the CodeMirrors override - that creates two conflicting rules with the same specificity and your style gets applied because it comes later in the DOM. where I work this is considered an anti-pattern but if you prefer it this way I'll just push it too.

Thanks,
Panagiotis

**Comment by fonsp** [comment]
Awesome! Can you move this rule to the end of the file between its codemirror buddies?

File: frontend/editor.css [hunk 1]
```diff
@@ -1606,6 +1609,15 @@ pluto-input:focus-within .cm-s-default .CodeMirror-matchingbracket {
     border-radius: 2px;
 }
 
+pluto-notebook .CodeMirror pre.CodeMirror-line,
+pluto-notebook .CodeMirror pre.CodeMirror-line-like {
```
fonsp, line 1613: use the same style selector as all other codemirror rules, can we just put this in 

https://github.com/fonsp/Pluto.jl/blob/v0.12.4/frontend/editor.css#L1467

?

frontend/editor.css
```diff
@@ -1606,6 +1609,15 @@ pluto-input:focus-within .cm-s-default .CodeMirror-matchingbracket {
...
+  font-variant-ligatures: none;
+}
+
+.cm-operator {
```
fonsp, line 1617: same here, it should be 
```
.cm-s-default span.cm-operator
```

fonsp (changes requested):

frontend/editor.css
pankgeorg: No, the rule in line 1467 does _not_ override the rule that enables ligatures, so we cannot put the code in the same codeblock, but to the spirit of your comment I added this right below it (probably that's what you meant all along):

image

and moved the operator to along its friends further below:
image

Hope that helps.

btw, I noticed that you use codemirror's _theme selector_ `cm-s-default` everywhere - so that is actually a theme customization. Of course that makes perfect sense for colors, paddings etc

I'm just writing that because when/if someone changes the theme in the future (as DarkMode seems to do) these rules will stop working and the ligatures will be enabled.

fonsp (approved):

fonsp:
Thanks @pankgeorg !

hros:
joining a few years after the initial discussion...

What's the current status? How can I enable ligatures (-> |> ...) in Pluto notebooks?

PR #592, opened 2020-10-20
FAQ link in footer

pbouffard:
Adds link to 🙋FAQ (opening in new tab) to footer on left hand side, and puts feedback form in middle and 📈 Statistics on right hand side. IMHO flows better.

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -1048,7 +1048,7 @@ export class Editor extends Component {
                         </td>
                         <td>
                         <form id="feedback" action="#" method="post">
-                            <label for="opinion">How can we make <a href="https://github.com/fonsp/Pluto.jl">Pluto.jl</a> better?</label>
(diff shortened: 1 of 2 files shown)


fonsp:
Can you write this using [CSS flex](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Basic_Concepts_of_Flexbox) instead of a table? And make sure that it works well on small screens (200px wide)

pbouffard:
Oct-20-2020 15-49-13

fonsp:
Thank you for the suggestion to include the FAQ in the footer!

It seems like we have different views about web design, which is why I decided to implement it myself. This is hard to discuss objectively, the only reason I can give is that I am very picky and closed-minded about Pluto's design. 🙃

https://github.com/fonsp/Pluto.jl/commit/0addbf76348ce68e85597d5aaaa18a40c23bcbdc

Issue #593, opened 2020-10-20
Failed to process update Error: BUFFER_SHORTAGE


GiggleLiu:
Some rare bug when I close a Pluto notebook. Just post an issue as the message says :P

```
Something went wrong!

Please open an issue on https://github.com/fonsp/Pluto.jl with this info:

Failed to process update
Error: BUFFER_SHORTAGE

[object MessageEvent]
```

Pocket-titan:
Ah now that `[object MessageEvent]` is very useful, thank you JavaScript :P I'll try to make the error display something more helpful so we can actually see it the next time it happens. Thanks for opening an issue!

fonsp:
Which browser do you use?

fonsp:
Oh and one thing @GiggleLiu :

> when I close a Pluto notebook.

Do you mean: 
1. _when you close Pluto using `Ctrl+C`, and the notebook is still open in the browser_ or 
2. _when I close a Pluto tab but leave the server running_?

GiggleLiu:
Sounds good.

I closed the tab most likely in Firefox, then I tried to open a new page in both Firefox and chrome, neither works.

2. when I close a Pluto tab but leave the server running. More accurately the sever dies. But it never happen again since then.

PR #594, opened 2020-10-20
Actually show the event that errored in the alert box

Pocket-titan:
JavaScript is not very smart at stringifying objects, so we get things like in #593 (which isn't that helpful). but `JSON.stringify`'ing the object does display it!


fonsp:
It looks like JSON doesn't show exceptions:

image

Pocket-titan:
True, but it's the MessageEvent which is stringified now, not the exception. And `JSON.stringify` only throws in two cases: when its arg has cyclic references, or when you pass it a BigInt, both of which `MessageEvent` doesn't contain (as of now) (right?)

PR #596, opened 2020-10-21
Implement duplicate line

pbouffard:
Was difficult to find an unused shortcut, maybe worsened by (possibly mistaken here) not being able to use `Ctrl` for keybindings on Mac as they all get converted to `Cmd`. I don't have a Windows machine so I'm not sure if what I chose, which is `Ctrl+Alt+Down`, is available on PC.

Oct-21-2020 15-44-56

image


pbouffard:
Sorry - messed the PR up and just realized there was a problem in the popup screen! Will create another PR...

PR #597, opened 2020-10-21
Shortcut to Duplicate a line

pbouffard:
Trying again.. 😅 Was difficult to find an unused shortcut, maybe worsened by (possibly mistaken here) not being able to use Ctrl for keybindings on Mac as they all get converted to Cmd. I don't have a Windows machine so I'm not sure if what I chose, which is Ctrl+Alt+Down, is available on PC.

image

Oct-21-2020 15-44-56


fonsp:
Can you use the same shortcut as VS Code? I'm sure its not Ctrl+Alt+Down, because that flips the screen upside down on windows 🙃

fonsp (changes requested):
Can you use the same shortcut as VS Code?

pbouffard:
> Can you use the same shortcut as VS Code?

That is the Windows & Mac shortcut; it doesn't look like there is a default keybinding in Linux: https://code.visualstudio.com/shortcuts/keyboard-shortcuts-linux.pdf

dralletje:
Copying line is shift-alt-up or shift-alt-down on both platforms, no?

dralletje:
Also can you not do the `mac_or_pc_join` thing? We don't have to feel this Mac native, as long as people understand it and it works :P

pbouffard:
> Also can you not do the `mac_or_pc_join` thing? We don't have to feel this Mac native, as long as people understand it and it works :P

I don't feel super strongly about it, just thought it might be nice. @fonsp what's your opinion?

fonsp:
I will clean this up and merge

dralletje:
Turns out, we have this now!

Issue #586, opened 2020-10-18
tabs/indenting


j-fu:
Hi, two issues with tabs:

currently, the tab key inserts a tab character (which in the editing widget seems to be equivalent
to 4 spaces). 

- I am a "spaces" guy in the tabs vs spaces flamewar. So I would prefer to have spaces instead of tabs for indenting. Kind of supported also by https://github.com/invenia/BlueStyle  . 

- Once this is settled ;) there is the question of the role of the tab key. In emacs (yes, on that side of the emacs-vs-vi debate :) ), the tab key just indents in the right way according to the surrounding code. Super cool for python, but very nice for all languages. Would this be possible here ?

I understand that these are very personal taste preferences, so there may be reasons that in Pluto preferences should be different from mine. But at  least it would be good to support some automatic indenting of code. May be it is possible to have a hotkey (like C-m for toggling markdown)  which  fixes code indenting in a cell.

pbouffard:
Maybe https://github.com/domluna/JuliaFormatter.jl could be useful?

I for one would happily allow such a tool to auto-format everything.

fonsp:
I don't want to implement a formatter yet (maybe in 6 months), but codemirror has built in auto-indent and you could add a shortcut for that.

pbouffard:
> I don't want to implement a formatter yet (maybe in 6 months), but codemirror has built in auto-indent and you could add a shortcut for that.

Just added #599 🙂

fonsp:
We will revisit this later - we don't do any customization right now.

SimonEnsemble:
following up on this, would be awesome to convert tabs to spaces!
every time I paste a worked out `function` from Pluto into terminal (vim), then push to github, the spacing is messed up! :eye:

clarkevans:
It'd be nice to let Pluto know to use 4 spaces instead of TAB character.

scls19fr:
Same problem here. Copy/paste code from VS Code or directly from Github to Pluto.jl ends up bith spaces/tabs mixed up. Issue opened https://github.com/fonsp/Pluto.jl/issues/2831

ederag:
That would not solve the copy/paste from various sources issue, but
as there is no consensus over tab vs space (I like the current defaults),
it might be nice to honor [`.editorconfig`](https://editorconfig.org/) if possible
(and as most standalone editors support it, it would make concurrent editing in both Pluto and the editor smoother).

mbauman:
[...]

PR #599, opened 2020-10-22
Shortcut for auto-indent in cells

pbouffard:
Uses CodeMirror `indentAuto` command on selection.


fonsp:
Is this the same shortcut as VS Code?

pbouffard:
Seems so on Mac and Windows, possibly not Linux: https://code.visualstudio.com/shortcuts/keyboard-shortcuts-linux.pdf

Actually I should say it's _close_ on Windows and Mac: https://code.visualstudio.com/shortcuts/keyboard-shortcuts-macos.pdf.. Ctrl+Alt+F (Windows) and Ctrl+Option+F (Mac) are actually "Format document" in VS Code while "Format selection" is a 2-step shortcut, Ctrl+K Ctrl+F. It think I saw in other threads that 2-step shortcuts were undesirable (and maybe impossible with CodeMirror?) and picked what I did.

Also, the reason to not just do "format document" (i.e. the whole cell) is that the the CodeMirror `indentAuto` command trashes any leading whitespace in multi-line strings which is not usually what you want.

dralletje:
That's a pretty stupid auto-indent 😂 
Seems a bit counter productive to have a formatting shortcut that possibly changes the behaviour of your program, if you run it while you have the wrong line selected.
I'd rather see a good formatter so we can make it "Format cell" without any surprises.. Know of a better formatter by any chance?

Pocket-titan:
We could use https://github.com/julia-vscode/DocumentFormat.jl from the `julia-vscode extension`?

dralletje:
DocumentFormat, even though it also doesn't look perfect, seems like a way better option indeed

pbouffard:
I had originally [suggested](https://github.com/fonsp/Pluto.jl/issues/586#issuecomment-713895178)  something similar but @fonsp [wants to](https://github.com/fonsp/Pluto.jl/issues/586#issuecomment-714348668) stick with a more basic auto-indent for now.

AFAIK the only issue with the current PR is how indentation in multi-line strings can get affected; IMHO code depending on leading whitespace in such strings is probably pretty brittle to begin with.

Getting DocumentFormat.jl integrated sounds like a more involved task and I'd be interested in helping but probably would need some assistance; so far I've just been nibbling at some of the easier items on the front-end side of Pluto.

fonsp:
We will not implement a formatter right now. 

> possibly changes the behaviour of your program, if you run it while you have the wrong line selected.

In that case, the shortcut should not auto-indent your selection, it should auto-indent the entire cell. Does that solve the problem?

dralletje:
How is Codemirror autoIndent not "implementing a formatter", but using a different package is? You make no sense

fonsp:
(We decided to come back to this later)

PR #600, opened 2020-10-22
Shortcut for toggle word wrap

pbouffard:
Implement toggle word wrap, include on shortcuts popup, clean up the popup a bit.

Wrap on (default):
image
Wrap off:
image

image

### DIFF
--- a/frontend/common/KeyboardShortcuts.js
+++ b/frontend/common/KeyboardShortcuts.js
@@ -1,6 +1,7 @@
 export let is_mac_keyboard = /Mac/.test(navigator.platform)
 
-export let ctrl_or_cmd_name = is_mac_keyboard ? "Cmd" : "Ctrl"
+export let ctrl_or_cmd_name = is_mac_keyboard ? "⌘" : "Ctrl"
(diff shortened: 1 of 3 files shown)


fonsp:
Sorry, this is not in line with our current design philosophy. We don't do any customization right now.

pbouffard:
Not sure I understand what you mean by customization. This just toggles the word wrap state locally and doesn't save any state anywhere; it's just making the editor a bit more capable. It would seem a shame to me to have that functionality offered up so easily by CodeMirror but not to use it.

Is there somewhere I could read up on the design philosophy so I can direct my energies accordingly?

dralletje:
@pbouffard every line of code, shortcut or change in UI we add, limits our ability to shape an opionated, well crafted editor experience. Maybe in the future there will be options to change this, but for now we don't want to worry about these things.

Adding/removing word wrap seems like a small one, and trust me I sometimes want to add it just as much as you want, but these things stack up.

Because of this we tend to err on the side of caution, sorry for demotivating you :(

You can make it into a seperate package that works with Pluto however. I do that for a lot of local stuff myself (better default log, easier package install) and there is also https://github.com/Pocket-titan/DarkMode

Issue #589, opened 2020-10-19
IPv6 address printing


staticfloat:
When printing out the "click here to connect to Pluto" message, ipv6 addresses with ports need to be surrounded by square brackets.  E.g. instead of:

```
Go to http://fd37:5040::1:1234/?secret=JqXhcn8x in your browser to start writing ~ have fun!
```
It should be:

```
Go to http://[fd37:5040::1]:1234/?secret=JqXhcn8x in your browser to start writing ~ have fun!
```

This disambiguates the IPv6 colon separator from the port separator..

PR #608, opened 2020-10-23
IPv6 Address Printing 

UtkarshChemE:
Solves #589. Checks if host address is in IPv6 format and adding squaring brackets around it.


staticfloat:
Thanks @UtkarshChemE!

PR #611, opened 2020-10-24
jsdelivr for everything

fonsp:
We use these, because they host woff2 and subset fontfaces and stuff. It seems like this should get very close to the google fonts API performance.

They publish each font as an npm package, which we get using jsdelivr

https://github.com/fontsource/fontsource/tree/master/packages/vollkorn


fonsp:
Pluto works offline now!

PR #615, opened 2020-10-27
Split up PlutoRunner.jl into seperate files

dralletje:
As the things clustered together in PlutoRunner were a mess.
I didn't change anything except moving the code to other file, adding some `import`-s and `export`-s to and from the main PlutoRunner module.

The test I changed is because now `@bind` "embeds" `PlutoRunner.Bond`


src/runner/Bonds.jl
```diff
@@ -0,0 +1,87 @@
+###
+# BONDS
+###
+
+module Bonds
```
fonsp, line 5: This could be a breaking change (at least for my funky notebooks), and I don't think that that is worth it

fonsp (changes requested):
Splitting this file is fine if it works, but let's keep it all inside a single module to avoid breaking notebooks that use internal things.

We're not missing much by having it as a single module

Issue #581, opened 2020-10-16
Low hanging fruit to make the UI faster?


fonsp:
This notebook is massive, and you start to notice UI slowdowns:

https://github.com/mitmath/18S191/blob/master/homework/homework3/hw3.jl

There is a general sluggish feel, but the easy way to see it is to drag around the selection box, since this triggers notebook state updates.

I **don't** want any crazy optimizations, but maybe there is some very simple thing to optimize? Pure components? or something?

Pocket-titan:
Idk why the link changed but it's now: https://github.com/mitmath/18S191/blob/master/homework/homework3/hw3.jl.

I'm not noticing any slowdowns with the selection area, but I do notice that adding or removing a cell is sometimes taking "too much" time (~400ms). Do these things also feel sluggish for you?

I think one way to fix this is to move these "local" notebook actions (add cell, remove cell) to the frontend, so instead of asking the server for a new cell/to delete a cell & having to wait for a response, you set the state immediately & just "notify" the server that you did these things (which can then propagate to other clients). For deleting a cell, this is relatively easy, adding a cell locally would however require you to also create the cell on the frontend (not sure if that's an issue or not). 
Trying to think of ways this could cause the UI to desync w the server, but I think we'd be good?

(If you didn't mean this by things feeling sluggish you can just ignore this)

fonsp:
Yes that's a good idea, but i meant that every notebook state update (like the docs query after every keystroke) takes too long for react to process. Maybe only on firefox? I am hoping that there is some super simple change that we can make, like `Component` -> `PureComponent` and boom its fast.

Pocket-titan:
I tested it on firefox and I'm not noticing any difference compared to chrome tbh, once every while (i.e. first time running a notebook) the docs query is "kind of slow" but after that it's good 2 go for a while (and the slowness is from the network query, probably).

Replacing `Component` with `PureComponent` could be nice, but I'm not sure if that would matter - I was thinking we could use the Preact Devtools + the browser extension (https://preactjs.github.io/preact-devtools/) to profile the Preact side, but it doesn't appear to be exported in the ESM packages & would require a bit of adjusting to make it work w web modules (at least from what I can see).

[...]

PR #626, opened 2020-11-01
Convert Cell.js to a pure component

novoselrok:
Partially addresses #581.


novoselrok:
@fonsp Can you check if this gives you a speed-up as well? The diff looks weird, but I only added the shouldComponentUpdate method and everything else is just indentation.

pankgeorg:
As I understand it, there is a bug that causes rerenders and with this fix, we just throw it under the hood. If that is so, I feel it's better to avoid this PR

novoselrok:
I'm not sure which bug you're referring to, but rerenders happen because of the `set_cell_state` function which unfortunately updates the entire `cells` array on any cell change. Refactoring the mentioned function would require quite some work, and @fonsp explicitly said in #581 that he only wants simple optimizations, which this PR is.

frontend/components/Cell.js
```diff
@@ -67,150 +67,191 @@ export const empty_cell_data = (cell_id) => {
...
-                        set_cm_forced_focus([{ line: e.detail.line, ch: 0 }, { line: e.detail.line, ch: Infinity }, { scroll: true }])
-                    } else {
-                        set_cm_forced_focus([{ line: e.detail.line, ch: ch }, { line: e.detail.line, ch: ch }, { scroll: true }])
+export class Cell extends Component {
+    constructor () {
+        super()
+    }
+
+    shouldComponentUpdate (next_props) {
+        const comparable_prop_keys = [
```
fonsp, line 76: Can we use `Object.keys` or one of its friends?

fonsp:
(Also here is the diff without whitespace: https://github.com/fonsp/Pluto.jl/pull/626/files?diff=split&w=1 )

frontend/components/Cell.js
novoselrok: I didn't want to compare all props, because some props are functions which will automatically be different on any change and the component will rerender :/

But I can use Object.keys, and automatically skip props with `typeof prop === 'function'`.

pankgeorg: Different functions should be memoized with `useCallback` and be `===` same across re-renders.

pankgeorg:
> I'm not sure which bug you're referring to, but rerenders happen because of the `set_cell_state` function which unfortunately updates the entire `cells` array on any cell change. Refactoring the mentioned function would require quite some work, and @fonsp explicitly said in #581 that he only wants simple optimizations, which this PR is.

It does update the cell array but it retains references to the same objects so that shouldn't be an issue here. The only 'new' property is line 25 of `Notebook.js` which should be memoized.

Of course I'm maybe missing something, I'm no expert in Pluto code whatsoever. But re-writing the diff algorithm for one component seems like an overkill, not a simple optimization. That's react's job.

fonsp:
React has PureComponent but we write it ourselves for silly reasons that i wont get into

pankgeorg:
And [React.memo](https://preactjs.com/guide/v10/switching-to-preact/#memo) for functional components

frontend/components/Cell.js
fonsp: Thanks for the suggestion @pankgeorg! @novoselrok implement that if you think it's appropriate, otherwise just copy your explanation as a comment into the code and its fine

fonsp (changes requested):

dralletje:
Not sure how late to the party I am, but can we use `useMemo`?

novoselrok:
> Not sure how late to the party I am, but can we use `useMemo`?

For the entire component? Not sure, I can give it a shot.

dralletje:
Yeah, like, make a seperate component (`MemoCell`) that wraps the normal cells in `useMemo` (it's like moving the `shouldComponentUpdate` outside of the cell)

pankgeorg:
> Yeah, like, make a seperate component (`MemoCell`) that wraps the normal cells in `useMemo` (it's like moving the `shouldComponentUpdate` outside of the cell)

+1 on this idea
Note that for nested objects like those with a body, you'll need to help the component understand changes. I think the best practice here is to create a new object destructuring the previous one like this:

```
setVarBody(newBody) {
  return {...this.var, body: newBody}
}
```
so it knows when to re-run the memo

pankgeorg:
Looks much better! Take a look at this change @novoselrok @dralletje https://github.com/novoselrok/Pluto.jl/pull/1 
Are there any other props that change in every render for all cells?

dralletje:
So I'd rather have the memo apart from the component:
```js
export let Cell = ({ ... }) => { ... }

export let MemoCell(props) => {
  return React.useMemo(() => <Cell {...props} />, [...]);
}
```

This way you can have the performance optimization really separate from the "real" implementation

novoselrok:
I think I included most of the necessary changes. @fonsp You can check this out locally if you have some bigger notebooks to test it on.

frontend/components/Cell.js
```diff
@@ -214,3 +214,20 @@ export const Cell = ({
         </pluto-cell>
     `
 }
+
+export const MemoCell = (props) => {
```
pankgeorg, line 218: Quick question: why are you depending on _some_ props and not all?
Even if they are stable _now_ I'd still add it

pankgeorg:
Given that this code has been included in `Notebook.jl`, should this PR get closed?

fonsp:
Yep, this was included in #710. Thank you Rok!!

Issue #631, opened 2020-11-02
High CPU load on client side with long running notebooks


dhanak:
It seems that when a notebook takes some time to calculate, even though the calculation itself is taking place on a remote computer, the browser still produces a fairly high CPU load. This is probably caused by the client-server communication in the background, but I'm sure that it should not be that expensive. This is obviously less noticeable when the Pluto server is running locally, but in that case, the above issue means that the browser is actually eating up resources that otherwise would go to Julia, which is a waste.

A fix would be very welcome!

fonsp:
(this is probably the CSS animations)

fonsp:
which browser, which version, how many cells, which CPU, how much CPU usage?

dhanak:
I'm using Chrome 86 on a Debian Linux, running on an Intel Core i7 processor with 4 cores, 8 threads, and it can use as much as 70-80% of the total CPU capacity.

fonsp:
Chrome is using 70% of a single core, or 280% total?

dhanak:
It's more like 600%! 70-80% on average on all eight hyperthreads.

fonsp:
I created this branch with animations disabled:
https://github.com/fonsp/Pluto.jl/tree/no-animations

Can you try this and see if the CPU usage is fixed?

https://github.com/fonsp/Pluto.jl/blob/no-animations/CONTRIBUTING.md#method-2-use-the-package-manager

pankgeorg:
I can confirm that the animation is the issue, on Chrome/Windows 10: 
This cell: 
![image](https://user-images.githubusercontent.com/8681967/97911497-04ce0100-1d54-11eb-9148-fe6662e2067c.png)

has this as result
![image](https://user-images.githubusercontent.com/8681967/97911480-fc75c600-1d53-11eb-8e3c-12cb1a978709.png)

Without the animation:
![image](https://user-images.githubusercontent.com/8681967/97911581-27601a00-1d54-11eb-82ba-cae56283dd90.png)

![image](https://user-images.githubusercontent.com/8681967/97911544-1adbc180-1d54-11eb-86b0-6a84f266574a.png)

fonsp:
crazy

pankgeorg:
If you agre, I'll prepare two fixes, one with stacking context, one with pseudo-elements and top animation. Will add explanation in a PR if I'm happy with the result.

dralletje:
@pankgeorg that would be amazing

[...]

PR #634, opened 2020-11-03
PJL-631 translateY keyframe animation

pankgeorg:
This pull request addresses [Issue 631](https://github.com/fonsp/Pluto.jl/issues/631) about High CPU Load with CSS animations.

- Use a variable to define the logic of `--patternHeight: 22.62741699797px`
- Set overflow of traffic-light to hidden
- Use a pseudo-element to render background image. Height of pseudo-element is `100% + 2 * patternHeight`
- Move element with animation, over the course of 1 second to 1 `patternHeight` lower.

Note: in order to run the animation, a parent of the element must have the `--patternHeight` variable set. That is if the animation is to be used elsewhere (this change makes the animation really reusable, because it _can_ be used in different settings with different pattern heights).

Result: In my computer (Chrome / Firefox on Windows 10) the cpu load drops from 60-70% to around 20%.

Let me know of your thoughts,

Thanks,
Panagiotis

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -879,6 +879,8 @@ pluto-cell.code_folded > pluto-shoulder > button > span::after {
 /* TRAFFIC LIGHT */
 
 pluto-trafficlight {
+    /* This is derived from the pattern length (16px) div sin(45deg) = 16px * sqrt(2)  */
+    --patternHeight: 22.62741699797px;
     box-sizing: content-box;
     margin-right: -1px; /* fix a visual glitch of imperfect alignment */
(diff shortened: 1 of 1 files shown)


fonsp:
@dhanak Could you try this out?

dralletje:
This is PERFECT
Just for reference

Paints before:
image

Paints after:
image

dhanak:
Wow, you guys are really quick! Kudos for the teamwork!

Pocket-titan:
Real nice @pankgeorg 🔥

PR #638, opened 2020-11-04
😴 Lazy load more items

### LINKED ISSUE #102: Interactive inspector - lazy load more items
And bring down the default number of rendered items.

It will help to send the tree structure as a JSON tree, and parse it to DOM using JS. (Instead of generating HTML code inside Julia). This would also boost performance.

fonsp:
Fix #102


src/evaluation/WorkspaceManager.jl
```diff
@@ -260,6 +256,15 @@ function eval_in_workspace(session_notebook::Union{Tuple{ServerSession,Notebook}
     nothing
 end
 
+function format_fetch_in_workspace(session_notebook::Union{Tuple{ServerSession,Notebook},Workspace}, cell_id, ends_with_semicolon, showmore_id::Union{PlutoRunner.ObjectID, Nothing}=nothing)
+    workspace = get_workspace(session_notebook)
+
```
cmcaine, line 261: eval_format_fetch_in_workspace changes directory before acting, it might be best if this function did the same: formatting code may also depend on the working directory.

PR #640, opened 2020-11-04
PLJ-636 Make the cell's runtime count up while the cell is running (#640

pankgeorg:
This PR assumes that cells are running when the `running` flag is truthy and then the `runtime` prop is set. Not sure if that's true in all edge cases (errors? other?)

It introduces a hook called `useMillisSinceTruthy` that lives in a new file within `imports` folder (that, to avoid cluttering `Cells` file with unrelated logic).

Known ~issue~ easter egg: refreshing resets counter (since it lives in client-side only). 

Interval for updating is 50 milliseconds (20 Hz / every 3 frames) which should be okey but no strong opinion (though bigger numbers seem to spoil the eagerness)


fonsp:
Very stylish!

fonsp:
#640

fonsp:
It looks very cool, thanks!

PR #652, opened 2020-11-06
Draft: Pluto tricks

lungben:
Hi,

here is the first draft of the Pluto Tricks sample notebook.
I am sure we can include more tricks, suggestions are welcome!

Best Regards
Benjamin


fonsp:
Can you undo your last commit? I don't want to install the Conda package when you open this sample notebook. Maybe it can be a separate notebook, but it doesn't really fit my fluffy world view to have a pycall sample notebook in pluto.

fonsp:
I don't want to include the python sample notebook into Pluto itself, but let's think of a better place to put it. Maybe in the PyCall repository?

lungben:
Hi Fons, 

I reverted the commit including the Python example notebook - perfectly understand your point. 
In general I was quite pleased how this works, it would be good to find another place for it - maybe only a Gist and a Discourse post (not sure if this would fit into PyCall).

For the tricks notebook I think it still needs a bit polishing and probably some more content. I want to add let-blocks, more suggestions are welcome!

lungben:
Python interop moved to here:

https://gist.github.com/lungben/b2766fb37d57cd429da05892fd26e364

fonsp:
Sorry for not getting back to you before. I'm uncomfortable with adding a sample notebook with workarounds: I would much rather fix these problems than explain them. But you are right for bringing it up, it might be good to have this in the meantime.

1. **Pkg** -- good suggestion!
2. **mutating functions** -- it might be better to suggest placing definitions and mutations inside the same cell. This is less "hacky" and avoids hidden global state. Another thing with the `sort!` example is that the reader will/should ask: "Why not just use `sort`?" A more realistic example could be: create a 100x100 array with `0` in the interior and `1` on the boundary.
3. **checkbox** -- don't put anything in the `else` block. Maybe move to the PlutoUI sample notebook?
4. This could be part of the PlutoUI sample notebook, see my comment later.
5. **linking to other notebooks** -- this is a cool trick, but I think that it is too niche to include here. (It's also not how I would solve this problem -- I would link to the notebook online, and have the reader open it manually, like we do in https://computationalthinking.mit.edu/Fall20/ )


Based on these comments, I think the focus of this sample notebook should be something like "Reproducible notebooks", and only cover Pkg, global state & options to share your notebook. What do you think?

You mentioned let blocks: I would love to have a sample notebook about this! I was planning to do a more general sample notebook about Pluto's lisp-iness, but haven't found the time.

fonsp:
We should definitely make it more visible than gist! I think that discourse is a good place, be sure to use a catchy title and some keywords in the post to make it findable through google

feanor12:
It would also be helpful to cover also the usage of `LocalResource`.
I often need to include images in the markdown section.

fonsp:
We already explain it in the PlutoUI sample notebook

fonsp:
@fonsp

PR #661, opened 2020-11-10
Pkg no GUI?

fonsp:
# OUTDATED: see #844 instead

Adventure to build a GUI for Pkg into Pluto - #142.

- It will be based on autocomplete: **if you do nothing, a reproducible, self-contained package environment will be set up**
- The GUI will be _inline_ with the `using` or `import` statement: the Pkg command and import command are fused. The GUI appears as you type (try it out!), and you can ignore it and all is well
- "Project.toml" but with support for git and local path
- Info will be stored in the notebook file's preamble (the stuff before the 1st cell). Exactly _how_ this is stored is not super important, but let's talk about this later

Very work-in-progress, just a place to post updates.

# TODO

- [ ] GUI
- [ ] Disable when you Pkg.activate
- [ ] Conversion tool for notebooks
- [ ] Actually use these versions
- [ ] Store in file
- [ ] Detect when notebook restart is needed and do it

# WIP screenshots
image

Schermafbeelding 2020-12-13 om 01 53 56


fonsp:
there's something magical about those codemirror widgets

lungben:
Maybe the notebook.assets folder from #707 could be used to store the Project and Manifest.toml files?

j-fu:
> Maybe the notebook.assets folder from #707 could be used to store the Project and Manifest.toml files?

Hi I would like to re-iterate here my idea to  use  inline toml directly in the notebook, also for more options in the cell descriptors: https://www.notion.so/Pluto-jl-a9982e79b7bb4c658e6216c15a9d4cab?p=61749c09aa1b450fa79f025d725bef4c

PR #665, opened 2020-11-11
[Frontend] Fix Alt-Down code shift bug

zerefwayne:
closes #665 

The bug was caused by the sort function used. The return type of the sort function ```(a, b) => delta * a < delta * b``` is boolean however sort function requires a number, i.e. -1 or +1. Hence ```(a, b) => delta * a < delta * b ? 1 : -1``` fixed the problem.

altfix


fonsp:
Thanks!

PR #666, opened 2020-11-11
End-to-end test improvements and fixes

novoselrok:
* Fixed the failing tests
* Improved entering text and other keyboard events into pluto input


novoselrok:
Hopefully I can get a few green commits in a row to see if the flakiness has been eliminated.

fonsp:
Would it be useful to use expose some internal things like:
https://github.com/fonsp/Pluto.jl/blob/v0.12.9/frontend/components/Editor.js#L211
or
the `.then` of the first request:
https://github.com/fonsp/Pluto.jl/blob/v0.12.9/frontend/components/Editor.js#L470

novoselrok:
@fonsp Finally managed to get a couple of green commits in a row. You can use a squash merge for this PR to avoid having my empty trigger commits in git history.

novoselrok:
Nice 👍

PR #646, opened 2020-11-05
🎞 Built-in Tables - interactive viewer

fonsp:
DataFrames will now show more rows and columns by default, in a scrollable view. More rows and columns can be loaded by clicking More. (Works with DataFrames, or any other `Tables.jl`-compatible table.)

dataframelazyload

# How to test out this branch

```julia
julia> ]
(v1.5) pkg> add Pluto#built-in-tables
julia> import Pluto; Pluto.run()
```

Then use a notebook with DataFrames as usual!

When you're done, go back to the registered version of Pluto, to make sure that you get the latest updates after this PR is done:
```julia
(v1.5) pkg> rm Pluto
(v1.5) pkg> add Pluto
```

# PR notes

To fix #292 

TODO:
- [x] Lots of stuff
- [x] Lazy load rows
- [x] Truncate columns
- [x] See how much the loading time is affected by adding the dependency. Investigate Requires.jl if it is too much
- [x] Scroll correction too much with overflowing rows
- [ ] the "... more" button for rows should stretch all columns

Can be added later:
- [x] Lazy load columns
- [x] Prettier schema

PR #672, opened 2020-11-13
Tables Interactive viewer - Sticky Headers

pankgeorg:
Builds upon #646 
- Removed `div` wrapper around Table view (was that needed?)
- add `sticky` on two first rows (types & names)
- add `sticky` (left, top) on first column's `th` 
- make `jlmore` sticky when on a row (needs `max-width`)

### DIFF
--- a/frontend/treeview.css
+++ b/frontend/treeview.css
@@ -128,6 +128,7 @@ jlmore {
     cursor: pointer;
     /* this only affects jlmore inside a table */
     width: 100%;
+    white-space: nowrap;
 }
 
 jlmore::before {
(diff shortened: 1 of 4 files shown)


frontend/treeview.css
```diff
@@ -247,6 +256,34 @@ table.pluto-table .schema-names {
...
+    left: -10px; /* padding-left of pluto-output*/
+    top: 2rem;
+}
+
+table.pluto-table td[colspan="999"] {
+    text-align: left;
+    overflow: unset;
+}
+
+table.pluto-table td[colspan="999"] jlmore {
```
fonsp, line 283: Can you add a class to the td instead of selecting with the colspan hack?

fonsp (changes requested):

PR #677, opened 2020-11-15
Add link to Jupyter-Pluto converter

DhruvaSambrani:
Add link to Jupyter-Pluto converter, with usual warnings 😄


fonsp:
Not in the readme, but you can add it to the wiki: https://github.com/fonsp/Pluto.jl/wiki

Issue #674, opened 2020-11-13
Crash on 0.12.10: MethodError: no method matching `unmake_workspace(::Pluto.Promise{Plut o.WorkspaceManager.Workspace})`


ericphanson:
I only got the stacktrace after ctrl-c'ing out of Pluto:
```julia
Worker 2 terminated.Distributed
.ProcessExitedException(2)
^C

Closing Pluto... Restart Julia for a fresh session. 

Have a nice day! 🎈

julia> ┌ Warning: HTTP upgrade failed for unknown reason
│   exception =
│    TaskFailedException:
│    MethodError: no method matching unmake_workspace(::Pluto.Promise{Plut
o.WorkspaceManager.Workspace})
│    Closest candidates are:
│      unmake_workspace(::Union{Tuple{Pluto.ServerSession,Pluto.Notebook},
 Pluto.WorkspaceManager.Workspace}) at /home/ec2-user/.julia/packages/Plut
o/SseIQ/src/evaluation/WorkspaceManager.jl:195
│    Stacktrace:
│     [1] (::Pluto.var"#219#226"{Pluto.Promise{Pluto.WorkspaceManager.Work
space}})() at ./task.jl:356
│    Stacktrace:
│     [1] sync_end(::Channel{Any}) at ./task.jl:314
│     [2] macro expansion at ./task.jl:333 [inlined]
│     [3] (::Pluto.var"#216#223"{Pluto.ServerSession})() at /home/ec2-user
/.julia/packages/Pluto/SseIQ/src/webserver/WebServer.jl:239               
│     [4] (::Pluto.var"#215#222"{Pluto.ServerSession,Base.RefValue{Functio
n}})(::HTTP.WebSockets.WebSocket{HTTP.ConnectionPool.Transaction{Sockets.T
CPSocket}}) at /home/ec2-user/.julia/packages/Pluto/SseIQ/src/webserver/We
bServer.jl:171                                                            
│     [5] upgrade(::Pluto.var"#215#222"{Pluto.ServerSession,Base.RefValue{
Function}}, ::HTTP.Streams.Stream{HTTP.Messages.Request,HTTP.ConnectionPoo
l.Transaction{Sockets.TCPSocket}}; binary::Bool) at /home/ec2-user/.julia/
packages/HTTP/IAI92/src/WebSockets.jl:160
│     [6] upgrade at /home/ec2-user/.julia/packages/HTTP/IAI92/src/WebSock
ets.jl:142 [inlined]
[...]

PR #682, opened 2020-11-16
Add missing wait()

fghzxm:
The kill-server function passed a `Promise` to `unmake_workspace` without waiting on it, resulting in a MethodError being thrown at shutdown.


fghzxm:
This should fix #674

fonsp:
😊

PR #683, opened 2020-11-16
Ignore IOError when closing connections

fghzxm:
Closing client streams can throw `IOError` if the client connection has been reset by the peer, in which case we simply ignore the error and continue the cleanup.


fonsp:
Nice! Can you rename it to `swallow_exception` and swap the function arguments?

fghzxm:
Should be OK now, I hope there aren't any minor problems left

fonsp:
Awesome, thanks!

PR #689, opened 2020-11-17
🐌 fuzzy completions

aviatesk:
Screen Shot 2020-11-18 at 1 29 12

xref: https://github.com/fonsp/Pluto.jl/issues/302#issuecomment-728808562

maybe TODOs:
- [x] ~~provide some configuration to switch between fuzzy completions and the existing REPLCompletion-based (with some customization based on `completion_priority`) ?~~ no configuration, just do fuzzy completions
- [x] make path completions fuzzy as well ?


fonsp:
We don't do configuration, fuzzy paths would be nice, if it's already part of FuzzyCompletions.jl

Project.toml
```diff
@@ -8,6 +8,7 @@ version = "0.12.10"
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+FuzzyCompletions = "fb4132e2-a121-4a70-b8a1-d5b831dcdcc2"
```
fonsp, line 11: Can you also add a `[deps]` entry?

src/runner/PlutoRunner.jl
```diff
@@ -677,15 +687,35 @@ function completions_exported(cs::Vector{<:Completion})
 end
 
 "You say Linear, I say Algebra!"
-function completion_fetcher(query, pos, workspace::Module=current_module)
-    results, loc, found = completions(query, pos, workspace)
+function completion_fetcher(args...)
```
fonsp, line 690: Remove the configuration, only do fuzzy

fonsp (changes requested):

src/runner/PlutoRunner.jl
aviatesk: done

Project.toml
aviatesk: you mean, `[compat]` entry ?
`FuzzyCompletions` is already in `[deps]` section:
```toml
```

fonsp: ah right yes

aviatesk: done

fonsp (changes requested):
It's really cool, thanks!

One problem is that empty queries no longer work. For example, `Dates.<TAB>` shows no results. Can you fix this? Autocomplete is kind of the only way to know what you can do with a package.

fonsp:
Awesome thanks! See my commit

aviatesk:
If we remove `filter!(≥(0) ∘ score, results) # too many candidates otherwise`, then it will show completions for `Module.<TAB>` case. 
How about only skipping the filtering _only when_ the query endswith `.` accessor ? It will suppress verbose completiosn for most cases, but also shows as many completions as possible on e.g. the module accessing case.

As a reference, in Juno, we skip the filtering when an user "forces" completions; i.e. Juno users can get these "exhaustive" suggestions by explicitly pressing <kbd>ctrl-space</kbd> (, which is Atom's default keybinding):
Screen Shot 2020-11-19 at 23 44 33
This tune suits the editor UI imho, because Juno (and julia-vscode) shows completions on every keystroke, and always showing verbose completions would be rather annoying (and even could be a performance problem).

Turning back to Pluto, Pluto only shows completions on explicit keystroke <kbd>tab</kbd>, so showing the "exhaustive" could be okay, but I guess an user expects more eager/correct completions if an user hits completions after some keystrokes and in that case too many suggestions would be very verbose. So imho only skipping the filtering on specific cases like `.` accessor case would be a good tune in the Pluto UI.

fonsp:
Interesting, thanks for the explanation!

I always wanted to enable autocomplete-on-every-keystroke, but felt like the autocomplete was too poor for that to be useful. But maybe we can reconsider this now that it has improved a lot.

In any case, let's merge this first!

fonsp:
Path autocomplete isn't working, it seems like the problem is that the `score` function does not work properly for `PathCompletion`s with negative score:

image

fonsp:
The problem is this:

image

fonsp (approved):

aviatesk:
> EDIT: it looks like this is on purpose? https://github.com/JunoLab/FuzzyCompletions.jl/blob/master/src/FuzzyCompletions.jl#L147

> EDIT: I have fixed this by overloading score(::PathCompletion)

yeah, but now I think this kind of tune should live in each consumer (i.e. Juno and Pluto), so I will remove `CompleteAlways` stuff from FuzzyCompletions from v0.3. Will let you know after the clean up.

PR #707, opened 2020-11-23
Pluto ux process file drop

pankgeorg:
Excel (and everything [TableIO](https://github.com/lungben/TableIO.jl) supports)

Summary of the status

- [X] Factor out Frontend logic in hook (@dralletje review)
- [X] Factor out backend logic in package (kudos @lungben !)
- [X] Notify user file is being saved (me)
- [X] Don't allow on non-empty cells with user friendly message (me)
- [X] Allow non-empty cells drop, add cell below
- [X] Drop on the whitespace results in a new cells at the bottom of the notebook
- [X] Use TableIOInterface (@lungben review + code)
- [X] Get extensions from TableIOInterface (@lungben review)
- [X] let-based code sample blocks (@dralletje review)
- [X]  drop base64 for filesave (@dralletje review)

https://user-images.githubusercontent.com/8681967/103364355-3ea65400-4ac6-11eb-8e26-4be48610f65b.mp4


lungben:
Looks great!

In Windows the file path is generated with backslashes as directory separators, which gives the following error: `syntax: invalid escape sequence`. It could be fixed by making it a raw string:

```
begin
    using TableIO, DataFrames
    df_test = DataFrame(read_table(raw"C:\Users\xxx\.julia\pluto_notebooks\test.csv",
    #, "Sheet1"  # Uncomment This if you need to read a specific sheet
    ); copycols=false)
end
```

However, I would prefer to use relative paths here so that the notebook stays portable. This could be either a temporary directory or relative to the home directory (as in your implementation).

```
begin
    using TableIO, DataFrames
    df_test = DataFrame(read_table(joinpath(homedir(), ".julia/pluto_notebooks"),
    #, "Sheet1"  # Uncomment This if you need to read a specific sheet
    ); copycols=false)
end
```

The line
 `    #, "Sheet1"  # Uncomment This if you need to read a specific sheet`
should only be shown when an Excel file is imported (currently only xlsx is supported by TableIO, but I could add at least xlsm (files with Macros)).
For CSV files in Windows, the MIME type may be set to Excel, therefore it is imho safer to rely on the file extension here. 

You could also support SQLite files (with extensions sqlite, sqlite3 or db), then adding the table name as 2nd argument to `read_table` (which needs to be filled in by the user). 

Maybe you can re-use the file extensions defined in TableIO here?
`TableIO.FILE_EXTENSIONS`, `_get_file_type(filename)` - I could add documentation and rename the latter to make them a public interface.

pankgeorg:
Thank you so much @lungben !
Yes I will add support for everything TablesIO can read.

I've noticed the CSV mime type, and the code handles that as Excel, but the TableIO _then_ handles correctly (kudos for that!)

I also agree with the relative paths:

Between 
1. `.assets` folder relative to notebook (that should be moved when you move the notebook file)
2. `~/.pluto/assets` directory in home

what do you think is best?

lungben:
I am not sure how the MIME types are set, by the browser? Excel associates itself per default with the csv extension, this gives (at least on my machine) the MIME type "application/vnd.ms-excel" for csv files. For "sqlite" files no MIME type is set at all for me. Thus I think it is safer to rely on the file extensions than on the MIME type (although not perfect).

For the storage of the imported files, I think they should be next to the notebook file, i.e. option 1. The folder should be unique per notebook so that multiple notebook.jl files in the same directory are supported, each with its own assets folder.

Suggestion: 
`my_notebook.jl` -> `my_notebook.assets`
where `my_notebook.assets` could be a folder or a zip archive. 

Not sure about @fonsp plans for the package environments (#142 ), but such an assets folder could also be used to store Project.toml and Manifect.toml files.

frontend/components/Cell.js
```diff
@@ -117,16 +120,129 @@ export const Cell = ({
...
+            () => alert("Pluto can't save this file 😥")
+        )
+        setSavingFile(false)
+        if (!success) {
+            alert("Pluto can't save this file 😥")
+            return "# File save failed"
+        }
+
+        const [name, extension] = file_name.split(".").map((a) => a.replace(/["\-,#@!\%\s+\;()$&*\[\]\{\}']/g, ""))
+        switch (file?.type) {
```
lungben, line 154: Relying on MIME types seems to be rather unstable for me, e.g. CSV files have on Windows (at least for me) Excel MIME type, SQLite db files no MIME type.
Probably just relying on the file extension would be more reliable.

frontend/components/Cell.js
```diff
@@ -117,16 +120,129 @@ export const Cell = ({
...
+ end`
+            case "application/vnd.apache.arrow.file":
+                console.log("Arrow file")
+                break
+            case "application/json":
+                return `begin
+    using TableIO, DataFrames
+    df_${name} = read_table(raw"${file_path}") |> DataFrame
+end`
+            default:
```
lungben, line 192: Could you extract the file type specific code part to a separate module so that it is easier to change / enhance?
Maybe this could even be put into a configuration file.

pankgeorg: Yes that's a good idea. Maybe it can also be part of the backend code too, and passed through the response (which can be useful if the julia part knows what modules are already imported/installed and what need to be `using`ed/`Pkg.add`ed.

lungben: If this logic is in the backend, I could extract the identification of file formats and mapping to the corresponding libraries as a separate little package (< 100 LOC, without dependencies of itself). Then both Pluto and TableIO can import this, making sure that file type mapping is always consistent and easy to enhance.

pankgeorg: I like it

lungben: First version of the file format mapping package:

https://github.com/lungben/TableFileFormats.jl

frontend/components/Cell.js
```diff
@@ -117,16 +120,129 @@ export const Cell = ({
...
+    file = open(raw"${file_path}")
+    txt_${name} = read(file, String)
+end`
+            case "application/vnd.ms-excel":
+            case "application/vnd.oasis.opendocument.spreadsheet":
+            case "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet":
+                return `begin
+    using TableIO, DataFrames
+    df_${name} = DataFrame(read_table(raw"${file_path}",
+    #, "Sheet1"  # Uncomment This if you need to read a specific sheet
```
lungben, line 173: remove leading `,`

Suggestion:
`#"Sheet1"  # Uncomment This if you need to read a specific sheet`

frontend/components/Cell.js
pankgeorg: updated -- moved this code in julia

frontend/components/Cell.js
pankgeorg: I moved the code to julia - I think that's better suited in a separate module indeed. I didn't use it in the latest update, but let's discuss!

lungben: I added the sample code generation here:
https://github.com/lungben/TableFileFormats.jl/tree/example_code_gen

Could you please check if this is working for you?

As discussed, TableFileFormats should be renamed to TableIOInterface to reflect the enhanced use case.

src/webserver/Dynamic.jl
```diff
@@ -251,39 +251,11 @@ end"""
...
 
-    elseif extension ∈ ["db", "sqlite", "sqlite3"]
-        requirements = [:TableIO, :DataFrames]
-        req_code = "using TableIO, DataFrames"
-        code = """begin
-    $(req_code)
-    df_$(varname) = DataFrame(read_table("$(path)"), "table" #== Replace this with the table ==#, copycols=false)
-end"""
+    elseif extension ∈  ["zip",  "csv",  "jdf",  "parquet",  "xlsx",  "xlsm",  "db",  "sqlite",  "sqlite3",  "dta",  "sav",  "sas7bdat",  "json",  "arrow"]
+        code = get_example_code(directory, filename)
```
lungben, line 258: it would be good if the supported file extensions are not needed to be defined in Pluto.
I could add a function in TableIOInterface to check if an extension is supported.

pankgeorg:
Summary of the status 
- [X] Factor out Frontend logic in hook (@dralletje review)
- [X] Factor out backend logic in package (kudos @lungben !)
- [ ] Notify user file is being saved (me)
- [ ] Don't allow on non-empty cells with user friendly message (me) 
- [X] Use TableIOInterface (@lungben review + code)
- [ ] Get extensions from TableIOInterface (@lungben review)
- [X] `let`-based code sample blocks (@dralletje review)
- [X] drop base64 for filesave (@dralletje review)

lungben:
I registered TableIOInterface yesterday evening, it should be available in the Registry in 3 days.
I'll add a function ˋis_extension_supported(extension)ˋ to the package probably today afternoon.

src/webserver/Dynamic.jl
lungben: I added the function `is_extension_supported(extension:: AbstractString)` in TableIOInterface.jl

lungben:
Looks great!

2 minor points:
1. Could you please rebase on master? It would be good to check if this PR works correctly with the most recent Pluto changes, especially with the automatic function wrapping.
2. When pulling a file on an empty cell, this cell is shortly marked with a non-empty cell warning before upload of the file is finished.

pankgeorg (commented):
> Looks great!
> 
> 2 minor points:
> 
> 1. Could you please rebase on master? It would be good to check if this PR works correctly with the most recent Pluto changes, especially with the automatic function wrapping.
> 2. When pulling a file on an empty cell, this cell is shortly marked with a non-empty cell warning before upload of the file is finished.

1. I rebased to master - can't say I understand if it works or breaks (Seems to work)
2. Silly me! Fixed! Thanks for catching!!

pankgeorg:
Update: A function-like test seems to break

lungben:
Cells which contain `import` are not wrapped into a function call, see

https://github.com/fonsp/Pluto.jl/pull/720/files#diff-04c7252262a0e6de2db2cded7aba9b1130c1e5ba6adff25bf8ec3a34e36a4788R928-R946

Therefore, in principle at least the TableIO templates should still work.

dralletje:
Good job on putting the javascript in its own file

lungben:
Works perfectly for me, thanks!!!

lungben:
One thing is missing: if the notebook is renamed / moved using the Pluto functionality, the .assets directory should be moved, too.

pankgeorg:
Very good point Benjamin; I'll try to fix this - I think I know how

lungben:
One more small thing:
when moving a Pluto cell with the mouse, the new error message "This cell is not empty" is shown for the adjacent cells.
You can still move a cell when you are careful to put your cursor exactly in the middle between 2 cells, but this is a bit tricky.

image

fonsp:
Some comments:
- Use `PlutoUI.LocalResource` for images, audio & video instead of `Images`+`ImageMagick`
- The red aggressive scary warning is something that _we_ should solve, not the user
- Pluto has a function `numbered_until_unique` that we can use for the local filename, I don't think it's a feature that old files with a clashing name are overwritten

What do you think?

lungben:
Maybe to make it really fancy:
if you drag a file to a cell where a file with the same filename is already imported, we could ask whether to replace the existing file or to add it as a new file. This could be useful when updating or correcting input data.
But this is rather a nice-to-have feature for me.

pankgeorg:
-> Added 4 lines that move the assets folder along with the notebook!

Benjamin, I'll fix the message too.
Fons, got the message for the drag & drop message too :) - I'll try to add the blue message and create a new cell (and remove the red message completely).

Also, LocalResource it is!

Also, files that don't have special code will get copied and we'll add a filename = "path_to_file" string so the user know what to do next. 

I'll also try the `numbered_until_unique` for extra drops of the same file

src/webserver/Dynamic.jl
```diff
@@ -197,4 +196,66 @@ responses[:reshow_cell] = (session::ServerSession, body, notebook::Notebook, cel
...
+        Dict(
+            :success => success,
+            :code => code
+        ), notebook, nothing, initiator)
+
+    putclientupdates!(session, initiator, msg)
+end
+
+get_template_code = (filename, directory, iofilecontents) -> begin
+    path = string(raw"$(dirname(@__FILE__))", "/", directory, "/", filename)
```
lungben, line 233: As discussed, please use this format so that it still works after moving / renaming the notebook.

`joinpath(split(@__FILE__, '#')[1] * ".assets", "my_file.jpg")`

pankgeorg: Note to self: make sure the name change the dir name changes too (maybe that happens already)

fonsp:
oops sorry

lungben:
Is anything still missing here?
It would be great to have this feature merged.

fonsp:
Can you add a `[deps]` entry for tableiointerface?

Project.toml
```diff
@@ -17,6 +17,7 @@ MsgPack = "99f44e22-a591-53d1-9472-aa23ef4bd671"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
+TableIOInterface = "d1efa939-5518-4425-949f-ab857e148477"
```
lungben, line 20: @fonsp do you mean this [deps] section?

fonsp: ah i meant `[compat]`

lungben: I suggest to add
`TableIOInterface = "0.1"`
to the `[compat]` section.

pankgeorg:
https://github.com/fonsp/Pluto.jl/issues/950
more features

PR #709, opened 2020-11-25
fix anonymous function canonalization, fixes #654

### LINKED ISSUE #654: Anonymous function using `function(x)` unrecognised
```julia
function image_skybox(img)
	f = function(position, skybox)
		lon = atan(-position[1], position[3])
		lat = -atan(position[2], norm(position[[1,3]]))

		get_index_rational(img, (lat/(pi)) + .5, (lon/2pi) + .5)
	end
	
	SkyBox([0.0, 0.0, 0.0], 1000, f)
end
```

gives

```
┌ Error: Failed to canonalize this strange looking function
│   ex = :((position, skybox))
└ @ Pluto.ExpressionExplorer ~/Pluto.jl/src/analysis/ExpressionExplorer.jl:827
```

Pangoraw:
Hi ! :wave: 

This fixes #654 

However this method will not work with the current duplicate methods detection but an assignment cannot declare a new method and only override the variable :thinking:

```julia
julia> f = function() end
#7 (generic function with 1 method)

julia> f = function(a::Int) a end
#9 (generic function with 1 method)

julia> methods(f)
# 1 method for generic function "#9":
[1] (::getfield(Main, Symbol("##9#10")))(a::Int32) in Main at REPL[11]:1
```

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -819,10 +819,11 @@ canonalize(e1) == canonalize(e2)
 function canonalize(ex::Expr)
 	if ex.head == :where
 		Expr(:where, canonalize(ex.args[1]), ex.args[2:end]...)
-	elseif ex.head == :call
-		ex.args[1] # is the function name, we dont want it
+	elseif ex.head == :call || ex.head == :tuple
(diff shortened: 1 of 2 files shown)


fonsp:
I was hoping that you would fix this one!

PR #710, opened 2020-11-25
Dralbase state managment

dralletje:
A rewrite of Pluto's internal state management that makes it work more like Google Firebase:

All clients and the server all have a local, mutable copy of the global shared state. Anyone can mutate their local state, get local effects instantly, and everything will be updated in the background.

## Client
The main advantage is that actions like 'add cell', 'delete cell' will have instant effect on the local, visual state. This would get rid of the small delay or these actions, caused by a round trip of information.

## Server
The server can be written in a more declarative style. Instead of manually handling each message (e.g. 'add_cell', 'delete_cell') with specialized code ([example](https://github.com/fonsp/Pluto.jl/blob/v0.12.11/src/webserver/Dynamic.jl#L36-L66)), there is only one function: `do_effects(old_notebook_state, new_notebook_state)`. When the server updates a cell output, it does just that, and Dralbase will figure out how to send this to the clients efficiently.


fonsp:
image

I navigated to samples while it was still loading, then went back, and now it's stuck loading until i refresh

fonsp:
image

One cell is stuck running but I can run other cells, refreshing did not fix

Are these helpful or no?

fonsp:
folding a cell does not trigger a file change

fonsp:
Bonds keep going for too long:

Schermopname 2020-12-11 om 19 06 20

fonsp:
Code changes are not synced between clients

fonsp:
`New cell -> change code to 123 -> Ctrl+Enter` gives an error sometimes

fonsp:
Old Pluto loaded cell outputs one-by-one instead of all at once, to improve time-to-first-markdown

Schermopname 2020-12-11 om 20 31 50

fonsp:
Clciking "Split into 3 cells" inside the error is broken

fonsp:
```js
notebook: /** @type {NotebookData} */ ({
    notebook_id: new URLSearchParams(window.location.search).get("id"),
    path: default_path,
    shortpath: "",
    in_temp_dir: true,
    cell_dict: {},
    cells_running: {},
    cell_order: [],
}),
cells_local: /** @type {{ [id: string]: CellData }} */ ({}),
```

Questions:
- Why is the cell state split up in two dicts?
- We could have better names, maybe `cell_inputs` `cell_outputs` instead of `cell_dict` `cells_running`? Same for `CellData`, `CellState` (sound like synonyms) and the structs inside Dynamic.jl
- What is `cells_local`? Why is it not part of the notebook state?

dralletje:
Anwsers:
- the two dicts represent different ownership: one for the users that write in notebook, one for the worker that computes outputs
- They should have better names
- cells_local is your yet unsaved cell input. I want to eventually sync that as well, so you can see when someone is editing etc. But instead of having "local_code" and "remote_code" in cells, it is "local_cells" and "(remote_)cell"

fonsp:
Thanks! And why split on ownership?

fonsp:
TODOs?

- [ ] Fix frontend tests
- [ ] Run Firebase notebook tests on CI
- [ ] Rename Firebase to something else like Dralbase or Firebasey
- [ ] Remove logging code from Dynamic.jl
- [ ] Delete old tests from test/Dynamic.jl

dralletje:
1. They represent different thing: one is the question, the other the answer.
2. Syncing gets easier this way, we can be sure that updates the worker make don't collide with updates the clients make

fonsp:
Copy pasting multiple cells (select, ctrl+C, Ctrl+V) is broken

fonsp:
Clicking "More" in the array viewer does not work:
```julia
MethodError: no method matching response_reshow_cell(::Pluto.ServerSession, ::Dict{Any,Any}, ::Pluto.Notebook; initiator=Pluto.Initiator(Pluto.ClientSession(Symbol("1lhyvdz"), HTTP.WebSockets.WebSocket{HTTP.ConnectionPool.Transaction{Sockets.TCPSocket}}
```

fonsp:
Cells that output images are broken I think

fonsp:
VET
 
Schermopname 2020-12-20 om 23 27 44

fonsp:
Bonds should lose their value when you run their definition:

image

image

pankgeorg:
There is a very strange bug where variables don't get defined even though they are defined.
This is from the front-end tests. Writing here 
image
if any of you have any christmas ideas, if (and hopefully) not, let's discuss in 2021!

fonsp:
@dralletje I have merged this PR to get my mind cleared but we still need to do the testing and documentation that I listed [here](https://github.com/fonsp/Pluto.jl/pull/710#issuecomment-743866161) before we tag the next release.

Thankyou!

Issue #379, opened 2020-09-07
Global and Local Scope inside the for and while loop


Dsantra92:
The notebook throws a `UndefVarError`  while accessing the global variables inside the for loop.
![error_log](https://user-images.githubusercontent.com/55111154/92412862-2a32fa00-f16b-11ea-9286-9f5b8e576021.png)

shashankp:
Its not an issue with the for loop as such

![image](https://user-images.githubusercontent.com/498813/92525792-15756580-f242-11ea-8b95-21f12e424fc4.png)

fonsp:
You need to keep all assignments and modifications (`+=`) within a single cell - otherwise reactivity can't work.

So:
```julia
begin
	s = 1
	for i in 1:10
		s += i
	end
	s
end
```

but this still won't work, because `for` **introduces a scope**. If you assign to `s` within the `for` loop, then you are assigning to a local variable, also called `s`, that is only defined within that loop. If you want to read more, this is called an _ambiguous assignment inside a soft scope_:
https://docs.julialang.org/en/v1/manual/variables-and-scoping/#Local-Scope

The solution is to use `local`:
```julia
s = begin
	local s = 1
	for i in 1:10
		s += i
	end
	s
end
```

or:

```julia
s = begin
	local my_s = 1
	for i in 1:10
		my_s += i
	end
	my_s
end
```

But `for` and `while` loops have so-called "functional" alternatives. The big ones are `map`, `filter`, `reduce` and `sum`:
```julia
begin
	s = 1
	s += sum(1:10) do i
		i
	end
end
```

Functional programming works very nicely together with Pluto - try it out!

fonsp:
Sorry for all that strange syntax!

Dsantra92:
@fonsp thank you for this wonderful clarification.

PR #720, opened 2020-11-26
🐰 Wrap code in functions to make it faster

fonsp:
We turn a simple cell like `p = plot(x,y)` into a _function_ so that it can be called a second time without compilation!

It takes the referenced globals as arguments, and returns the definined globals. `p = plot(x,y)` would become:

```julia
function(plot, x,y)
	result = p = plot(x,y)
	(result, (p,))
end
```

Function generation is here:
https://github.com/fonsp/Pluto.jl/pull/720/files#diff-0cc97f3d6a0f647a05e5f913d416242ecb00f6e67e12004a7204b8846a3fa44cR96-R104

and it is called here:
https://github.com/fonsp/Pluto.jl/pull/720/files#diff-0cc97f3d6a0f647a05e5f913d416242ecb00f6e67e12004a7204b8846a3fa44cR112

---

Not every cell will be run as a function. Cells that import, define a function, or type, or call a macro still run with the old method:
https://github.com/fonsp/Pluto.jl/pull/720/files#diff-04c7252262a0e6de2db2cded7aba9b1130c1e5ba6adff25bf8ec3a34e36a4788R928-R946

---

And the best part is: **all of this is completely invisible**! You get superfast code without having to rewrite anything ❤


fonsp:
Right now I did the rewrite but no cell is being function wrapped yet

fonsp:
Getting there!

image

fonsp:
TODO:
- [ ] Remove the tip to wrap your code inside a function in the Interactivity notebook 😊
- [ ] Measure performance boost
- [ ] Measure increase in RAM usage?

fonsp:
Somewhat extreme example to make it more visible on video:

Before:



After:

fonsp:
@NHDaly this is what we talked about a long long time ago 😊

src/runner/PlutoRunner.jl
```diff
@@ -50,33 +57,156 @@ function set_current_module(newname)
...
+"Wrap `expr` inside a timing block."
+function timed_expr(expr::Expr, return_proof::Any=nothing)::Expr
+    # @assert ExpressionExplorer.is_toplevel_expr(expr)
+
+    linenumbernode = expr.args[1]
+    root = expr.args[2] # pretty much equal to what `Meta.parse(cell.code)` would give
+
+    @gensym result
+    @gensym elapsed_ns
+    # we don't use `quote ... end` here to avoid the LineNumberNodes that it adds (these would taint the stack trace).
```
lungben, line 149: you could use `MacroTools.striplines(ex)` for removing LineNumberNotes from expressions generated by `quote` (if this would be beneficial for you).

lungben:
looks cool, I'll try it out tomorrow!

fonsp:
@lungben Let me know if you find anything, but I'm pretty sure that it's done! Yayyy

lungben:
looks good so far!

a)
```
function do_stuff(x)
	y = 0.0
	for i = 1:100000000
		y += x/i
	end
	y
end

do_stuff(x)
```

b)
```
begin
	y = 0.0
	for i = 1:100000000
		global y += x/i # global keyword may be obmitted in this PR because of implicit function wrapping
	end
	y
end
```

Timings:
without this PR:
a) ~112 ms
b) 10s (`global y` is required)

with this PR:
a) ~112 ms
b) 10s (with `global y`) / 112 ms (without `global y`)

Observations:
* I tried a few of my own notebooks and nothing crashed :-)
* if just a function is executed in a cell, the timings are the same (especially, no performance loss)
* the scoping rules change for loops inside a cell because they are now implicitly inside a function - this may be unexpected, but is not bad imho, it is essentially the same as the new Julia 1.5 REPL scoping (which is currently not used in Pluto).
* removing the global keyword in the loop gives with this PR the same performance (a factor of 1000 better!) as manually wrapping it into a function. 

Suggestions:
* It would be good to have a way to define "local" variables, which are not returned to global scope. This could be helpful when dealing with large amount of data to allow GC to clean up intermediate results. We could use the Python convention of variables starting with underscore (e.g. `_x`) for private variables which are not exported. This would somehow complement let blocks, where only the return values are exported and everything else is local.
* A blog post / discourse, etc. on this change would be helpful - understanding this change allows users to write more efficient notebooks (e.g. by using function scoping instead of globals).

fonsp:
About local variables: this is why Julia has `local`: https://github.com/fonsp/Pluto.jl/issues/379#issuecomment-689130899

lungben:
Thanks!
Using `local` for `y` the code runs in 8s without this PR and 112 ms with this PR 😄 
Furthermore, the `local` keyword prevents the variable from being exported to global scope, therefore my suggestion above is already possible without me recognizing it.

fonsp:
That's right, 40x faster than the REPL!

image

NHDaly:
🎉!! @fonsp this is awesome, I'm super glad to see this improvement! 🎉  Thanks for pinging me! :)

🤩 Amazing results!! 🤩

NHDaly:
Also, wow, it's so cool how _much_ faster this got.

Hehe technically this isn't including the compilation time in the time result, but that's standard for what julia does for `@time` (to time the _actual_ time including compilation, you can time `@eval`, which can be important if your code actually takes potentially several minutes to compile and then time reports it ran almost instantly):
```julia
julia> @time 2+2
  0.000000 seconds
4

julia> @time @eval 2+2
  0.000970 seconds (53 allocations: 3.516 KiB)
4
```

But still, after this change, even if you include compilation time, things get much faster because you've changed a non-const global into a function argument. For the old case before this PR, julia had to check every iteration of the loop that `y`'s value hadn't changed (because it's just a global, and it could have been affected by e.g. another thread somewhere or something), and in the new case, it can rely on the value being a constant, which matches the reactive notebook paradigm much better! :)

So even if you include the compilation time, the new system is still faster than evaling the block old way! 🎉 
Screen Shot 2020-12-15 at 11 52 49 AM
Screen Shot 2020-12-15 at 11 53 49 AM


------

I guess Pluto could consider to separately report the compilation time as well? Julia 1.6 has started doing that: 
https://github.com/JuliaLang/julia/pull/37678
(Although they are missing some "top-level" compilation time, as reported here: https://github.com/JuliaLang/julia/pull/37938. Pluto is better positioned to get this report correct, if that's desirable, but i'm not sure whether it is or isn't. 🙂)

Issue #241, opened 2020-07-30
Artifacts to proxy&cache CDNs


staticfloat:
It would be nice for startup time (especially on a slow connection) to bundle all web assets locally; if this is a relatively slowly-moving target, this can be done easily by using Artifacts; simply write a script that collects all files into a single location inside of `create_artifact()`, archive the artifact into a tarball, upload that tarball somewhere, bind the location of the artifact to a name in your package, and you'll be able to serve the files from a tarball that gets downloaded at the same time as your package.

As a reference, these are the assets that my browser requests after a force-refresh on Pluto `v0.11.1`:

![image](https://user-images.githubusercontent.com/130920/88981873-cb30ba00-d27b-11ea-905d-c7c1c19a455a.png)

fonsp:
All these assets are cached by a service worker, so the second load should even work offline!

fonsp:
But I am very interested in the artifacts system - do you think we can use it to make our project independent of the CDNs? Ideally you could run an old version of Pluto 20 years from now, when (JS hype has moved on to something else and) some of these CDNs might have gone offline.

staticfloat:
> Improvements to the first load time are welcome of course, but they might be overshadowed by Julia compiling Pluto's methods. There is still the use case of deploying Pluto notebooks online - this would be useful there.

Ah, that makes sense, yes.  I was working on a tethered phone, so it took many seconds to fetch those resources the first time.  A pathological case for sure.

> do you think we can use it to make our project independent of the CDNs?

Yes, definitely!  I'd be happy to help you walk through it; the short version is that you would run a script to bundle together your content and upload it as a tarball.  Then in `Pluto.jl`, you would have an `Artifacts.toml` that points a name (such as `javascript_bundle`) to that file, and when you want to access, for instance, `js/foo.js`, you'll just do something like `joinpath(artifact"javascript_bundle", "js", "foo.js")` and it will reach inside of the artifact (which will have been installed by Pkg automatically when `Pluto.jl` itself was installed, and it will have been installed by the Pkg server, so the tarball will have been cached for all eternity in the Julia Storage Servers, not to mention hosted wherever you chose to host it (I recommend a GitHub release somewhere).  So you can then manually serve out those files.  I imagine you could still use a service worker but attempt to load from the server it's being stored on first, and in the event of a 404, then you could grab it from a CDN, and in that way you would get the best of both worlds.

c42f:
[...]

PR #722, opened 2020-11-27
Prevent leaking the secret to cross origins

gregod:
Currently the secret is exposed to third parties (CDN , Google) through the referrer sent for the linked asset requests. _"Referrer : http://pluto-server:1234/?secret=XXXX"_

This is a quick-fix proposal to set a Referrer-Policy on each http request. The header instructs browsers to send only the origin part of the url as referrer on cross origin requests. 

For a better fix (and better offline support #241 )  i would suggest removing the external dependencies altogether.  This could either be done via "vendorizing" them by keeping a local copy (lowest initial maintenance impact) or switching to a build script for the front end (requires build step). Let me know if you want me to contribute something.

### DIFF
--- a/src/webserver/WebServer.jl
+++ b/src/webserver/WebServer.jl
@@ -210,6 +210,7 @@ function run(session::ServerSession)
             request.response::HTTP.Response = response_body
             request.response.request = request
             try
+                HTTP.setheader(http, "Referrer-Policy" => "origin-when-cross-origin")
                 HTTP.startwrite(http)
                 write(http, request.response.body)
                 HTTP.closewrite(http)


fonsp:
Thanks! Let us know if you find anything else :)

We will probably solve 241 by 'scraping' for asset URLs and bundling those into a Julia artifact. The frontend service worker will then use the Pluto backend as a cache (I think).

fonsp:
I changed this to [`same-origin`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy) (which is more restrictive) in https://github.com/fonsp/Pluto.jl/commit/e6b55d1c285b3cc07721eb21b241b5dca46aa136

Issue #621, opened 2020-10-30
End-to-end test improvements and additions


novoselrok:
Fix tests failing when they shouldn't:
* https://github.com/fonsp/Pluto.jl/runs/1332947231

Additional functionality to test:
* Key shortcuts

fonsp:
https://github.com/fonsp/Pluto.jl/runs/1355798132

Downloading Ubuntu gave 402 Payment required ? computers are strange

novoselrok:
@fonsp Not sure how, not sure why, but this commit: https://github.com/fonsp/Pluto.jl/commit/2e561e9ad31aade638ab92125420a5f99f92c407 broke the e2e tests. I also confirmed it locally. e2e tests are working on the previous commit.

fonsp:
It take about 5-10secs for Julia to precompile the autocomplete stuff, which I triggered to happen on page load. So this means that the backend is unresponsive for 10 seconds longer than before after launch

PR #728, opened 2020-11-28
add checks to bind macro

### LINKED ISSUE #486: CheckBox() not working in session after coding error
Hi!

I was using CheckBox() from PlutoUI but I accidentally wrote `$(@bind cost1 CheckBox(), default=true)` instead of `$(@bind cost1 CheckBox(default=true))` and apparently you can't use CheckBoxes anymore in that session (even if removing that line of code in the cell) so I had to restart the session. The error is

```
┌ Warning: unknow use of `=`. Assignee is unrecognised.
│   ex = :(CheckBox())
└ @ Pluto.ExpressionExplorer ~/.julia/packages/Pluto/upwxH/src/analysis/ExpressionExplorer.jl:160
```

Pangoraw:
Hello! :wave: 

This prevents executing code that could break some functions by defining new methods.

Example:
```julia
@bind a CheckBox(), default=true
```

Would throw an error before executing where it would not have
previously, redefining `CheckBox()` to `true` (see #486).

There seems to be currently no tests for `@bind` (https://github.com/fonsp/Pluto.jl/blob/master/test/runtests.jl#L23).
Should I add some ? 

Fixes #486

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -1060,7 +1060,8 @@ The first cell will show a slider as the cell's output, ranging from 0 until 100
 The second cell will show the square of `x`, and is updated in real-time as the slider is moved.
 """
 macro bind(def, element)
-	if def isa Symbol
+    is_element_valid = element isa Symbol || (element isa Expr && element.head ∉ [:(=), :(::)])
+	if def isa Symbol && is_element_valid
 		quote
 			local el = $(esc(element))
             global $(esc(def)) = Core.applicable(Base.get, el) ? Base.get(el) : missing


fonsp:
Thanks! But I would rather fix the underlying issue with ExpressionExplorer: https://github.com/fonsp/Pluto.jl/issues/486#issuecomment-735245821

fonsp:
About tests: this is difficult because you need to run a browser, but it is being worked on: https://github.com/fonsp/Pluto.jl/issues/621

fonsp:
For example, this code:
```julia
@bind x myslider=Slider(1:100)
```
is legal, as is:
```julia
@bind x y::Slider
```

The bug in our expression explorer meant that the user-generated function `CheckBox` did not get removed, otherwise it would have been fine.

Pangoraw:
You're right ! I had not fully understood the scope of the problem. Sorry for the time wasted

fonsp:
No worries, always happy with your contributions!

Issue #712, opened 2020-11-25
Sticky Headers break when column name is large


pankgeorg:
![image](https://user-images.githubusercontent.com/8681967/100276252-ee3d5300-2f69-11eb-95fb-2ed8d50f4050.png)
becomes
![image](https://user-images.githubusercontent.com/8681967/100276294-fbf2d880-2f69-11eb-8f30-1aa85dad530e.png)
(note the white lane)

That's the type row trying to get on top or something along that
Will investigate

fonsp:
There's also this: these cells should not wrap

![image](https://user-images.githubusercontent.com/6933510/100475756-354f5380-30e4-11eb-9966-29cf4ecf8dff.png)

I remember finding this issue when playing with the table styling myself, which is why i gave up on sticky headers :)

pankgeorg:
Ok I'll look into this with priority; wrapping is super-weird tbh

fonsp:
If there is no header then the first row index is playing hide-and-seek

![image](https://user-images.githubusercontent.com/6933510/100480215-e956db80-30f0-11eb-8856-39659d4ae403.png)

PR #729, opened 2020-11-28
PLJ-712 Sticky Headers Fix

pankgeorg:
Addresses #712 

- Only make names sticky (because counting header style for types' top is not worth the javascript overhead)
- Move transform within th to prevent a stacking context creation and allow for headers z-index to work (first row above first column)
- _Should_ work for headerless tables too (removed top from columns)
- make sure to  `white-space`: `pre` in `tables` > `pre`


dralletje:
Very nice

PR #757, opened 2020-12-09
When text is selected and either (, [ or { is pressed wrap selection in (), [], {} instead of replacing it.

pankgeorg:
This PR adds key bindings to codemirror in order to handle the press of `(`, `[` or `{` more intuitively (citation needed).

When user has done a text selection and presses `(`, `[` or `{`, the character _replaces_ the text.

With this PR, the behaviour changes to wrap the selection in brackets.

Note that the cursor is moved after the text  `(selected-text)<HERE>` vs for example VSCode that goes to `(selected-text<HERE>)`.

Also remove the line that activates ''autoCloseBrackets" (because it doesn't!)

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -291,10 +291,14 @@ export const CellInput = ({
                 return CodeMirror.Pass
             }
         })
-        const open_close_selection = (opening_char, closing_char) => (cm_) => {
-            for (const selection of cm_.getSelections())
-                if (selection?.length) cm_.replaceSelection(`${opening_char}${selection}${closing_char}`)
-                else cm_.replaceSelection(opening_char)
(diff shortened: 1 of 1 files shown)


yuhan0:
Hi, I looked into this a few days ago too, and found that uncommenting this line 
https://github.com/fonsp/Pluto.jl/blob/031770d727fc824c085d4803ea0ff01a5e7ea1c1/frontend/editor.html#L31
was enough to enable the feature of wrapping selected text in brackets, along with auto closing brackets.

Unless there was a reason for reimplementing this feature, I think it's better to just use the native codemirror addon?

pankgeorg:
The native codemirror addon does more things than what this implementation provides.

This just adds open/close brackets when you **have a selection** and click parenthesis. Does not open/close in all cases.

PR #801, opened 2020-12-23
WIP Gist notebooks

ctrekker:
This is a work in progress pull request for implementation of Gist notebooks. Feedback is welcome as I continue to develop this feature!


fonsp:
This is amazing!!!! The one-click auth with redirect is perfect, the autosaving works, really cool!

Some things:
- Auth in a popup window would be better
- Why is the gist ID a textbox? Showing the link to the gist somewhere is enough maybe?

I guess the next goal is to replace file UI with gist UI. For example:
- In the main menu: How do I see my Pluto gists, and how do open them?
- In the notebook: The file picker at the top should be the gist name. Gist should not be an extra feature (like the export menu), but an alternative to storing the notebook as a file. Maybe a toggle next to the file box?

I think we want to redo the file UI anyways (typing absolute paths is bad) @pankgeorg , and maybe it’s best to first get something small out the door before we replace too much. But so far it looks very good!

The gist API is called from the frontend right? That seems like the way to go.

ctrekker:
In terms of the UI design choices, I imagined that a system similar to export would work best with the current way files are handled, but if thats going to change then I could see how gists being a different save medium altogether would be better.

* I agree auth in a popup would be much better!
* The gist id is its own textbox because each gist can contain multiple files
  *  For the sake of UI simplicity it might be better if we only supported 1 notebook file per gist? That way gist notebooks could be identified using only a filename rather than an ID and a filename

I'll start working on the idea of gists being a disparate save medium rather than a secondary save step for filesystem notebooks. I'm imagining a github icon next to the gist filename in the "recent sessions" section to distinguish gist saves from local saves.

ctrekker:
Also yes, the gist API calls are made entirely from the frontend. There is one backend endpoint I made for saving the github OAuth token, but I should be able to remove that soon

fonsp:
Sounds great! Definitely restrict to file per gist, the goal here is to make the "file UI" as simple as google docs. 

Another thing we could do to make this simpler is show both options when you press "Save notebook..." on a new notebook, and afterwards it's local or gist forever (you can always switch by going to gist.github.com yourself and uploading/downloading/deleting/etc). Fewer options is often better!

fonsp:
(I just merged a big PR on master, let me know if you need help rewriting your code)

ctrekker:
I think I got it rewritten enough to where it works. I still have some work left surrounding the autosave efficiency and robustness under the new state management system.

frontend/components/Editor.js
```diff
@@ -402,6 +406,9 @@ export class Editor extends Component {
         // these are update message that are _not_ a response to a `send(*, *, {create_promise: true})`
         const on_update = (update, by_me) => {
             if (this.state.notebook.notebook_id === update.notebook_id) {
+                if(this.state.save_medium) {
+                    this.state.save_medium.scheduleSave()
```
ctrekker, line 410: I'm not sure if this is where this line should go under the new state management. This line schedules a notebook to be saved to gist. It is debounced, so it can be called many times without causing problems. However it should only be called by the client who made the changes. Otherwise saves will be scheduled on every open client corresponding to a session. Is that what `by_me` tracks? 

If so then `if(this.state.save_medium && by_me)` would fix this?

fonsp:
This is AWESOME

fonsp:
Like I mentioned in the zulip, can you make a separate branch without gist, but with native file API, and once the user gives permission, we hide the existing UI and just use two buttons: open and save.

ctrekker:
> Like I mentioned in the zulip, can you make a separate branch without gist, but with native file API, and once the user gives permission, we hide the existing UI and just use two buttons: open and save.

I used the SaveAdapter abstraction that I wrote in this PR to make this work. Could I branch from this PR and just delete the GistSaveAdapter class from that branch? That would remove all the gist stuff but keep the SaveAdapter abstraction

fonsp:
Exactly! On GH Desktop you just click "New branch" with the gist branch currently checked out

Pocket-titan:
> Sounds great! Definitely restrict to file per gist, the goal here is to make the "file UI" as simple as google docs. In fact, could you look into gdrive integration? The big limitation with gists is its useless UI: you can't search your gists (what!), no folders, no collab, no forking.

I just randomly found this PR (which is amazing!) BUT just wanted to mention that gist _does_ support forking + searching (yours or others') gists (not that it's very accurate though). Also gists can be added to git/Github desktop & managed just like a normal repo (which is cool) (but no collab 😢). Just wanted to mention that before someone ends up spending too much time wrestling w google's api & finding out gists are pretty ok 🙃

PR #805, opened 2020-12-23
PLJ-785 CodeMirror Eject to TextArea - Reload from TextArea in Offline HTML export

pankgeorg:
This PR changes the way the app loads CodeMirror: It creates a textarea in `CellInput` from which the CodeMirror is created.

On export, we get the code written in the CM into a fresh textarea, in the clone body. When the HTML reloads, the textarea gives its place to a CM instance.

Client JavaScript required.


fonsp:
Can you (or @fonsp) run this, and post the resulting .html file as an example?

pankgeorg:
https://gist.github.com/pankgeorg/e3f0ea302fa4fa081034451ba68fde8a

fonsp:
Looks great! I can't copy text though

pankgeorg:
> Looks great! I can't copy text though -- are these real codemirrors with input disabled, or standalone-mode outputs like highlighted code inside markdown?

real codemirrors with input disabled (and cursor disabled)
I can re-open the cursor to enable focus & copy (but then people will see a cursor which is like telling them - hey you can write - which they can't)

fonsp:
Ah now I understand. I guess that showing the cursor is the lesser evil? That would look like https://fonsp.com/article-test-3/mycoolnotebook.html

Maybe we can set the blink speed to 99999999 or something hacky like that: https://github.com/codemirror/CodeMirror/search?q=cursorBlinkRate

pankgeorg:
> Ah now I understand. I guess that showing the cursor is the lesser evil? That would look like https://fonsp.com/article-test-3/mycoolnotebook.html
> 
> Maybe we can set the blink speed to 99999999 or something hacky like that: https://github.com/codemirror/CodeMirror/search?q=cursorBlinkRate

-1 seems to work perfectly!

PR #816, opened 2020-12-31
Regard `:=` in macros as einsum notation (take 2)

mcabbott:
This is another, simpler, attempt at my initial idea in #407. It interprets macro calls like `@einsum A[i] := B[i,j]` as assignment to `A`.

A small example notebook, which gives errors without this PR: 
https://gist.github.com/mcabbott/cf49093313612b084614b90d4f96af0d

The function `strip_indexing` makes this not quite the minimal version. It ensures that Pluto doesn't think the result depends on `i,j` (if these are defined elsewhere) since they are loop variables. If this is removed, it should still work without errors, but sometimes updates things unnecessarily.

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -261,6 +261,18 @@ end
 assign_to_kw(e::Expr) = e.head == :(=) ? Expr(:kw, e.args...) : e
 assign_to_kw(x::Any) = x
 
+"Turn `A[i] * B[j,K[l+m]]` into `A[0] * B[0,K[0+0]]` to hide loop indices"
+function strip_indexing(x, inside::Bool=false)
+    if Meta.isexpr(x, :ref)
+        Expr(:ref, strip_indexing(x.args[1]), strip_indexing.(x.args[2:end], true)...)
+    elseif Meta.isexpr(x, :call)
+        Expr(x.head, x.args[1], strip_indexing.(x.args[2:end], inside)...)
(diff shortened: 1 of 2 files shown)


mcabbott:
Searching on JuliaHub to see if there are packages which use this symbol for some other purpose, I found:
* https://juliahub.com/docs/RecipesBase/8e2Mm/1.1.1/syntax/ uses it for defining rules
* https://juliahub.com/docs/RBNF/nX819/0.2.2/#Structures-of-Defining-A-Language for something like struct definitions
*https://juliahub.com/docs/QuickTypes/csU4i/1.6.3/autodocs/#QuickTypes.@destruct-Tuple{Expr} uses it in a docstring but not in code example
* https://juliahub.com/docs/FilteredGroupbyMacro/1Dk08/0.1.0/#Assignment-syntax-1 uses it to define functions?

None of the examples of these involve indexing-notation, so the only affect of this PR will be that `:=` is replaced by `=` before further processing inside ExpressionExplorer.

fonsp:
Thank you for your patience with me! This is a good middle ground, I'd like to reconsider your other PR in the next couple of months, handling macros correctly is one of the last big challenges left on the julia side of pluto

fonsp:
Let me know if you have more ideas to make low-disruption changes to add support for major packages

mcabbott:
Sure! The general case sounds super-tricky, and I totally understand the reluctance to commit too-complicated stopgaps.

PR #833, opened 2021-01-09
Built-in browser local file saving in Chrome

ctrekker:
File handles for a user's local files can be obtained upon request via the [File System Access API](https://web.dev/file-system-access/#status). This API is currently only available on Chrome, so as a fallback the current method of file saving will be used instead.


fonsp:
To repeat what we discussed in the Pluto call:

**This is AWESOME, and we will use it for binder soon!**

We won't enable it by default for the registered Pluto yet, because we only get the file, not its location. Notebooks that depend on other files and use code like `./data/friends.csv` will not work :( Not yet sure how to solve this smoothly

PR #844, opened 2021-01-14
🎁 Built-in Pkg management

fonsp:
Schermafbeelding 2021-01-18 om 01 30 49

**Watch a [15min video explaining this PR](https://www.youtube.com/watch?v=hVeW_JjmOKQ), or read the [documentation](https://github.com/fonsp/Pluto.jl/wiki/%F0%9F%8E%81-Package-management)**

# Table of contents
1. Motivation
2. Try it out
3. More details
4. Screenshots
5. TODOs
6. Description of changes

# 1. Motivation
The two main goals are:
> 1. **Package versions inside the notebook file 👉👉 _by default_!**
> 2. `using Example` should install `Example`

This feature is designed with _beginner programmers_ in mind, but I think that it will cover most use cases.

## Example
A notebook with only this code:
```julia
using JSON
```
```julia
json([1,2,3])
```
will have the used version number of JSON.jl **stored in the notebook file**. Loading a Pluto notebook will instantiate from this data. So:
1. Package dependencies are reproducible
2. You do not need to use the Pkg REPL to install JSON

---

# 2. Try it out! 
(Use Julia 1.5, 1.6 or nightly (1.8))
```julia
pkg> activate --temp
pkg> add Pluto#experiment-pkg-ui-2
```
Now, open a new notebook **(not an existing notebook!!)** and import some packages!

---

# 3. More details

I want to offer a limited subset of Pkg functionality directly inside Pluto, mostly automated behind the scenes, to achieve these two goals. People who understand `Pkg` are not the intended audience, but there will be a way to opt out and use `Pkg` commands like before.

- It will be completely automatic: **if you do nothing, a reproducible, self-contained package environment will be set up**
- Project.toml and Manifest.toml inside the notebook file as plaintext. The file format will be forwards-backwards-compatible
- By using the magic command `Pkg.activate` anywhere in your code, Pluto's package management will be completely disabled, and you get the old behaviour. This way, notebooks that already use `Pkg` calls to manually set up a package environment will continue to work. This also means that we don't need to implement every feature to make advanced users happy.
- The previous PR added a GUI to choose versions. After talking with @staticfloat and @StefanKarpinski, they made the brilliant suggestion _not_ to have a version-picker GUI, but to just use the latest version, with unobtrusive UI. We will still use the inline GUI to show status (when a package can be updated, for example), and maybe also for `] dev`.

### Notebook load times

This change will also make notebook loading faster, compared to "`Pkg.add` notebooks" like in https://computationalthinking.mit.edu/ because:
- Notebooks don't need to _resolve_ a manifest from just the list of dependencies. The manifest is fixed and does not need to resolve when you load a notebook.
- Pkg handling is done on the server process, so only one registry update will be triggered if you launch multiple notebooks.
- (Advanced) we can do some tricks to make sure that all homeworks use the same versions of dependencies by modifying their embedded manifests. This means that the second homework loads faster, and we can precompile packages on binder.
- (Advanced) in projects like PlutoBindServer.jl and static notebook exports, we currently can't safely launch all notebooks in parallel because of https://github.com/JuliaLang/Pkg.jl/issues/2219#issuecomment-761172547. This PR implements a "Pkg lock" (within the server process), which means that multiple notebooks with plutopkg can launch in parallel, improving export times.

---

# WIP screenshots
(i did not have these two packages installed before!)

https://user-images.githubusercontent.com/6933510/104786674-dd50d700-578d-11eb-8a3b-cdc665d87c04.mov

https://user-images.githubusercontent.com/6933510/110706592-687fa380-81f8-11eb-84c1-d5ce3e6fa1b9.mov

I LOVE removing documentation, this is from the new PlutoUI sample notebook:
image

# _(advanced)_ Edit notebook environments using Pkg REPL
I am adding a function `Pluto.PkgUtils.activate_notebook("path_to_notebook.jl")` that activates a notebook Pkg environment in the REPL. It uses FileWatching to sync changes from either side (i.e. Pkg REPL changes are written to the notebook, editing the notebook updates the Pkg REPL env).

https://user-images.githubusercontent.com/6933510/111483040-b9c9ee80-8734-11eb-97bc-f51703fc57cc.mov

For more (outdated) screenshots, see #661 

---

# 5. TODO
## TODO - UI

- [x] Write a help page -- https://github.com/fonsp/Pluto.jl/wiki/%F0%9F%8E%81-Package-management
- [x] Link to the help page from the popup
- [x] Update sample notebooks
- [x] Inline GUI mechanism (#661)
- [ ] Show a message next to `Pkg.activate` about disabled PlutoPkg
- [x] Show available updates in GUI -- https://github.com/JuliaLang/Pkg.jl/issues/2582
- [x] Update button
- [x] Update terminal output
- [x] Don't show GUI for local modules
- [x] Show installed version with GUI
- [x] Show installed version with GUI in static export
- [x] Show to-be-installed version with GUI
- [x] Autocomplete for all registered packages
- [x] 'Busy' icon next to installing package
- [x] Tell user that registry update is happening
- [x] Detect when notebook restart is needed
- [x] Communicate notebook restart urgency to user
- [x] Experiment with relaying IO to user -- see https://github.com/fonsp/Pluto.jl/pull/1210
- [x] Show that another notebook is blocking Pkg

## TODO - backend

- [x] Ability to restart a notebook
- [x] Add packages when you run `using Asdf`
- [x] Ability to update packages
- [x] Remove packages when the last `using Asdf` is gone
- [x] Ignore local modules
- [x] Disable when you use `Pkg.activate`
- [x] Actually use these versions
- [x] Store Project.toml + Manifest.toml in file
- [x] Pkg.instantiate the loaded notebook environment
- [x] Pkg.resolve the loaded notebook environment and remove the Manifest.toml if that makes the env resolvable
- [x] Remove compat entries for packages that became stdlibs
- [x] Use the backward-forward-compatible format that I came up with after I implemented the backward-compatible one
- [x] Write automatic `[compat]` entries to Project
- [x] Conversion for old notebooks
- [x] Show notebook while Pkg is being loaded
- [x] Set `LOAD_PATH` to `["@"]`
- [x] Reset `LOAD_PATH` when PlutoPkg is disabled
- [x] What to do when Pkg throws an exception? (Solution: Write a backup, clear the embedded Project+Manifest and require notebook restart.)
- [x] Backup & continue on corrupted embedded Manifest/Proejct
- [x] What to do with Tables.jl -- ~stack? requires.jl? copyofrequires.jl?~ #895
- [x] What to do when `workspace_use_distributed == false`? Seems fine to ignore, lets revisit later
- [x] What to do with Markdown and InteractiveUtils? Seems fine to ignore
- [x] What to do when you `Pkg.add` without `Pkg.activate`? (solution: disable plutopkg)
- [ ] Think about first "New notebook" launch time
- [x] Test on Pkg Julia 1.7
- [x] Benchmark new Dynamic.jl fields, maybe cache

## TODO - tests

- [x] Create a custom registry for testing -- https://github.com/JuliaPluto/PlutoPkgTestRegistry
- [x] Tests for auto-installing packages
- [x] Tests for installing packages later
- [x] Tests for loading a notebook with manifest
- [x] Tests for file format -- bijective
- [x] Tests for file format -- backwards compat with Pkg calls
- [x] Tests for file format -- backwards compat without Pkg calls -- should write backup file?
- [x] Tests for file format -- forwards compat

## EXTRAS

- [x] Don't recommend restart for stdlibs
- [x] Don't recommend restart when nothing got removed from the manifest
- [x] Update sample notebooks
- [x] Plots.jl can't have a Manifest for every supported Julia version
- [ ] Try on windows
- [x] Util scripts
- [x] Hide Pkg io during tests
- [x] A way to update the registry? Nah
- [x] Write Julia version in Project.toml? No. Will be in Manifest.toml [starting in Julia 1.7](https://github.com/JuliaLang/Pkg.jl/pull/2561)
- [x] What are custom registries? Do I already support that? Maybe someone from Pkg/juliacomputing can help?
- [x] We currently assume that package names are unique. Is this a problem? How can I test that?
- [x] We use unexported Pkg API, what to do with that?
- [ ] Get people to try it out

## FUTURE

- Show available updates in GUI and warn if it cannot be installed because it is version-bound? -- https://github.com/JuliaLang/Pkg.jl/issues/2583
- "Version xx will be installed in this notebook" should respect constraints -- https://github.com/JuliaLang/Pkg.jl/issues/2582#issuecomment-851970637
- Factor out `Pkg.Types.Context`
- What to do when `workspace_use_distributed == false`?
- Show progress bar?
- Detect broken registry?
- Package from local path
- Package from git repo
- Add no-longer-supported warning for Julia 1.0 - 1.4
- Test "Pluto inside Pluto" on windows -- [`5eb81bf` (#844)](https://github.com/fonsp/Pluto.jl/pull/844/commits/5eb81bfb1f7e4cca47138834c2dc57c758f5bc31)
- Any future Pkg features to take into account?

---

# 6. Description of changes

Most of the added lines are because I updated our sample notebooks to use the new format, which includes a lengthy Manifest for the Plots.jl sample notebook. (907 additions are for sample notebooks)

## Backend
- `src/packages/Packages.jl` is the most important. It contains the "Pluto package manager".
- `test/packages/Basic.jl` to get a more high-level overview of the new behaviour.
- `src/packages/PkgCompat.jl` contains calls to (internal) Pkg API.
- `src/notebook/Notebook.jl` contains new code to read and write the Pkg data in the notebook file.

## Frontend
- `frontend/components/PkgStatusMark.js` is the little checkmark/icon that appears inline next to a package name. It is not a Preact component, but a real DOM element, because that works better inside a codemirror widget. This file also contains the helper text code.
- `frontend/components/PkgPopup.js` is the popup.
- `frontend/components/CellInput.js` contains nasty code to detect `using` and `import` statements and insert and update status marks inline. Note that this is just a visual frontend, it has no functional role in the package manager.


frontend/components/Notebook.js
```diff
@@ -129,7 +140,7 @@ export const Notebook = ({
                     key=${cell_id}
                     cell_result=${notebook.cell_results[cell_id] ?? {
                         cell_id: cell_id,
-                        queued: false,
+                        queued: true,
```
dralletje, line 143: Smart

oschulz:
Hope this is in the right place here:  I've wondering whether the "Pluto-internal" package management might be able to use PackageCompiler behind the scences (mainly transparent to the user), to speed things up when notebooks get loaded several times. Maybe at least for some packages like Plots.jl (if present in the notebook). I guess that would probably be something for the future, after Pluto's internal package management is stable and been used in the wild for a while. But I thought I'd add the idea here for future reference, in case other's have thoughts in this direction already.

fonsp:
@oschulz Thanks for the suggestion! That's definitely interesting, but like you say, let's talk more about that after a couple of months.

oschulz:
@fonsp , should go ideas like this here (or issues in general) or to notion.so ?

fonsp:
notion.so would be best! we found that github issues is great for technical discussions with an agreed-upon objective, but less nice for exploratory discussion and design problems. you can also find us on zulip: https://gist.github.com/fonsp/db7d00fd3fe5bc0b379b4af9ec6674b6

Pangoraw:
Awesome work! :confetti_ball:  

I am not sure if it is something only on my computer but the UI is different in Firefox (version 87.0) compared to Chrome. Just reporting so you can know!
image

briochemc:
> I am not sure if it is something only on my computer

Just to confirm it's not only your computer, I see it too on FireFox 🦊 

Screen Shot 2021-03-30 at 11 44 12 am

MasonProtter:
I'd like to argue against the use of the `Manifest.toml as I think this is liable to lead to gigantic bloat in `.julia`'s package version cache, as well as making users unnecessarily encounter pre-compilation much more often.

For reproducibility in bug reports, you could keep a `Manifest.toml` in the Pluto notebook, but instead of using that manifest to force the usage of those specific package subversions, you instead read from the enclosing environment what versions are being used and update the `Manifest.toml` based on that. 

That is, I propose that the `Manifest.toml` be used as a (updating) record of what exact package versions are used, rather than the way it currently forces the package manager to pick specific versions.

crstnbr:
Minor "bug" report:
```Julia
[ Info: Instantiating
0 dependencies successfully precompiled in 0 seconds (37 already precompiled)
75 dependencies errored
┌ Info: PlutoPkg: success!
└   pkg_result = (did_something = true, used_tier = Pkg.Types.PRESERVE_ALL, restart_recommended = false, restart_required = false)
```
Precompilation seems to have failed but `PlutoPkg` didn't notice?

crstnbr:
> as well as making users unnecessarily encounter pre-compilation much more often

During the PlutoCon presentation I was thinking of this point as well. As much as I like this feature, the way it is currently implemented seems to take the concept of a julia environment from a folder level to a file level. I'd argue that, typically, one has one Julia environment per project, which consists of multiple julia files / notebooks that use the same environment, the same packages. With the proposed implementation it seems that in such a case every julia file / notebook contains a separate julia environment. I really wonder how this interacts with pre-compilation.

> That is, I propose that the `Manifest.toml` be used as a (updating) record of what exact package versions are used, rather than the way it currently forces the package manager to pick specific versions.

While I agree with the sentiment, this will lead to even more precompilation on notebook startup because versions of dependencies of dependencies might have / will likely have changed.

fonsp:
> Minor "bug" report:

Can you try to create a minimal example of how to reproduce this problem? It is not really a problem, precompilation failure is fine, as it is just an optimization, but maybe my code is doing something that causes the failures.

fonsp:
For the Pkg developers slack:

> [Kristoffer Carlsson](https://github.com/KristofferC): The whole system is designed in a way that having multiple projects should work well. If it doesn't, it is up to us to fix it. As an example, one such fix was https://github.com/JuliaLang/julia/pull/32651. I personally don't really worry about any bloatness from storing the manifest inside notebooks as you describe. Most of the versions of common dependencies would be the same and share precompilation files. 

Read the full discussion here: https://gist.github.com/fonsp/22a11cd9c03573e1134a6ee91d11c452

MasonProtter:
Ah, I didn't know that the caching system is designed to not grow forever. Certianly if if the Pkg devs say it's a good idea, I wouldn't presume to disagree. Thanks for the reference.

nluetts:
I like the idea of auto-installing packages and keeping track of the environment via an embedded Project.toml+Manifest.toml, especially in combination with the easy opt-out for more advanced usage—just activate your own external environment. I have two questions regarding this feature:

1. In the PlutoCon talk you showed that activating an external environment will make Pluto delete the Project/Manifest info from the notebook file immediately, without further notification. I think sometimes the user might then unintentionally delete the embedded environment information because they are not aware that Pkg.activate silently removes it. Perhaps you could add a notification in the form of a dialog box or a warning symbol next to Pkg.activate (like the download notifications when `using` a package).

2. Say I have a notebook that I want to share, but it is part of a larger project so it does not come with the environment info embedded. How would I embed this environment info? I suppose I could copy/paste the contents of Project/Manifest.toml, but maybe you could add an export option “export with environment” (or similar; I guess this would have to comment out Pkg.activate in the exported notebook ...)?

fonsp:
For 1: good idea! Added to my TODO

For 2: this sounds too niche/hypothetical for us to worry about right now

greimel:
> 2\. Say I have a notebook that I want to share, but it is part of a larger project so it does not come with the environment info embedded. How would I embed this environment info? I suppose I could copy/paste the contents of Project/Manifest.toml, but maybe you could add an export option “export with environment” (or similar; I guess this would have to comment out Pkg.activate in the exported notebook ...)?

This feature could be useful in combination with custom sysimages.

Imagine you have a set of notebooks for a course. You want these notebooks to be simple and standalone for students that run them locally. But you also want the notebooks to be runnable on binder for the students with technical problems. So you want the notebooks to share a Manifest so that you can provide a single sysimage for all notebooks.

fonsp:
Right! This feature was absolutely **designed for that exact use case**. So for educational material, I very highly recommend using the built-in package manager --  it's very easy to use, and the best guarantee you can get that your notebook will work on students' local setups, **binder**, the new export features, the new sliderserver, etc. It is my attempt to solve most technical issues that remain in our course https://github.com/mitmath/18S191, which currently need my Pluto-background knowledge to solve.

I will be sure to create utilities to generate a sysimage from a notebook file that uses the built-in Pkg manager. We did a lot of experiments with running and optimizing notebooks for binder, and having a built-in package manager makes it much easier, and it means that we can apply our optimizations automatically to notebooks, because we fully control the environment in every stage of the process.

StatisticalMice:
I tried this out as suggested above. It didn't work, so here's a bug report.

```
...
  [8e850ede] + nghttp2_jll
  [3f19e933] + p7zip_jll
[ Info: PlutoPkg done
[ Info: Instantiating
  0 dependencies successfully precompiled in 0 seconds (10 already precompiled)
  125 dependencies errored. To see a full report either run `import Pkg; Pkg.precompile()` or load the packages
┌ Info: PlutoPkg: success!
└   pkg_result = (did_something = true, used_tier = Pkg.Types.PRESERVE_ALL, restart_recommended = false, restart_required = false)
```

Screenshot 2021-05-07 at 12 27 51

Do you need more information?

fonsp:
@StatisticalMice Don't add the `.jl` when importing a package, I guess I can add a check for that

StatisticalMice:
> @StatisticalMice Don't add the `.jl` when importing a package, I guess I can add a check for that

For some reason I made a typo. GLMakie seems to work now, it produced a plot, but the error message about "125 dependencies errored." still exists. I exited Pluto, reloaded the notebook, and it is still there.

Is this the suggested reaction to the error message on the console?

Screenshot 2021-05-07 at 14 08 04

Console:
```
Closing Pluto... Restart Julia for a fresh session. 

Have a nice day! 🎈

Opening http://localhost:1234/?secret=6Pe0hH87 in your default browser... ~ have fun!

Press Ctrl+C in this terminal to stop Pluto

to_add = String[]
[ Info: Instantiating
  0 dependencies successfully precompiled in 0 seconds (10 already precompiled)
  125 dependencies errored. To see a full report either run `import Pkg; Pkg.precompile()` or load the packages
┌ Info: PlutoPkg: success!
└   pkg_result = (did_something = true, used_tier = Pkg.Types.PRESERVE_ALL, restart_recommended = false, restart_required = false)
to_add = String[]
to_add = ["Pkg"]
   Resolving package versions...
    Updating `/private/var/folders/2s/qc19sf6x6q99ds44rfhy46dw0000gn/T/jl_FrE84i/Project.toml`
  [44cfe95a] + Pkg
  No Changes to `/private/var/folders/2s/qc19sf6x6q99ds44rfhy46dw0000gn/T/jl_FrE84i/Manifest.toml`
[ Info: PlutoPkg done
[ Info: Instantiating
  0 dependencies successfully precompiled in 0 seconds (10 already precompiled)
  125 dependencies errored. To see a full report either run `import Pkg; Pkg.precompile()` or load the packages
┌ Info: PlutoPkg: success!
└   pkg_result = (did_something = true, used_tier = Pkg.Types.PRESERVE_ALL, restart_recommended = false, restart_required = false)
Precompiling project...
  4 dependencies successfully precompiled in 55 seconds (131 already precompiled)
``

**Comment by fonsp** [comment]
@StatisticalMice thanks, this was already reported here https://github.com/fonsp/Pluto.jl/pull/844#issuecomment-817106594

**Comment by fonsp** [comment]
@ericphanson I stumbled on https://github.com/JuliaLang/Pkg.jl/issues/2525, which means that you are the ideal person to help me!
- Do we already support custom registries? (Can I `import MyCustomPackage` and it works?)
- What happens when a package is registered in two registries? Is this the desired behaviour?
- Is registry information stored in the manifest? Can others run your notebooks?

**Comment by ericphanson** [comment]
Hi @fonsp, happy to help!

> Do we already support custom registries? (Can I import MyCustomPackage and it works?)

Yep! `Pkg.Registry.add(RegistrySpec(url="...", name="..."))` (or even just `] registry add url`) works fine to add the registry. That only has to be done once and it clones the registry to `~/.julia/registries/MyRegistryName`. Then it will be available for any `Pkg.add`, `Pkg.dev` etc commands, and will be just like using a package from General. (In fact, General is only special in that it is pre-installed and can be installed by name without a URL `] registry add General`, but otherwise I think is treated just like another registry).

> What happens when a package is registered in two registries? Is this the desired behaviour?

If the package has the same name and UUID (== it's the same package, from Pkg's point of view), then `Pkg.add("PkgName")` will grab the latest version from whatever registry it is in; it works very seamlessly. However, `Pkg.add` with a branch or URL, or `Pkg.dev`, will seemingly choose an arbitrary registry. (That's what https://github.com/JuliaLang/Pkg.jl/issues/2525 is about). This can be very confusing for users, especially if they aren't expecting any issues.

If the package has the same name and a different UUID in two registries (== it's a different package but there's a name clash, from Pkg's point of view), then when you do `] add MyPackage`, it will open a little interactive menu in the REPL to choose which registry you mean. At Beacon, our internal registry does not have any packages named the same as General, so I haven't experienced this myself but it's what I've heard from some of the Invenia folks who have a big internal registry.

If the package has a different name and the same UUID in two registries, I'm not sure what happens 😅 . Probably you get an error somewhere.

> Is registry information stored in the manifest? Can others run your notebooks?

No, it is not. The Manifest entries look just the same as the ones from General. So you need to tell users to install the registry first (or maybe even do a `Pkg.Registry.add` in the notebook).

> Is it easy for me to set up a custom registry to test this PR? Does such a 'testing registry' already exist?

[LocalRegistry.jl](https://github.com/GunnarFarneback/LocalRegistry.jl) makes it really easy to create registries, and you don't even need to make a github repo, etc, although you can. There is not a testing registry setup, but if you are interested in Pkg's interaction with a registry (rather than actually registering new packages there etc), then it might be enough to experiment with a public registry like https://github.com/HolyLab/HolyLabRegistry.

**Comment by fonsp** [comment]
@ericphanson Thanks for the info! LocalRegistry.jl looks really useful for testing!

Did you try out this PR with your custom registry? Instructions are at the bottom of the PR description.

**Comment by ericphanson** [comment]
> Did you try out this PR with your custom registry? Instructions are at the bottom of the PR description.

Just gave it a try, and it worked perfectly with a package from our registry! I also tried with a package that has older versions in our registry and newer versions in General (Lighthouse.jl) and it installed the latest version as I hoped too. Really cool stuff!

I also tried adding a `using Pkg; Pkg.Registry.add` cell and that worked fine (did not move it out of the new notebook package management mode which is good). I also tried a cell
```julia
if !occursin("OurRegistryName", sprint(io -> Pkg.Registry.status(; io=io)))
	Pkg.Registry.add(RegistrySpec(...))
end
```
and that seemed to work fine too. So that might be a way to make notebooks with other registries more portable.

fonsp:
That's great! Though the `Registry.add` trick won't currently work, as we do package management before running notebook code, but this would be one reason to opt-out of PlutoPkg by calling `Pkg.activate`.

fonsp:
@ericphanson Thanks for the tips! I made https://github.com/JuliaPluto/PlutoPkgTestRegistry for testing

fredrikekre:
My `using` cells blew up like this (using Firefox):
Screenshot from 2021-06-10 23-34-42

fredrikekre:
`using Example` when there are multiple packages `Example` in the registry hangs (from the perspective of the notebook) since the prompt will show up in the terminal waiting for user input.

DilumAluthge:
> `using Example` when there are multiple packages `Example` in the registry hangs (from the perspective of the notebook) since the prompt will show up in the terminal waiting for user input.

Is there a way for the user to "drop into" a terminal and provide the user input manually?

fonsp:
@fredrikekre thanks! first one is fixed

fonsp:
@DilumAluthge @fredrikekre Can you point me to the Pkg source code that asks for input?

fredrikekre:
https://github.com/JuliaLang/Pkg.jl/blob/6c98f54bd16b315e732b2d21e7cb6e2cf1f89b70/src/Types.jl#L939-L947

fonsp:
@DilumAluthge based on the code, it looks like the prompt and input happen on the terminal where you launched Pluto, so you should be able to make the choice there, but I don't see an easy way to:
- detect a prompt and show a message in the browser to go to the terminal
- relay the prompt itself to (a terminal in) the browser

I think the best solution for now is to try to set temporarily `isinteractive()` to false on the server. How common is this same-name-different-uuid collision?

fonsp:
**I am merging this to `main` so that more people can try it out, and to rebase other PRs. But please continue experimenting, and continue leaving feedback!**

We will probably tag the `0.15` release with the built-in package manager on July 1st, and https://github.com/fonsp/Pluto.jl/tree/release-0.14 can be used to backport fixes to `0.14` in the meantime.

dralletje:
WHAAAATTTTTTTTTTTTTT

AshtonSBradley:
In trying this out (woop woop!), I found myself wanting to also use Pkg version syntax like `using Makie@0.14.1` to generate a specific manifest.

Possible?

fonsp:
> use Pkg version syntax like `using Makie@0.14.1` to generate a specific manifest.

@AshtonSBradley We can break this in two parts:
- the ability to specify a version. I talked about this a bit in the PR description, but we might add it back later at some point.
- inputting a version using that syntax: the notebook file must be a legal `.jl` file, so not exactly that, but we can do something keyboard-friendly

Issue #875, opened 2021-01-27
Package Manager Not Loading


BrettKnoss:
Since updating to v0.19, my package manager, and all packages will not load.  I'm using Julia 1.6.0 beta, and Vivaldi | 3.5.2115.87 (Stable channel) (64-bit), which uses JavaScript | V8 8.7.220.31, on a x86-64 bit machine running Ubuntu 20.04.1 LTS.

fonsp:
Can you send more information?

BrettKnoss:
I've had the same results with my book, but this is your plots sample:

![Screenshot from 2021-01-27 16-07-13](https://user-images.githubusercontent.com/73511645/106061004-6a632c80-60ba-11eb-9ceb-e8f80a58edc1.png)

I also have this error message:

```
[5948:5948:0127/160618.351734:ERROR:chrome_content_client.cc(343)] Failed to locate and load the component updated flash plugin.
Opening in existing browser session.

```
,but I think it was there when everything was working.

bolgarbe:
Can confirm with Pluto 0.12.19 on Julia 1.6.0-beta1 with Ubuntu 20.04.1 + Chromium, Firefox and Vivaldi. Notebooks load fine but hang immediately after (i.e. can't add or run code cells). The only message I get is

`┌ Warning: Response function to message of type complete failed
└ @ Pluto ~/.julia/packages/Pluto/A96Bh/src/webserver/WebServer.jl:363`

However, it works with `Pluto.run(workspace_use_distributed=false)`.

lungben:
Same issue for me with Windows 10 (and Julia 1.6 beta).

fonsp:
There is no output in the terminal where you launched Pluto besides https://github.com/fonsp/Pluto.jl/issues/875#issuecomment-768990215 ?

fonsp:
I made some changes that hopefully fix this problem, can you test it using:

```
pkg> activate --temp
pkg> add Pluto#60f6b2b
julia> import Pluto; Pluto.run()
```

and let me know whether that fixed the problem

lungben:
It hangs for me in this line:

```
Pkg.instantiate(pkg_ctx) # we don't suppress IO for this one because it can take very long, and that would be a frustrating experience without IO
```

lungben:
When I run the same commands from Loader.jl locally, it works:

```
[...]

PR #883, opened 2021-01-28
Merge branch 'master' of https://github.com/lungben/Pluto.jl

lungben:


lungben:
This should fix https://github.com/fonsp/Pluto.jl/issues/875

lungben:
Disabling auto precompilation should not have a performance impact here because the Pluto dependencies should be already precompiled (either when installing or when importing Pluto).
Therefore, I think this fix should not introduce side-effects.

PR #886, opened 2021-01-31
Fix anonymous function declaration

### LINKED ISSUE #776: `function() end` not recognised

Pangoraw:
Another small fix :wave: 

Removes a no-op instruction which was accessing the first element of
the expression's args, crashing for anonymous functions with no
parameters like `function() x end`.

Fixes #776


fonsp:
🙈 thank you!!

fonsp:
image

PR #888, opened 2021-02-01
Notebook progress bar

fonsp:
https://user-images.githubusercontent.com/6933510/106509813-5eb6a200-64ce-11eb-8880-502ec41e074a.mov

https://user-images.githubusercontent.com/6933510/106509826-624a2900-64ce-11eb-8c45-31324b253361.mov


fonsp:
The binder feature that I added also has a progress bar, which creates a (non-trivial) conflict. Ways to resolve it:

- Add two progress bars on top of each other 🌈
- Make the progress bar more general: a system for any piece of code to say "i will do work that takes approx. x seconds", "work y is at z%" and "work y is done".

I think the first solution might be best for now: they use the same css class ✔ and the binder load will not coincide with cells running ✔️

Issue #13, opened 2020-02-25
Visualize cell reactive dependencies


fonsp:
Suggestion:
When focused on a cell, all cells that reference that cell get highlighted. (Either with depth 1, or with full depth.) This makes it easy to see what reevaluating the cell will do.

Vice versa might also be useful.

malyvsen:
Variables could have a nice `rectangular box with rounded corners` around them. The color of the box would depend on the section that the variable was defined in (section being determined by last markdown header).

fonsp:
Still not sure how to do this with CodeMirror, maybe you need to define your own `mode`?

karlwessel:
Highlighting depending cells won't work as nicely in a large notebook since most of the depending cells might be out of view. 

I'm not sure anymore about solving this by a sort of birds view like I suggested in #116 . My impression from similar features in editors, where for example the scrollbar is a map of the code, is that it's clutters the UI more than it really helps.

Instead one could visualise a kind of weight of a cell based on the number of cells depending on it or maybe the estimated sum of execution time of all cells depending on it.

I'm not sure how to visualise the weight intuitively. Simplest would be a number in some corner giving the total execution time. This way you also get a simple profiling mechanic. 
Giving heavy cells a stony look and light cells a cloudy look is probably to childish/confusing :-).

I also like the idea from malyvsen to give sections or maybe areas of a notebook different (background?) colours, so you get an intuitive feeling where you are ("I'm currently in the blue section of my notebook"). In that theme, the weight of a cell could be visualised by a rainbow of the colours of depending cells. But that might again be hard to get right without becoming distracting or childish looking.

karlwessel:
The stony/cloudy cells could be the same mechanism as the locking of cells fonsp mentioned in https://github.com/fonsp/Pluto.jl/issues/116#issuecomment-631121124

fonsp:
I love the idea of showing the _weight_ of a cell visually!

Cells could also show a row of dots in the corner, each corresponding to a dependent cell. The radius/opacity of a dot can be the corresponding runtime (in log scale), the colour can be the corresponding section colour.

fonsp:
(No idea is too childish!)

karlwessel:
The dots could be interpreted as a chain of different sized weights that the cell is dragging along. Each weight could be clickable to jump to the according depending cell.

j-fu:
[...]

PR #891, opened 2021-02-03
Visualize cell dependencies (Backend)

lungben:
#13 
I started as a very first step with some helper functions on Julia side to identify the cell dependencies. 
Next step is to pass this information to the frontend and show it in the Pluto GUI.


lungben:
@pankgeorg I think I got most of the Julia part - the new data is now in the Cell struct and should also be filled. 
I am now trying to understand how this information is passed to the frontend, but my complete lack of JS is not really helpful here ;-)
Could you please give me a hint?

The idea would be to add the current cell number and the input / output cell numbers next to the runtime information below each cell.
Example - this cell is stored as the 5th in the notebook.jl file, depends on cells 2 and 4 and is itself prerequisite for cells 7, 15 and 22:
# 5 ➡ 2, 4 ⬅ 7, 15, 22

pankgeorg:
> Could you please give me a hint?

YES!!!!
Give some time - I'm kind of more-than-usual busy these days but will prepare something really soon!!

src/webserver/Dynamic.jl
```diff
@@ -106,6 +106,9 @@ function notebook_to_js(notebook::Notebook)
                 "queued" => cell.queued,
                 "running" => cell.running,
                 "errored" => cell.errored,
```
pankgeorg, line 108: `notebook_to_js` function takes our cell variables and throws it to the other side

src/webserver/Firebasey.jl
```diff
@@ -285,17 +285,17 @@ function diff(o1, o2)
 end
 
 # ╔═╡ b8c58aa4-c24d-48a3-b2a8-7c01d50a3349
-function diff(o1::Nothing, o2)
+function diff(o1::Union{Nothing, Missing}, o2)
```
pankgeorg, line 288: Since `cell_execution_order`, `referenced_cells`, `dependent_cells`, but also `runtime`(!) start as _missing_, we need diff to handle that.
Otherwise it's `TypeError: non-boolean (Missing) used in boolean context`.

frontend/components/Editor.js
```diff
@@ -94,7 +94,10 @@ const Main = ({ children }) => {
  *  queued: boolean,
  *  running: boolean,
  *  errored: boolean,
```
pankgeorg, line 96: These are just type declarations, so our environment knows better

pankgeorg:
@lungben I've added what is needed to push the data in the front-end!
On the javascript part:
Now the notebook state includes the new fields!!
you can inspect that if you go to the console <F12> and try to type

`editor_state.notebook.cell_results[<tab>].cell_execution_order`

Here https://github.com/lungben/Pluto.jl/blob/cell_dep_visualizer/frontend/components/Notebook.js#L127 we need to add the extra variables that we just created.
These variables get into what is called `useMemo` in the React ecosystem: it uses the variables that are given as inputs to memoize the cell output. [_Side Note_ To make sure that the memo won't be spoiled when only the order changes, can we plug in 'cell_id' instead of 'order of cell'?]

Then, after we pass the variables we added above, here https://github.com/lungben/Pluto.jl/blob/cell_dep_visualizer/frontend/components/Notebook.js#L6 CellMemo must take the new variables, do the memo stuff and pass them to the Cell component https://github.com/lungben/Pluto.jl/blob/cell_dep_visualizer/frontend/components/Notebook.js#L26

The Cell component must add the new variables here https://github.com/lungben/Pluto.jl/blob/cell_dep_visualizer/frontend/components/Cell.js#L24 and then you can use them!!

The Cell markup starts here https://github.com/lungben/Pluto.jl/blob/cell_dep_visualizer/frontend/components/Cell.js#L76
You may want to add this close to the RunArea (the > button, and the runtime), here https://github.com/lungben/Pluto.jl/blob/cell_dep_visualizer/frontend/components/Cell.js#L152

This is the hint version but if you prefer not to play with JS we can split it - I can do the front-end part - and discuss the exact visuals later and you can try to push the `cell_id` information in dependency arrays so we avoid re-renders!

pankgeorg:
I also messed something with the tests in old julia versions :(  (1.5, 1.6.0-rc work)

lungben:
Thank you very much! 
I did not have time to check it yet, hopefully this evening.

src/webserver/Firebasey.jl
lungben: Good finding! Maybe it makes sense to correct this in a separate PR?

lungben:
> I also messed something with the tests in old julia versions :( (1.5, 1.6.0-rc work)

No idea what the root cause of this issue is (something with missings), but I got it working with a dirty try/catch. As it only affects legacy Julia versions, it is probably OK.

src/webserver/Firebasey.jl
pankgeorg: Yes - Let me write something up so Fons and Michiel can also share thoughts!

fonsp (commented):
This is great, thanks for taking the initiative!

Some confessions:
- The current structure with `Notebook`, `NotebookTopology` and `TopologicalOrder` (which is unnecessarily [recomputed every time you save the notebook](https://github.com/fonsp/Pluto.jl/blob/4c592952f16dedb21c233d6a3cae22f29fac75bf/src/notebook/Notebook.jl#L105-L113)) is a bit chaotic, and not 100% ready for this feature
- I already had some ideas about how to do this, and I should have written that down


Two overall comments:

## 💚 What information to send?
Right now we are sending:
1. Cell _execution_ order
2. Non-recursive upstream (aka referenced aka dependencies) and downstream (aka dependent) cells for each cell -- i.e. the directed graph of **cells**

Other possible things to send:
3. The **variables** defined and referenced by each cell.
4. The **recursive** upstream and downstream cells for each cell. So not just the cells that are directly before or after a cell, but everything before and after.
5. [The extra order heuristic used to help our analysis](https://github.com/fonsp/Pluto.jl/blob/4c592952f16dedb21c233d6a3cae22f29fac75bf/src/analysis/Topology.jl#L116-L133)

### Which do we need?

We want to (have the possibility to) visualize **all four of them**, so all of this information should be available to the frontend.

However, there is redundant information here:

- 3 and 5 **need to be computed by Julia**, since they come from syntax analysis.
- 2 follows from 3, simple algorithm [[downstream]](https://github.com/fonsp/Pluto.jl/blob/4c592952f16dedb21c233d6a3cae22f29fac75bf/src/analysis/Topology.jl#L70-L72), [[upstream]](https://github.com/fonsp/Pluto.jl/blob/4c592952f16dedb21c233d6a3cae22f29fac75bf/src/analysis/Topology.jl#L92-L98)
- 4 follows from 2, simple algorithm [[downstream]](https://github.com/JuliaPluto/PlutoBindServer.jl/blob/01836ca4e9fc332e97f6095796c2b0f3baccc154/src/MoreAnalysis.jl#L33-L50), [[upstream]](https://github.com/JuliaPluto/PlutoBindServer.jl/blob/main/src/MoreAnalysis.jl#L55-L72)
- 1 follows from 3 and 5, tricky algorithm that we don't want to rewrite to JS [[main part]](https://github.com/fonsp/Pluto.jl/blob/master/src/analysis/Topology.jl#L11-L57). It's tricky because it resolves ambiguities, and because of #538

i.e. 1 and 3 and 5, are enough to reconstruct 12345, together with some simple JS code.

So maybe we should only send 1 3 5? Two arguments are:
- it keeps the julia backend slightly less complicated
- the frontend won't always need the 2 and 4 info of **every cell**, you probably only want it for a single cell at once. So the computational burden is not that high
- ideally, we would not just show _"cell A depends on cell B"_, but _"cell A depends on cell B because A defines ... and B ..."_. This means that the frontend needs to know which cell dependencies follow from which variables, i.e. it actually needs to do the 2 -> 3 algorithm.

## 💙 What to cache?

Right now, only the result of 3 is cached (in the form of `notebook.topology`). Like I describe [here](https://github.com/fonsp/Pluto.jl/blob/4c592952f16dedb21c233d6a3cae22f29fac75bf/src/notebook/Notebook.jl#L105-L113), it would be nice to also cache 1. 5 is super fast so does not need caching.

(Adding the properties `referenced_cells` and `dependent_cells` to the `Cell` struct means caching 2.)

src/webserver/Firebasey.jl
fonsp: Why did that not give errors before?

lungben:
Hi @fonsp ! 
Thanks for your suggestions! I agree that it would be very helpful to have also 3. in the frontend. I am not sure how to display it nicely in the GUI, but this is also something we can think about later.
Regarding 2. - I suggest to keep it in for now (just change to UUID), we can replace it later with JS code (this way we can still use it for testing).

src/webserver/Firebasey.jl
lungben: In Dynamic.jl, runtime is converted to nothing if missing:
`                "runtime" => ismissing(cell.runtime) ? nothing : cell.runtime,
`
Maybe this is the reason?

src/notebook/Cell.jl
```diff
@@ -27,6 +27,11 @@ Base.@kwdef mutable struct Cell
     module_usings::Set{Expr}=Set{Expr}()
     rootassignee::Union{Nothing,Symbol}=nothing
     function_wrapped::Bool=false
+
+    # information to display cell dependencies
+    cell_execution_order:: Union{Missing, Int} = missing
+    referenced_cells:: Union{Missing, Vector{Int}} = missing
+    dependent_cells:: Union{Missing, Vector{Int}} = missing
```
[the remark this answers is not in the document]
lungben: adjusted

src/analysis/CellDependencyVisualization.jl
```diff
@@ -0,0 +1,39 @@
...
+
+"""
+Gets a list of all cells which are dependent on the current cell.
+Changes in the current cell cause re-evaluation of these cells.
+"""
+function get_dependent_cells(cell:: Cell, notebook:: Notebook):: Vector{Cell}
+    node = notebook.topology.nodes[cell]
+    return Pluto.where_assigned(notebook, notebook.topology, node.references)
+end
+get_dependent_cells(uuid:: UUID, notebook:: Notebook):: Vector{Cell} = get_dependent_cells(notebook.cells_dict[uuid], notebook)
```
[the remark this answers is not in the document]
lungben: done

src/notebook/Cell.jl
```diff
@@ -27,6 +27,11 @@ Base.@kwdef mutable struct Cell
     module_usings::Set{Expr}=Set{Expr}()
     rootassignee::Union{Nothing,Symbol}=nothing
     function_wrapped::Bool=false
+
+    # information to display cell dependencies
+    cell_execution_order:: Union{Missing, Int} = missing
```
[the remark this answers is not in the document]
lungben: Thanks, I missed this in my tests.
Is corrected now.

lungben:
I corrected the cell order and changed the dependencies to UUIDs.
Next step is to include the defined and referenced variables.

lungben:
The cell struct now contains dictionaries mapping the global variables referenced / defined in each cell to the corresponding cell uuids:
```
referenced_cells:: Union{Missing, Dict{Symbol, Vector{UUID}}}
dependent_cells:: Union{Missing, Dict{Symbol, Vector{UUID}}}
```
This covers points 2+3 and already contains their linkage, removing the need to reconstruct this information in the frontend.

Next steps:
* add 5.
* maybe add ordered cells to notebook level structure?
* check performance impact

src/webserver/Dynamic.jl
```diff
@@ -377,6 +381,7 @@ responses[:reshow_cell] = function response_reshow_cell(🙋::ClientRequest)
     end
     run = WorkspaceManager.format_fetch_in_workspace((🙋.session, 🙋.notebook), cell.cell_id, ends_with_semicolon(cell.code), (parse(PlutoRunner.ObjectID, 🙋.body["objectid"], base=16), convert(Int64, 🙋.body["dim"])))
     set_output!(cell, run)
+    set_dependencies!(cell, 🙋.notebook)
```
lungben, line 384: I am not sure if this is needed here.
Is it possible that the cell value actually changes in `response_reshow_cell`? If not, this could be removed for performance reasons.

lungben:
The execution order of the whole notebook is now accessible in the frontend as `notebook.execution_order` (vector of UUIDs). 
I kept the execution order as integer for each cell in for now, then it does not need to be calculated again in the frontend if it is required.

src/notebook/Notebook.jl
```diff
@@ -26,8 +26,7 @@ Base.@kwdef mutable struct Notebook
     cells_dict::Dict{UUID,Cell}
     cell_order::Array{UUID,1}
     
-    # i still don't really know what an AbstractString is but it makes this package look more professional
-    path::AbstractString
+    path::String
```
lungben, line 29: A bit OT, but I suggest to change the type here to String for type stability.

src/evaluation/Run.jl
```diff
@@ -80,6 +80,9 @@ function run_reactive!(session::ServerSession, notebook::Notebook, old_topology:
 	delete!.([notebook.bonds], to_delete_vars)
 
 	local any_interrupted = false
+	ordered_cells = get_ordered_cells(new_order)
+	set_dependencies!(notebook, ordered_cells)
```
fonsp, line 84: The order of _any_ cell can change when the reactivity graph changes, not just that of cell that will rerun. For example: in a notebook with ten cells, each with code `123`, the cells will execute in oder 1, 2, ..., 10. But if you delete the first cell, the last cell changes its order to 9. The last cell is not a depender of the first cell, so it does not appear in `to_run`.

See my next comment :))))))))))

fonsp (changes requested):
I don't think that `run_reactive!` is the right place to update the reactivity info. For example, `run_reactive!` runs every time you move a slider, where no graph changes can happen, and we want this to be super fast. `run_reactive!` runs (mutliple times) after the topology is determined, we want to store the reactivity info whenever the topology is recalculated.

A better place could be (before, after, or during) `updated_topology`. One option is to store/cache the cell execution order as an extra field in the `NotebookTopology` struct. Then the `save_notebook` function can also use the same cache.

Let me know if this makes sense! I would be happy to talk more about it on a call!

src/webserver/Dynamic.jl
fonsp: Not needed 🌈

fonsp:
Here is the big notebook that I promised:

https://github.com/fonsp/disorganised-mess/blob/master/big.jl

To run a notebook from your REPL:
```julia
using Pluto
path = download(notebook_url)

session = Pluto.ServerSession()
nb = Pluto.SessionActions.open(session, path; run_async=false) # run async=false means that it will run the entire notebook and return when its done

@assert nb isa Pluto.Notebook
```

src/analysis/CellDependencyVisualization.jl
```diff
@@ -0,0 +1,39 @@
+"""
+Gets the cell number in execution order (as saved in the notebook.jl file)
+"""
+function get_cell_number(uuid:: UUID, notebook:: Notebook):: Int
+    cell_id = findfirst(==(uuid), notebook.cell_order)
+    return cell_id === nothing ? -1 : cell_id
+end
```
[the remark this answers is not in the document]
lungben: done

src/notebook/Cell.jl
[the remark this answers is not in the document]
lungben: cell order added as notebook property.

src/notebook/Cell.jl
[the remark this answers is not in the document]
lungben: I added cell order (as a list of all cell UUIDs) as notebook property, but kept the order of each cell (as integer) in as cell property for now. If this causes issues, it could be removed later.

src/webserver/Dynamic.jl
lungben: Thanks, is removed.

src/evaluation/Run.jl
lungben: In `updated_topology`, all information is available to calculate / set the depenencies (the TopologicalOrder needs to be recalculated).
The cell order (as `Vector{UUID}`) is now stored in `notebook`, this information could be re-used when saving the notebook.
In general I think I need better ways to test the dependencies inside a notebook - maybe some JS inside a Pluto cell could do this.

lungben:
For a simple (and very ugly) way to display the dependencies in Pluto, paste the following code into a cell:

```
html"""
<h1>Cell Dependency Data</h1>
<p id="dependencies"></p>
<script>
var i;
var text = "";
var uuid;
var cell;
var references;
var dependents;
for (i = 0; i < editor_state.notebook.cell_execution_order.length; i++) {
	uuid = editor_state.notebook.cell_execution_order[i];
    cell =  editor_state.notebook.cell_results[uuid];
    references = cell.referenced_cells;
	text += "<h3>" + uuid + "</h3>";
	text += "<h4>Referenced cells</h4>";
    if (references) {
		Object.keys(references).forEach(function(key) {
			text += key + " : " + references[key] + "<br>";
	})};    

    dependents = cell.dependent_cells;
	text += "<h4>Dependent cells</h4>";
	if (dependents) {
		Object.keys(dependents).forEach(function(key) {
			text += key + " : " + dependents[key]+ "<br>";
	})}; 
  
}
document.getElementById("dependencies").innerHTML = text;
</script>"""
```

src/evaluation/Run.jl
lungben: I did the changes and it seems to work for me locally.
But unfortunately the Frontend tests fail, I do not know why. Do you have any idea?

fonsp:
Really cool that the reactivity info is accessible to user-written JS!

src/evaluation/Run.jl
fonsp: @pankgeorg will know

lungben:
Yes! Probably people will make a far better use of it than myself - this was the very first JS code I ever wrote ;-)

Is there any way to pass information from JS to Julia inside a Pluto cell (vice versa is trivial)?

src/analysis/Errors.jl
```diff
@@ -44,5 +44,8 @@ function relay_reactivity_error!(cell::Cell, error::Exception)
 	cell.persist_js_state = false
 	cell.errored = true
 	cell.runtime = missing
+	cell.cell_execution_order = missing
```
lungben, line 47: remove it because it is redundant to `notebook.cell_execution_order`

lungben: done

pankgeorg:
I'm investigating (And soon, fixing) the front-end test failing issue.
It is a problem with the state synchronization - the patches created are exactly not the ones the front-end expects. We've had this issue before, and fixed it, so I'm optimistic :)

pankgeorg:
Added a commit to check if the tests pass to make sure I've found the perpetrator! 
I'm almost there!!

Notes:
It seems to be an issue with deep diffing in the notebook_to_js and/or Firebasey. Also investigating if immutability will help.

pankgeorg:
Ok that was a weird bug. Firebasey doesn't like mutable structs so we create a copy of the dependency dicts. Without the copy the diffs are wrong and the patches are inconsistent with each other, so boom 🎊🎊🎊.

An idea for the future is to recover from errors like this: if patches fail to get applied, a request for the whole state should be invoked (making sure to log 🪵 these state resets to find bugs like this(!))

Also I was wondering if we can replace the missing returning an empty Dict. I tried to change the type to drop the union with missing but the React.jl test breaks.

lungben:
Thanks!
The mutability of Dicts is not required on Julia side, they are constructed in simple comprehensions and returned, e.g.

```
return Dict(sym => get_cell_uuids(where_assigned(notebook, notebook.topology, Set((sym,)) ))
        for sym in dependent_symbols)
```

We could replace the Dicts by NamedTuples to avoid mutability issues. 

Not sure about the `missing` - I think it is better to keep them to be able to distinguish between no dependencies (empty Dict / NamedTuple) and dependencies have not yet been calculated (missing), at least on Julia side.

pankgeorg:
> Not sure about the `missing` - I think it is better to keep them to be able to distinguish between no dependencies (empty Dict / NamedTuple) and dependencies have not yet been calculated (missing), at least on Julia side.

Oh I see; yes that makes sense!

Ok I guess the julia side is perfect, front-end seems stable; we've very close to good to go!

fonsp:
I refactored the code to better fit inside the current structure. Commit here: [`da678be` (#891)](https://github.com/fonsp/Pluto.jl/pull/891/commits/da678bef763600e5438e856b7d9ac5feb22b65df)

I am not sure about the [backend test](https://github.com/lungben/Pluto.jl/blob/5d76ed652c70deef98f6dbd6c941b731ce780f04/test/DependencyCache.jl), and I am thinking that we should delete it altogether, because:
- The test is _primarily_ testing Pluto's syntax analysis, reactivity and ambiguity resolution. We already have tests for this
- They include hardcoded tests for execution order in an ambiguous notebook (`@test get_cell_number`). I am sure that this was unintentional, but testing Pluto's ambiguity resolution is tricky. We don't want to test that a cell runs precisely as 4th cell, we want to test whether it runs after ... and before .... If there are new tests for specific execution order, then that increases the future maintenance overhead: every time we change heuristics, we need to analyze this notebook. If you are interested, these tests are [here](https://github.com/fonsp/Pluto.jl/blob/e95d6dc729cd0e8c9224320c62d33916d3c65709/test/React.jl#L142-L224).
- These tests only test implementation details, like variable names. 

I don't think that tests are very necessary for this PR, but if we do, then:
- Tests that topology caches are _updated_ when code changes
- To get closer to an end-to-end test, and less an implementation test, you could call `notebook_to_js` on a notebook to get the same object that the frontend would get, and write tests for its fields.

fonsp:
I still want to make some changes before we merge:
- Delete tests
- The "reactivity cache" objects should not be part of the `Notebook` struct, but the immutable  `NotebookTopology`. Right now there are too many mutable caches in different places
- Maybe move some computations to the frontend side.

fonsp:
You merged the execution barrier code into this branch -- was that an accident? Can you undo?

lungben:
> @lungben You merged the execution barrier code into this branch -- I am assuming that this was an accident and undoing

Yes, thanks! I have just seen this morning and wanted to un-do it myself now ;-)

src/notebook/Cell.jl
```diff
@@ -23,11 +23,20 @@ Base.@kwdef mutable struct Cell
     last_run_timestamp::Float64=0
     "Whether `this` inside `<script id=something>` should refer to the previously returned object in HTML output. This is used for fancy animations. true iff a cell runs as a reactive consequence."
     persist_js_state::Bool=false
+
+    # information to display cell dependencies
+    # Open: move to another place?
+    downstream_cells_map::Dict{Symbol,Vector{Cell}}=Dict{Symbol,Vector{Cell}}()
+    upstream_cells_map::Dict{Symbol,Vector{Cell}}=Dict{Symbol,Vector{Cell}}()
+    precedence_heuristic::Int=99
+
```
lungben, line 32: @fonsp is it fine to keep this information in the Cell struct or should it be moved to somewhere else (e.g. to the new `ExprAnalysisCache`)?

lungben: I put these information into the immutable struct `cell_dependencies`, resembling the structure in JS.

fonsp (approved):

PR #912, opened 2021-02-12
CSS animation performance

fonsp:
# Problem statement
When you move a slider in a graph, a significant number of repaints happen to the _whole page_, according to Google Chrome devtools

> The expectation would be for only the dependent cell _outputs_ to repaint.

# What we know happens
The [animation](https://github.com/fonsp/Pluto.jl/blob/master/frontend/editor.css#L1025) that controls pluto-trafficlight (the 'queued/running/errored-running' indicator) is causing the page repaints.

We know that because removing the animation stops the repaints.

# Why we think that happens
This animation is _supposed to_ run fast: Modern browsers state that changes in `transform: translate` properties should only trigger _composite layers_ recalculation [(a source)](https://www.html5rocks.com/en/tutorials/speed/high-performance-animations/). Though, little is being documented about how hard it is to **_start/stop_** an animation, which is exactly what happens when you move a slider many times:

1. bind sends new data
2. julia replies near instantly that some cells are '`queued`'
3. julia updates state to 'some cells are '`running`'
4. julia updates state to `ready` 

repeat X times in a second (one for each slider value update). That is a simple animation, but is also 50 times to start/stop it. _Probably_ that is slow.

---

# What [this](https://github.com/fonsp/Pluto.jl/pull/912/commits/c2e28938a00be56e3aa5f52f6ddea1984cbdc242) commit does

We make _all cells_ **contantly** run the animation **all the time**, and **hide it** with `opacity: 0`
When a cell changes from `ready` -> `running` we change the opacity with `opacity: 0.97`, so the animation becomes _visible_ (even though it was there the whole time!)

## Next steps

- [x] Identify if that is actually the cause of the repaints
- [x] Be smart about animations: a long forgotten cell should not have an animation running perpetually, _butalso_ a slider-controlled element should not restart animation but smarlty 're-use' the same.
- [x] Compare the idle CPU with the same big notebook without any. Highly likely browsers optimize that (even though that should be what they do in the first place)

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -910,11 +910,13 @@ pluto-trafficlight {
 pluto-trafficlight::after {
     content: "";
     /* Calc needs units everywhere to work */
-    top: calc(0px - 2 * var(--patternHeight));
+    top: calc(0px - 10 * var(--patternHeight));
     left: 0;
     width: 100%;
-    height: calc(100% + 2 * var(--patternHeight));
+    height: calc(100% + 10 * var(--patternHeight));
     position: absolute;
+    opacity: 0;
+    animation: 10s linear 100ms infinite running scrollbackground;
 }
 
 /* in ascending order of severity: */
@@ -959,14 +961,14 @@ pluto-cell.errored > pluto-trafficlight {
 pluto-cell.queued > pluto-trafficlight::after {
     background: repeating-linear-gradient(-45deg, rgba(0, 0, 0, 0), rgba(0, 0, 0, 0) 8px, rgba(0, 0, 0, 0.1) 8px, rgba(0, 0, 0, 0.1) 16px);
     background-clip: content-box;
-    animation: 1s linear 0s infinite running scrollbackground;
+    opacity: 0.97;
     background-size: 4px var(--patternHeight);
 }
 
 pluto-cell.running > pluto-trafficlight::after {
     background: repeating-linear-gradient(-45deg, rgba(0, 0, 0, 0.1), rgba(0, 0, 0, 0.1) 8px, rgba(0, 0, 0, 0.2) 8px, rgba(0, 0, 0, 0.2) 16px);
     background-clip: content-box;
-    animation: 1s linear 0s infinite running scrollbackground;
+    opacity: 0.97;
     background-size: 4px var(--patternHeight);
 }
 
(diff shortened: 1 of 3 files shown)


fonsp:
Can you measure and post the performance gain?

pankgeorg:
Trying to find the performance, I figured (this)[https://github.com/fonsp/Pluto.jl/pull/912/commits/a03f50d15eb3738c88f7977451aa0f3998253bd9) out, which seems to avoid some renders (rerun the preact diff for the whole Cell component (_after_ the CellMemo) every 50ms for every running cell)

About this https://github.com/fonsp/Pluto.jl/pull/912/commits/239002962de1c33ccd7c2a1bdbc8a4d2863c9b13, It seems it makes a difference when scrolling, but not so much when idle. These measurements are not very precise, of course, I'm trying to figure out if there is way to actually measure this efficiently.

## Disable animation
6 seconds when idle:
image
(pretty much the same)

10 seconds while scrolling:
image
(better)

## Leave animation on
6 seconds when idle:
image
(pretty much the same)

10 seconds while scrolling: 
image
(worse)

fonsp:
@pankgeorg with "scrolling" you mean moving the slider?

frontend/editor.css
```diff
@@ -913,11 +913,17 @@ pluto-trafficlight {
...
     width: 100%;
-    height: calc(100% + 2 * var(--patternHeight));
+    height: calc(100% + 10 * var(--patternHeight));
     position: absolute;
+    opacity: 0;
+}
+
+/* This class will toggle animation. Other classes will make animation visible with opacity */
+pluto-cell.activate_animation pluto-trafficlight::after {
+    animation: 10s linear 100ms infinite running scrollbackground;
```
fonsp, line 926: 100ms?

frontend/editor.css
```diff
@@ -962,14 +968,14 @@ pluto-cell.errored > pluto-trafficlight {
 pluto-cell.queued > pluto-trafficlight::after {
     background: repeating-linear-gradient(-45deg, rgba(0, 0, 0, 0), rgba(0, 0, 0, 0) 8px, rgba(0, 0, 0, 0.1) 8px, rgba(0, 0, 0, 0.1) 16px);
     background-clip: content-box;
-    animation: 1s linear 0s infinite running scrollbackground;
+    opacity: 0.97;
```
fonsp, line 971: .97?

frontend/editor.css
pankgeorg: Will switch to 0. That made some sense when animation-triggering state updates were not batched behind `activate_animation`.

frontend/editor.css
pankgeorg: Here I'm being extra careful to make sure we're creating an extra GPU layer (needs opacity < 1).

fonsp:
ah that’s cool! just write that in a comment i guess

Issue #893, opened 2021-02-04
Built-in array viewer for `Set`


fonsp:
For `AbstractArray`, `AbstractDict`, tuples, and more, we use Pluto's built-in interactive tree viewer:

![image](https://user-images.githubusercontent.com/6933510/106891759-7323b600-66eb-11eb-80d4-b3a5db9cc44f.png)

We should also use it for `Set`:

![image](https://user-images.githubusercontent.com/6933510/106891959-b54cf780-66eb-11eb-96ea-da63f62a5ac5.png)

vladflore:
Hey @fonsp ,

I'd like to have a got at this (I've already worked a bit on it and managed to get some results) :).

P.S. beginner here, Julia, Pluto and OS contributing so please bear with me :)

Best,
Vlad

PR #926, opened 2021-02-18
Use Pluto tree viewer for Set data type (issue 893)

vladflore:
Hey @fonsp ,

When you get to this please have a look and let me know if this PR is on the right track or is already a total mess :).

This PR relates to issue #893 

Thanks,
Vlad


fonsp:
Hey @vladflore, the code so far looks good! Can you post a screenshot?

vladflore:
Yes, I can do that, will do it a bit later this day though :)

vladflore:
Here is a `gif` of how it looks like:

pluto-tree-viewer-set

Let me know if you need anything else.

Best,
Vlad

fonsp:
Thanks! It looks like you made all the necessary changes to the backend (Julia), but we still need to change the appearance, so that people can tell that the object is a set.

Of course, our main inspiration is the julia REPL, where sets are displayed like:
```julia
julia> Set([1,2,3]) |> print      # 'non-verbose' mode
Set([2, 3, 1])

julia> Set([1,2,3])               # 'verbose' mode
Set{Int64} with 3 elements:
  2
  3
  1
```

I think that our collapsed form should look exactly like the non-verbose mode, i.e. `Set` or `Set{Int64}` as prefix, `([` and `])` as begin and end, and `,` as delimiter.

Secondly, when expanded, we should not show the indices (`1:`, `2:`, etc) because `Set`s are unordered objects.

---

@vladflore To figure out how to do this, try to find out how Pluto displays the prefix, begin, end, delimiter and indices for arrays and dictionaries. Take a look at `treeview.css`, and use the developer tools of your browser to analyze and write CSS. https://developers.google.com/web/tools/chrome-devtools/css You should be able to do this without fully understanding CSS&JS 🌝

fonsp:
Also, can you test this on a `Set` with 100 elements? e.g. `Set(1:100)`

vladflore:
Hey @fonsp ,

I made the changes you mentioned. Thanks for the tips and hints! Please have a look when you have the time and let me know if I need to change something else.

Here is a gif of the current status:

pluto-tree-viewer-set

Best,
Vlad

fonsp:
Really nice, thanks @vladflore ! What do you want to work on next?

fonsp:
Released in Pluto 0.14!

PR #947, opened 2021-02-24
load Revise first

ericphanson:
Adds a heuristic step to load Revise before other packages so it can revise them properly. It only works with `using Revise` or `using X, Y, Revise, Z` etc, but not with `import` since it relies on `compute_usings` which only looks for `using` expressions, not `import` expressions.


fonsp:
Thanks!

PR #953, opened 2021-02-26
Move to backtick latex delimiting

jlewis91:
Per the very helpful tip in [this discourse thread](https://discourse.julialang.org/t/pluto-markdown-not-working/53433), it seems like there might be a lot to be gained and little to be lost by encouraging users to use backticks instead of dollar signs for latex snippets, so that dollar signs are reserved for (and more reliably parsed) in the case of `$(1+1)` style interpolation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "Pluto"
 uuid = "c3e4b0f8-55cb-11ea-2926-15256bba5781"
 license = "MIT"
 authors = ["Fons van der Plas <fonsvdplas@gmail.com>"]
-version = "0.12.21"
+version = "0.14.0"
 
 [deps]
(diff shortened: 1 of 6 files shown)


fonsp:
Thanks! Did you check all sample notebooks?

jlewis91:
@fonsp, yes, via regex, so I’m 98% sure I got everything. :)

jlewis91:
@fonsp Can now in good faith say that I've reviewed all of the sample notebooks. Some of the diff updates are because I reran the notebooks in 0.14; let me know if you'd prefer I remove these commits.

fonsp:
Yeah can you try to get smaller diffs?

jlewis91:
@fonsp Ok, think we have a clean diff now. :)

PR #956, opened 2021-02-28
Demo Using CommonMark in Pluto.jl

jlewis91:


sample/Getting started.jl
```diff
@@ -13,23 +13,42 @@ macro bind(def, element)
...
+	
+	parser = Parser()
+	enable!(parser, MathRule())
+	enable!(parser, TableRule())
+	
+	function cm_md(str_ex)
+		ast = parser(str_ex)
+		html_ex = CommonMark.html(ast)
+		HTML(html_ex)
+	end
```
MichaelHatherly, line 31: You _should_ be able to just do something like

```julia
begin
    using CommonMark

    macro cm_str(str)
        return Parser()(str)
    end
end
```

since the AST object that `CommonMark` returns implements `show` for `MIME"text/html`, unless there's something Pluto-specific that requires something else?

jlewis91: Thanks! Between this tip and looking at the existing markdown lib code, I was able to cobble something together that works with math syntax (e.g. properly escaped LaTeX) as well. :)

fonsp:
Hey! Thanks for helping with our markdown samples!

For this PR, I think that the complexity of installing and setting up commonmark does not belong in the "getting started" notebook. This notebook should not use any packages, and the Markdown formatting issues are very niche compared to the challenge of learning Julia and Pluto.

We could make a separate sample notebook about CommonMark, but the reader might wonder: "If this is such an essential package, then why does Pluto come with Markdown pre-imported, but not with Commonmark?" **Which is a great question!** What is the status of replacing Markdown with CommonMark in the julia stdlib?

MichaelHatherly:
> What is the status of replacing Markdown with CommonMark in the julia stdlib?

In simple terms, it requires some time on my part (or someone willing to learn the CM.jl and MD.jl codebases) to make sure we have feature parity with MD.jl and then execute a transition phase away from it to CM.jl, which will be time consuming since their underlying datastructures are not similar and most heavy users of MD.jl are using it's internals.

Timeframe is "when it's ready" 😄

jlewis91:
Marked this PR as a draft, idea is to track how close `CommonMark.jl` is to being a full replacement for `Base.Markdown` in the context of `Pluto.jl`'s use case. As for the headache of 'boilerplate code', my proposal would be (once CM.jl, MD.jl feature parity in context of `Pluto.jl`'s use of the package is reached), to add `CM.jl` as a dependency (naturally, can already see reasons why you might have objections to this).

Importantly, however, while you are right that:
> formatting issues are very niche compared to the challenge of learning Julia and Pluto.
nonetheless, these issues are the difference for many users between a 'just works' experience where LaTeX, interpolation and markdown can be used together harmoniously in notebooks and an experience where there are many mysterious formatting issues. I tried to highlight this by linking to Github Issues already posted on this topic, would be interested to hear whether you can see a trend in the anonymous feedback as to whether this is a recurrent issue. :)

jlewis91:
I think that the proposal in https://github.com/JuliaLang/julia/issues/37337 is quite coherent in putting forward the idea of **different** ideas of 'feature parity', depending on the use case / application. In this context, it would seem like `Pluto.jl` could be satisfied by Markdown feature parity much sooner than a library using `Markdown.jl` object manipulations.

fonsp:
This all sounds great, thanks for your work! I definitely support it from the Pluto side.

Some notes:
- Adding CM as a dependency is difficult right now, because the dependencies of Pluto.jl as a notebook server and the dependencies of your notebook code are currently intertwined in a strange way. (Long story short: notebooks only have inherent dependencies on stdlibs: Markdown and InteractiveUtils, to avoid problems). But this could be a possibility after https://github.com/fonsp/Pluto.jl/pull/844
- https://github.com/fonsp/Pluto.jl/pull/953 is a good initiative, and we could use syntax highlighting to warn users about `$`-latex during the transition
- Pluto does do so special treatment for Markdown, besides importing it by default. Mainly https://github.com/fonsp/Pluto.jl/blob/7b1041b7c76bcd7f6ba0db528501f62535f3ac6b/src/evaluation/Run.jl#L158-L170 and macro expansion to detect interpolated variables: https://github.com/fonsp/Pluto.jl/blob/7b1041b7c76bcd7f6ba0db528501f62535f3ac6b/src/analysis/ExpressionExplorer.jl#L717-L723 . But this does not depend on Markdown's internal structure.

jlewis91:
@fonsp quick update, thanks to @MichaelHatherly CommonMark.jl now has a branch which fully supports a markdown macro including interpolation (but which does not, by default, support dollar math), I've updated this PR to direct import the package in case you're curious

fonsp:
That's awesome! I think that disabling dollar math is a good choice.

I am excited for CommonMark.jl, but I won't be able to follow its development very closely. Right now I think the best way forward is to wait for #844 and maybe reconsider using CommonMark as a default package, but ideally I would like to see it implemented in the stdlib. Tag @fonsp if you need me in the meantime!

DhruvaSambrani:
@jeremiahpslewis , a reninder that #844 is merged and released

jeremiahpslewis:
Thanks!

sample/Getting started.jl
```diff
@@ -13,23 +13,29 @@ macro bind(def, element)
     end
 end
 
+# ╔═╡ 9b62f2c6-8037-486f-b8ec-a95b577b8d74
+# Code targeted for potential inclusion in the Pluto.jl package:
+begin
+	using Pkg
+	Pkg.add(PackageSpec(url="https://github.com/MichaelHatherly/CommonMark.jl",
+	rev="mh/interp"))
```
MichaelHatherly, line 21: Should be available on latest release btw @jeremiahpslewis.

jeremiahpslewis: Fixed, think this is no longer necessary thanks to the package manager!

jeremiahpslewis:
@DhruvaSambrani  @MichaelHatherly think this is now ready for consideration; really blown away by the new package manager in Pluto.jl, feels like magic to use, but still miraculously produces intuitive boilerplate and commits. 🦄

sample/Getting started.jl
```diff
@@ -13,6 +13,9 @@ macro bind(def, element)
     end
 end
 
+# ╔═╡ 9b62f2c6-8037-486f-b8ec-a95b577b8d74
+using CommonMark
+
 # ╔═╡ b129ba7c-953a-11ea-3379-17adae34924c
 md"# _Welcome to Pluto!_
```
MichaelHatherly, line 20: Should these be `cm"` here if this is meant to exercise CM.jl rather than MD.jl? Or have I missed some magic elsewhere?

jeremiahpslewis: You're entirely right, thanks. I've fixed this.

sample/Getting started.jl
```diff
@@ -38,103 +38,109 @@ end
...
 if !@isdefined power
-	md"Uh oh! Pluto doesn't support multiple expressions per cell. This is a conscious choice - this restriction helps you write less buggy code once you get used to it. To fix the code, you can either split the above cell, or wrap it in a `begin ... end` block. Both work."
+	cm"Uh oh! Pluto doesn't support multiple expressions per cell. This is a conscious choice - this restriction helps you write less buggy code once you get used to it. To fix the code, you can either split the above cell, or wrap it in a `begin ... end` block. Both work."
 else
-	md"**Well done!** Your cats have powers now."
+	cm"**Well done!** Your cats have powers now."
 end
 
 # ╔═╡ 5edadcd2-9554-11ea-1714-b5b7692c4797
 html"""<p>We're almost done! It's time to share your amazing story. Scroll to the top of the notebook, and click on <img src="https://cdn.jsdelivr.net/gh/ionic-team/ionicons@5.5.1/src/svg/shapes-outline.svg" style="width: 1em; height: 1em; margin-bottom: -.2em;"> to see the export options - or you can always share this notebook's save file. (The file is pure Julia, by the way, and it's runnable! You'll learn more about this in the advanced introduction.)</p>"""
```
DhruvaSambrani, line 134: Convert to cm

sample/Getting started.jl
```diff
@@ -38,103 +38,109 @@ end
 html"""<p>To edit any code, just click on it. When you're done programming, press the <img src="https://cdn.jsdelivr.net/gh/ionic-team/ionicons@5.5.1/src/svg/caret-forward-circle-outline.svg" style="width: 1em; height: 1em; margin-bottom: -.2em;"> in the lower-right corner of a cell to run the code. You can also use `Shift+Enter` if you are in a hurry.</p>"""
```
DhruvaSambrani, line 38: Convert to cm

sample/Getting started.jl
```diff
@@ -13,23 +13,23 @@ macro bind(def, element)
...
 
-# ╔═╡ 3e8e381e-953f-11ea-3d3e-71d0fea52560
-cat = "Ks*;lj"
-
 # ╔═╡ aeb3a6bc-9540-11ea-0b8f-6d37412bfe68
 if cat == "Ks*;lj"
-	md"Oh no! Someone messed with my pretty introduction. Change the code above to give our cat a proper name!"
+	cm"Oh no! Someone messed with my pretty introduction. Change the code above to give our cat a proper name!"
 else
 	HTML("""<p><b>Well done, your cat is called $cat now.</b> This text gets updated every time you change the name. To see how the magic works, click on the <img src="https://cdn.jsdelivr.net/gh/ionic-team/ionicons@5.5.1/src/svg/eye-outline.svg" style="width: 1em; height: 1em; margin-bottom: -.2em;"> to the left of this text.</p>""")
```
DhruvaSambrani, line 34: Convert to cm

sample/Getting started.jl
```diff
@@ -38,103 +38,109 @@ end
...
 if @isdefined scary_dog
-	md" $scary_dog is terrorizing the neighborhood! We must do something about it!"
+	cm" $scary_dog is terrorizing the neighborhood! We must do something about it!"
 else
-	md"You saved the neighborhood! Referencing `scary_dog` leads to an `UndefVarError`, as if it never even existed."
+	cm"You saved the neighborhood! Referencing `scary_dog` leads to an `UndefVarError`, as if it never even existed."
 end
 
 # ╔═╡ 36cd006a-9551-11ea-3c0c-df8b7f2843c1
 HTML("""<p>To delete a cell like the one defining $scary_dog, click on the <img src="https://cdn.jsdelivr.net/gh/ionic-team/ionicons@5.5.1/src/svg/close-circle-outline.svg" style="width: 1em; height: 1em; margin-bottom: -.2em;"> on the right of its code.</p>""")
```
DhruvaSambrani, line 102: Convert to cm

sample/Getting started.jl
DhruvaSambrani, line 134: Convert to cm

sample/Getting started.jl
jeremiahpslewis: cm isn't intended as an html replacement

jeremiahpslewis:
@DhruvaSambrani think you may have seen an older diff, all the md macros were covered in my commit https://github.com/fonsp/Pluto.jl/pull/956/commits/3046d303d41844bad0dfe479f9aae2ebb82f82ce

DhruvaSambrani:
Yup, there are some HTML macros left though

jeremiahpslewis:
@DhruvaSambrani Not sure whethe rut really makes sense to use cm markdown macros for raw html. What is your reasoning for wanting to replace them?

DhruvaSambrani:
Ah ok,
So my thinking was that the only use of HTML in this notebook is to add attributes to an image. But commonmark allows us to add attributes and I felt that we should illustrate that in this notebook too!

MichaelHatherly:
> to add attributes to an image

`cm"` _should_ support those fine, though it's never going to manage to be 100% successful with _all_ embedded HTML from what I understand of the spec. I think it's probably worth at least trying to use `cm"` where possible, but falling back to `html"` for the more complex cases.

jeremiahpslewis:
@DhruvaSambrani @MichaelHatherly Ok, so I tested replacing the html macro as well as the HTML( function, the former works with cm and has been swapped out, the latter is going to need to stay as-is with the function call.

DhruvaSambrani:
```julia
cm"""**Well done, your cat is called $(cat) now.** This text gets updated every time you change the name. To see how the magic works, click on the ![eye](https://cdn.jsdelivr.net/gh/ionic-team/ionicons@5.5.1/src/svg/eye-outline.svg){ width: 1em, height: 1em, margin-bottom: -.2em }to the left of this text."""
```

jeremiahpslewis:
Syntax is fine, rendering fails to interpolate `$(cat)` to `cat` afaict.

MichaelHatherly:
> Syntax is fine, rendering fails to interpolate $(cat) to cat afaict.

Is the specific to Pluto, or is it also happening in the REPL?

MichaelHatherly:
Is it note just because https://github.com/fonsp/Pluto.jl/pull/1032 is not implemented yet?

jeremiahpslewis:
@MichaelHatherly Tried running it in the repl, can confirm that this issue is reproducible outside of Pluto. See https://github.com/MichaelHatherly/CommonMark.jl/issues/22 which I just created for documentation

jeremiahpslewis:
@MichaelHatherly @DhruvaSambrani after another round of feedback, think we've reached a steady state: because of CommonMark spec (and reasonable idea of separation of concerns), let's keep html snippets in html, otherwise we're going to start encountering the same unpredictable behavior in CM.jl that we are trying to avoid from MD.jl. If you want good, no headaches markdown, then use the cm macro, if you want html, then go for the existing Julia html utilities, they work well with no hassles. I've added a 'doc' snippet to issue https://github.com/MichaelHatherly/CommonMark.jl/issues/22 to explain this, can imagine something similar might belong in doc pages to CM.jl when it's time to staff up that part of the package.

DhruvaSambrani:
While the CommonMark issue is a bug that needs to be fixed, why does it affect our example? The <p/> tags are not necessary, as they should be appended while displaying.

MichaelHatherly:
> While the CommonMark issue is a bug that needs to be fixed, why does it affect our example? The <p/> tags are not necessary, as they should be appended while displaying.

Yes, removing the `p` tags is enough to allow the interpolation in this case since `p` and `b` are in different HTML "types" as defined by the commonmark spec. `p` is a block of HTML whereas `b` is just a single tag.

Unless the commonmark spec changes there'll be no changes to the parsing precedence in `CommonMark.jl` for this case. You're welcome to look into a some kind of fix to make it work, but it needs to fit in with the spec.

jeremiahpslewis:
@DhruvaSambrani Think @MichaelHatherly has a good point here that CommonMark.jl is spec driven, not a bug.

Not sure I clearly expressed my other concern: why push html into markdown when we can just use html, e.g. if not broken, don't fix it; html function allows for full expressiveness of html; less of a question of what we can force to work for the minimal demo script.

fonsp:
Sorry for leaving this PR for so long! I took at look at it again today, and this led to a fun experiment! https://github.com/MichaelHatherly/CommonMark.jl/issues/31

Since we have #844 , it is now much easier to use packages inside sample notebooks, and the plan is to add lots and lots of new sample notebooks during the next months! We should definitely do one that uses CommonMark, either directly, or using the new CommonMark macro if https://github.com/MichaelHatherly/CommonMark.jl/issues/31 turns out to work as good as it seems.

fonsp:
Hello from the future!

https://github.com/MichaelHatherly/CommonMark.jl/issues/31 became https://github.com/JuliaPluto/MarkdownLiteral.jl which is definitely **fun** and it works well, but the problem is `\` for latex. 

We need to choose between:
- `@md """` which has reliable interpolation, but latex needs double backslashes everywhere `\\`
- `md"""` which has buggy interpolation, but latex works with `\`.

<details>
<summary>Example of interpolation</summary>

image

</details>


<details>
<summary>Example of latex interpolation</summary>

image

</details>


Since we used `md"""` in the past, and "LaTeX is more important than interpolation", we should stick with a `md"""` style macro as the default.

fonsp:
@vchuravy gave me the idea that we can give a different macro in Pluto, than in the `.jl` notebook file. This way, we could give a better Markdown experience in Pluto, with a graceful fallback in the veeeeery rare case that people run the `.jl` file standalone and require the `md"""` macro to have the added features.

This means:
- The `.jl` files stay the same, and they will use the default `@md_str` macro from stdlib Markdown
- When running in Pluto, the `@md_str` will be bound to another macro. I'm thinking of CommonMark for this :)

For this change, we need:
- Check how compatible CommonMark can be with stdlib Markdown (it should not break existing notebooks)
- Write documentation for the new "Pluto Markdown flavour"

PR #962, opened 2021-03-02
Keyword args are now explored last in a function definition

### LINKED ISSUE #915: An unexpected cylic reference error when using kwargs
The kwargs search for args first, then the global variables.

image

```julia
julia> x = 3
3

julia> function f(x; y=3x)
       return y
       end
f (generic function with 1 method)

julia> f(9)
27
```

So, the correct expression explore should detect no cyclic reference.

For testing:
```julia
begin
	using LinearAlgebra
	function f(x;
		y=LinearAlgebra.mul!(3, x)
	)
	end
end

x = LinearAlgebra.mul!(3, 4)
```

Pangoraw:
Hello! :raised_hands: 

Another small fix for the ExpressionExplorer :+1: 

According to https://docs.julialang.org/en/v1/devdocs/ast/#Calls
the `:parameter` part of the arguments created using `;` in the argument
list is the second argument of the created `Expr`. If it is present,
it is set to be the last parameter explored, allowing references to
other parameters.

Closes #915

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -627,19 +627,25 @@ end
 Is also used for `struct` and `abstract`."
 function explore_funcdef!(ex::Expr, scopestate::ScopeState)::Tuple{FunctionName,SymbolsState}
     if ex.head == :call
+        params_to_explore = ex.args[2:end]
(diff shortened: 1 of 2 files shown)


fonsp:
Wow I did not know about those devdocs, amazing! We can go through the list and make sure that everything is covered.

fonsp:
image

Issue #965, opened 2021-03-04
Can't paste code


fonsp:
https://user-images.githubusercontent.com/6933510/109964255-bb83c300-7ced-11eb-81d3-2de56aece5df.mov

PR #966, opened 2021-03-04
PLJ-965 can't paste code

pankgeorg:
Addresses #965 

Currently draft - let's merge after I add some puppeteer tests, because as we keep responding to new user events this may break again.


pankgeorg:
Added paste tests / merging!

Issue #275, opened 2020-08-12
Cells with @bind value run twice


fonsp:
From @GiggleLiu in https://github.com/fonsp/Pluto.jl/issues/262:

Thanks for you suggestion @fonsp , there is not error anymore with PlutoUI.

But there is still an issue. The last cell is refreshed multiple times when I update a cell. This is not desired because the last cell is expensive to evaluate in my application.

The cells causing multiple update looks like

### The Controller
```julia
if optimizer == "ADAM"
	md"""
η (learning rate) = $(@bind adam_η Slider(0.0:0.001:1.0; default=0.01))
$(html"<br>")
β1 = $(@bind adam_β1 Slider(0.0:0.001:1.0; default=0.9))

β2 = $(@bind adam_β2 Slider(0.0:0.0001:1.0; default=0.999))

[help on ADAM](https://ruder.io/optimizing-gradient-descent/index.html#adam)
"""
elseif optimizer == "SPSA"
	md"""learning rate = ``c/m^γ`` (``m`` is the step)
$(html"<br>")
c = $(@bind spsa_γ Slider(0:0.001:1.0; default=0.02))

γ = $(@bind spsa_ξ Slider(0.01:0.001:1.0; default=0.101))
$(html"<br>")
gradient search range = ``a/(m+A)^α``
$(html"<br>")
a = $(@bind spsa_δ Slider(0.01:0.001:1.0; default=0.1))

α = $(@bind spsa_α Slider(0.01:0.001:1.0; default=0.602))
$(html"<br>")
A = $(@bind spsa_A Slider(0.0:0.1:50.0; default=5.0))

[help on SPSA](https://pdfs.semanticscholar.org/bf67/0fb6b1bd319938c6a879570fa744cf36b240.pdf)
"""
elseif optimizer == "AdaGrad"
	md"""
η (learning rate) = $(@bind adagrad_η Slider(0.0:0.001:1.0; default=0.01))

[help on AdaGrad](https://ruder.io/optimizing-gradient-descent/index.html#adagrad)
"""
elseif optimizer == "MGD"
	md"""
learning rate = ``c/m^γ`` (``m`` is the step)
$(html"<br>")
c = $(@bind mgd_γ Slider(0.0:0.001:1.0; default=0.02))
$(html"<br>")
γ = $(@bind mgd_ξ Slider(0.01:0.001:1.0; default=0.101))
$(html"<br>")
gradient search range = ``a/(m+A)^α``
$(html"<br>")
[...]

PR #975, opened 2021-03-06
PLJ-928 Bond value fix

pankgeorg:
There is a flag that's currently ignored : `is_first_value`. 

There is also a prophylactic (`προ-φύλαξις` - guard proactively) comment in [this line](https://github.com/fonsp/Pluto.jl/blob/54e6658ab7cb4b6a1c9572e9e4958f1f6f740604/frontend/components/Editor.js#L413) that talks about #275 - it looks that #275 doesn't re-appear.

@fons, @dralletje remember why the comment was for?

This fix does exactly the opposite of the comment, so let's discuss!


dralletje:
It was meant as a reminder to use `is_first_value` again if #275 would be un-fixed because of Firebasey.
Something like #275 seems to be happening here though, so using `is_first_value` is exactly right.

However, with this simple prophylactic we prevent html binds from setting their initial value:
```julia
# ╔═╡ 20bd321a-9cec-4d6a-a618-cc8132fbc8c7
@bind y html"<input type=range>"

# ╔═╡ fbf09000-632d-419a-83b8-c2a739c13cdd
y
```
image

`is_first_value` is now inferred by the server, looking if the client sends an `add` patch or an `update` patch. Which is kinda hacky actually (I mention #275 there again) because it only sends an `add` patch when the bond wasn't set by anyone before (so it will send it on a refresh, because it thinks "x already existed on the bonds state, so it isn't the first value!".
I changed this because sending the `is_first_value` to the server is a bit more tricky. But I changed that back now.

dralletje:
Dangit, I'm too used to Firebasey "mutating" by copying that I totally forgot to test that Bond Value mutation line 😂

pankgeorg:
Hahaha! Hopefully the test will warn if #275 comes back too!

pankgeorg:
Ok the good thing is that the test now passes.
The weird thing is that the value it passes with is 3?! (the value stays stable through refreshes)

dralletje:
Yeah, initially it will run the bonds one-by-one because that how it is implemented on the server, will have a look at that now

dralletje:
I made this waaaaay to big very quickly, so I'll put some smaller parts of this out first 😅

pankgeorg:
> I made this waaaaay to big very quickly, so I'll put some smaller parts of this out first 😅

Let me know what chuncks I can bite off this !

src/webserver/Dynamic.jl
```diff
@@ -168,9 +172,11 @@ A placeholder path. The path elements that it replaced will be given to the func
...
 struct Wildcard end
 
-@enum Changed begin
-    CodeChanged
-    FileChanged
+abstract type Changed end
+struct CodeChanged <: Changed end
+struct FileChanged <: Changed end
+struct BondChanged <: Changed
+    bond_name::Symbol
```
fonsp, line 179: omg this is genius

pankgeorg:
@dralletje I think I fixed the `is_first_value` part - I'm not sure if the part where multiple bonds get updated together works properly.

The test that ensures bond stability through refreshes passes.

Can you take a look? 😇😇

fonsp:
Whoopsa this broke custom inputs like `PlutoUI.Scrubbable`

Issue #298, opened 2020-08-16
Feature idea: a temporary execution barrier


marius311:
Hi guys, over the last few days, after having toyed with Pluto for a while, I spent a solid chunk of time using it for my actual real work. Aside from continuing to love reactivity and wondering how I've gone so long without it, I think I pinpointed one feature which I feel like could take my productivity while using Pluto to the next level, which I thought I'd propose (and apologies if this has already been discussed somewhere). 

For reference, I'm a physicist doing lots of free-form analysis work, probably pretty typical of your users in academia. Basically the thing I kept running into is I have a fairly "mature" notebook set up, and then I want to tweak something near the "beginning" of the DAG, but I might want to look at a few cells beyond that to check things are right, before letting the whole DAG run to the end, which may actually take minutes to tens of minutes in my case.

What I usually ended up doing is to just make the change, look at the "quick check" cells as the whole thing evaluates, and then just interrupt and try again if "things weren't right". Unfortunatey interrupt is not always reliable or even possible if external libraries are called, and somehow the whole thing felt frantic. 

So my idea is basically just to add a way to place a barrier in the DAG either before or after a given cell (I ran into cases where both would be useful). I made a mock-up of what the UI might look like below. The button could live alongside the hoverable add cell above/below button, and once clicked would remain persistent (and can be clicked again to turn off). To visually indicate the temperary break in reactivity, the UI should dim all the subsequent cells in the DAG and hide their output. It would be something like this:

![image](https://user-images.githubusercontent.com/1510968/90331497-b5a0dd00-df69-11ea-8b8a-63e238cfaff3.png)

In terms of what is saved in the notebook file, maybe the break in the DAG can be indicated by just commenting out code which is beyond the barrier, maybe with a `#= Pluto barrier ... =#` or something to avoid confusing with user code. 

I can imagine you might have some reservations about temporarily breaking reactivity, which I can definitely understand. I do think though that this isn't really getting rid of reactivity, instead its an optional "power" feature that kind of fits in the reactivity paradigm really well. If I recall too in your JuliaCon talk you mentioned work on a way to "check in" a bunch of changes at once. Maybe that's similar and you've already thought of this, it sounded like its trying to solve the same underlying problem, but if that's similar to "pausing" evaluation while editing several cells at once, I'll mention that the key thing here is that you get to look at the output of _some_ cells after you've made a change. 

[...]

Issue #259, opened 2020-08-06
Delay running the notebook when it starts


ndgnuh:
Hi, thanks for the super awesome package. This is just some suggestion. Sometime I want to edit something before running all the cell, and there will be a very time-taking cell at the end. But stopping after the cell run is not very responsive and might kill the worker. I know I can edit the `jl` file before running Pluto but it's not very convenient. Also, with delayed run, a `run all (*, above, below)` button would be nice. :)

fonsp:
I wrote a bit about it [here](https://www.notion.so/Notebook-security-d8162867a8c04b1793912d78cefee02a) - it is also a good idea to do this by default for notebooks that were written by someone else.

j-fu:
In a similar spirit I would like to have this option for running cells one-by one after loading when teaching based on a-priori prepared notebooks. 
Another thing nice to have in this context  would be the option to hide all source code of markdown cells at once.
Together this would provide an experience similar to that of Jupyter when loading a fresh notebook.

fonsp:
About automatically hiding markdown: I don't this is necessary when you write your notebooks inside Pluto, but in your case, it can do this as part of the export process (the fold info is stored in the file):
```julia
c = Pluto.Cell("123")
c.folded = true
# save the notebook
```

Running cells one by one used to be the default (before https://github.com/fonsp/Pluto.jl/commit/d2015176af6ef437ad4feafedf1a01dae05426c9), but there were too many problems with this that we could not fix with friendly UI - the issue is that you launch your notebook into a broken state by default - running a cell might error depending on which cells you manually ran before. It had a lot of the same state problems as Jupyter that reactivity should solve. (The goal is not to make 'Jupyter++', but to try something else.)

But if you just want the 'reveal' effect, then you can add a `;` at the end, and delete it and re-run when you want the output to show.

j-fu:
> About automatically hiding markdown: I don't this is necessary when you write your notebooks inside Pluto, but in your case ([#132 (comment)](https://github.com/fonsp/Pluto.jl/issues/132#issuecomment-675764316)), it can be done as part of the export process (the fold info is stored in the file):
> 
> ```julia
> c = Pluto.Cell("123")
> c.folded = true
> # save the notebook
This looks good - it seems that it would be easy to set this for all markdown cells bevore saving, at least as an option.

> 
[...]

Issue #151, opened 2020-06-06
Interrupt child cell without interrupting parent cell


karlwessel:
I have two cells A and B. Cell B depends on A. 

When both cells are running and i stop cell B it always also stops the execution of cell A.

Would it be possible to only stop cell B so basically remove it from the execution queue of the worker without interrupting the whole worker?

fonsp:
Nice! It should also continue running cells that are not dependent on any interrupted cells.

fonsp:
This should be the default for the stop button. `Ctrl+Q` should stop the entire notebook.

fonsp:
If anyone wants to work on this: you might experience problems with interrupt during testing - _please report those_!

karlwessel:
Here is a simple and I think minimal notebook that is not interruptable:

```
### A Pluto.jl notebook ###
# v0.12.4

using Markdown
using InteractiveUtils

# ╔═╡ fe4609b2-0ebd-11eb-26f4-f90c3a390d1e
begin
	using Pkg
	Pkg.add("Images")
end

# ╔═╡ 07dc02e0-0eb1-11eb-281e-b368f38ddf48
begin
	Base.show(io::IO, a::Int) = print(io, 6)
	5
end

# ╔═╡ Cell order:
# ╠═fe4609b2-0ebd-11eb-26f4-f90c3a390d1e
# ╠═07dc02e0-0eb1-11eb-281e-b368f38ddf48
```

The execution of the cell seems to actually already have stopped, but the notebook doesn't seem to notice it:

```
On worker 2:
StackOverflowError:
Stacktrace:
 [1] remotecall_fetch(::Function, ::Distributed.Worker, ::Module, ::Vararg{Any,N} where N; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /build/julia/src/julia-1.5.2/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:394
 [2] remotecall_fetch(::Function, ::Distributed.Worker, ::Module, ::Vararg{Any,N} where N) at /build/julia/src/julia-1.5.2/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:386
 [3] #remotecall_fetch#146 at /build/julia/src/julia-1.5.2/usr/share/julia/stdlib/v1.5/Distributed/src/remotecall.jl:421 [inlined]
[...]

PR #985, opened 2021-03-09
Temporary Execution barrier

lungben:
My first works on #298 

Still very WIP!

Use Cases:

- give a method to prevent auto-update of cells which take too long to compute to keep the notebook reactive and/or save computational resources.
- step-by-step computations for educational purposes

Suggestion:

- Allow the users to add execution barriers to Pluto cells from inside Pluto.
- These barriers block execution of the cell they are added to (i.e. located before the cell) if active.
- The barriers should be stored in the notebook.jl file, e.g. as comments.
- If the notebook.jl file is executed in Julia (without Pluto), the whole file should run, ignoring execution barriers (there would be no way to turn them off without Pluto anyhow).
- If an execution barrier is active, the cell and all cells depending on this cell should be marked gray in Pluto, so that it is obvious that they are not executed.
- When deactivating an execution barrier, the corresponding cell and all dependent cells are executed immediately and not grayed out anymore.

Technical implementation:

- [x] Execution barrier information (is there a barrier? is it active?) could be added to the Cell structs in the backend.
- [x] Save this information in notebook.jl files and load it from there
- [x] Based on which cells are deactivated and the cell dependency information, the backend decides which cells to evaluate in case of changes.
- [x] Pass the barrier information to the frontend
- [x] In the frontend, the information whether a cell is grayed out is determined based on the execution barrier information in the Cells and the cell dependency information from #891
- [x] Frontend functionality for activating and deactivating of execution barriers plus visualization of execution barrirers required.
- [x] Pass changes in barrier information back to the backend
- [x] Run cell if its execution barrier is deactivated and it does not depend on any activated barrier
- [x] Tests?
- [x] Merge current master (currently not working for me out-of-the-box)

Note: this PR is based on #891, the first commit specific to this PR is 5975d25.

# Try it out!

```julia
julia> ]
pkg> activate --temp
pkg> add https://github.com/lungben/Pluto.jl#execution_barrier
```


lungben:
Basic functionality on the backend side is there (execution barrier still needs to be added manually in the notebook.jl file).

```
### A Pluto.jl notebook ###
# v0.14.0

using Markdown
using InteractiveUtils

# ╔═╡ 826fa5c0-80d0-11eb-3ea4-779667653151
x = 7

# ╔═╡ 68c72926-8390-4e43-9a56-3b5feaa73832
# ═execution_barrier
y = 2x

# ╔═╡ c185a4af-68e8-4493-8c83-fa7eea02c833
z = sqrt(y)

# ╔═╡ 37652a74-ea98-4b09-8dd7-4422c96ee2d4
a = 3x

# ╔═╡ Cell order:
# ╠═826fa5c0-80d0-11eb-3ea4-779667653151
# ╠═68c72926-8390-4e43-9a56-3b5feaa73832
# ╠═c185a4af-68e8-4493-8c83-fa7eea02c833
# ╠═37652a74-ea98-4b09-8dd7-4422c96ee2d4
```

image

Note that the execution barrier is before the `y = 2x` cell. This cell and the cell dependent on it are not executed, the other cells are.

Note that all frontend functionality is still missing.

lungben:
image

The 1st cell is long-running. I can deactivate it by clicking on the thumb next to the cell:

image

Except for the stop sign, nothing changed yet, the results for `y` and `z` are still correct.

When I now modify `x`, the execution barrier prevents execution of the 1st cell (which defines `y`). The cells dependent on `y` are also not re-calculated (here the 2nd cell).
These cells still show their old results (before changing of `x`), they are shown in gray to indicate that they are outdated.

image

When adding a new cell or modifying an existing cell, variables defined in deactivated (gray) cells cannot be used:

image

Clicking on the stop sign of the 1st cell deactivates the execution barrier and re-executes this cell and all its dependencies (including the cell added in the previous step which raised an error before).

image

As shown in the code snippet above, execution barriers are stored as comments in the notebook.jl files, therefore the state of the barriers is remembered when closing and re-opening a notebook.

lungben:
From our call:

- [ ] Base on refacturing of cell dependency information #1008 
- [ ] Use Pluto cell output opacity to signal deactivated cells, instead of graying out the cell. Maybe use different opacities for directly (barrier) and indirectly deactivated cells?
- [ ] Show deactivated cells opaque immediately after deactivation
- [ ] Other solution for cell deactivation button - maybe as Pause button next to run button (below cell)?

marius311:
Played with this a little bit today, this is really awesome to finally see, thank you! 

One bug report in case you were aware, and two comments/suggestions. 

### Bug

Consider the following two-cell notebook where cell 2 depends on cell 1:

<details>

```
### A Pluto.jl notebook ###
# v0.14.1

using Markdown
using InteractiveUtils

# ╔═╡ a831cc1c-c08f-42b9-a434-1ed24065dec3
x = 0

# ╔═╡ c08c9239-ec89-4f6f-93d1-1051dbbd9ce0
y = x

# ╔═╡ Cell order:
# ╠═a831cc1c-c08f-42b9-a434-1ed24065dec3
# ╠═c08c9239-ec89-4f6f-93d1-1051dbbd9ce0
```
</details>

Now comment cell 1, then comment and uncomment cell 2. Cell 2 will run (and error) instead of the correct behavior which is that should remain commented since its a dependent of cell 1. 

### Comment / suggestions
1. I think it would be good to dim or even remove the traffic-light entirely for commented cells. This will make it easier to visually ignore commented cells, which as-is I found was still a little hard to do. (otherwise I thought all the visual aspects of this were great)
2. I'd reconsider the trigger of right-click on the run button. Its extremely hidden and I doubt anyone would ever find this on their own. Maybe this is the intention though since Pluto seems to be very anti-complexity and so you're trying to hide complexity. But the natural thing to me seems to add another button next to the eye, probably with the icon of a `#` to mimick comment.
3. A keyboard shortcut would of-course be great (maybe there is and I missed it). Probably some modification of Ctrl+/ which is ubiquitous for commenting a line, so maybe Ctrl+Shift+/ or something (and equivalent for Mac keyboards)

lungben:
Thanks for your feedback!

Currently, the downstream dependencies of cells are only determined when the cells are deactivated. When reactivating a 
 previously deactivated cell, it is currently not checked if it should stay (indirectly) deactivated because an upstream dependency is deactivated. But I agree that this is something which should be done!

To your suggestions:
1. Good suggestion - I'll remove the traffic light (bar left to the cells) for deactivated cells.
2. + 3. The right-click on the Run button will not be the final solution. We still have come to a conclusion about how to allow cell activation / deactivation in the GUI. But a keyboard shortcut would definitively be helpful!

Possibilities for cell deactivation / activation GUI:

1. new icon next to the run / stop icon below each cell
2. replace the delete cell button (x) by a context menu with deletion, execution barrier (and maybe more choices later)
3. more ideas?

lungben:
The traffic light is now not visible anymore for deactivated cells.
If a cell is activated which depends on a still deactivated cell, it does not error anymore (but still shows the cell to be queried in the traffic light).

lungben:
@pankgeorg looks great!

One suggestion: I think it would be good to keep the feature that the run button (below the cell) is switched to a "pause" sign when the barrier is active. Deactivation of the barrier could then be (alternatively) done by clicking on that pause button.

What do you think?

pankgeorg:
Yeap, we need decide and then tell the user what the run button (on _any_ greyed out cell) does, and it must be explicit to the user that if it runs, the barrier is off and all hell breaks loose!!

pankgeorg:
In order to avoid committing forever to have the first comment to be the exectution barrier, let's move the barrier into the session (and not change the file at all). Then we can merge and decide file changes later!

Also this gives us the flexibility to change this feature if/when we get extra feedback.
This, combined with the delayed execution at startup means that we can still 'barrier' long-running cells on startup (#259 )

Let's split out the file change and decide later!!!


(from fons in our weekly!)

lungben:
I agree - the file save of the barriers is currently not so useful because it gets removed when you deactivate your barrier, i.e. the content of your file depends on which barriers are active.
Maybe we should pick up an idea from the very beginning - give 2 options for a disabled cell, one to remove the barrier (in this case it is also removed from the file) and one to just deactivate it in your current session. For the latter, the barrier still stays in the file and will be active if you close and re-open the notebook.
This has the advantage that you can "shield" your long-running cells more reliably from being executed when opening a notebook.
What do you think?

pankgeorg:
> This has the advantage that you can "shield" your long-running cells more reliably from being executed when opening a notebook.

I agree with this idea, and there are many ways to do it, for example a macro could also work: `@defaultbarrier` or `@slowcode` (which has the extra benefit of providing semantics!!!!)

Or we could detect the `timed` cells and add barriers to these [100% backward compatibility!!] (!)

---

I've commented out the file-related code and I'd guess this PR is ready for @fonsp !

src/evaluation/Run.jl
```diff
@@ -4,6 +4,54 @@ import .ExpressionExplorer: FunctionNameSignaturePair, is_joined_funcname, Using
...
+function cell_deactivation!(cells_in:: Vector{Cell}, deactivated_uuids:: Set{UUID})
+    # activate all cells before checking which cells are affected by execution barrier
+    for cell in cells_in
+        cell.is_deactivated = false
+    end
+    # identify cells affected by active execution barrier and its references
+    for cell in cells_in
+		if cell.cell_id ∈ deactivated_uuids
+			cell.is_deactivated = true
+			cell.queued = false
```
fonsp, line 48: why `queued=false`?

fonsp (changes requested):
Can we add tests?

fonsp:
Like in the context menu, I like the term "Disable cell" more than "Add execution barrier". Can you update the variable names to reflect this?
`has_exectuion_barrier` should be `is_disabled`, `is_disabled` should be `is_indirectly_disabled`

src/evaluation/Run.jl
```diff
@@ -4,6 +4,54 @@ import .ExpressionExplorer: FunctionNameSignaturePair, is_joined_funcname, Using
...
+    for c ∈ referenced_cells
+        get_deactivated_cell_uuids!(deactivated_uuids, c)
+    end
+	nothing
+end
+
+"""
+Get UUIDs for all deactivated cells (directly or indirectly)
+"""
+function get_deactivated_cell_uuids(cells:: Vector{Cell}):: Set{UUID}
```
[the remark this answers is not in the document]
lungben: I need to identify the cells which need to be deactivated based on all cells, not only the ones scheduled for re-execution, to detect if a removed barrier should still be deactivated due to an upstream barrier.
But directly setting `is_deactivated` in the Cells of `notebook.cells_dict` did not work for me, this has not been propagated to the cells in `to_run_raw` (maybe there is an implicit copy of cells when creating the topology?). 
This is why I used this approach - first identify the cell IDs based on all cells, then deactivate these cells in `to_run_raw`.

src/evaluation/Run.jl
lungben: My original idea is that deactived cells could not be queued for execution, therefore it is safer to set it explicitly to false. But not sure if this is really needed.

lungben:
> Like in the context menu, I like the term "Disable cell" more than "Add execution barrier". Can you update the variable names to reflect this?
> `has_exectuion_barrier` should be `is_disabled`, `is_disabled` should be `is_indirectly_disabled`

Done - note that all cells which are `is_disabled` are also `is_indirectly_disabled`.

pankgeorg:
> Done - note that all cells which are is_disabled are also is_indirectly_disabled.

What about the name `is_execution_paused` or `is_blocked_from_running` or something like this?

ederag:
Great work !
Easy to discover, grayed-out cells are still readable enough to see their last output, very good.

Disabling a cell does not stop it though.
It becomes disabled only after it has fully finished,
that is, once the work has been done anyway.


> the whole file should run, ignoring execution barriers (there would be no way to turn them off without Pluto anyhow).

I'd expect the file to run exactly the same as in Pluto (disabling would be like commenting-out all dependent blocks).

Congratulations again, please let me know if this is the time and place for more feedback.

lungben:
Thanks!

> Disabling a cell does not stop it though.
> It becomes disabled only after it has fully finished,
> that is, once the work has been done anyway.

Good idea, I added a function to interrupt the current cell if the barrier is activated for it. Unfortunately this does not really help on Windows because cell interruption in general does not work there yet.

Currently, execution barriers are a feature of a Pluto session and are not persistent to the notebook.jl file. The issue with commenting them out is that the Pluto expression explorer ignores comments, therefore you loose dependency information required for dynamic activation / deactivation. I think it makes sense to store barrier information in the notebook file in some way, but more discussion on what to store and how to store are needed. 

We are planning to first merge this feature and then think about the next steps.

frontend/components/RunArea.js
```diff
@@ -1,11 +1,43 @@
...
+                            new CustomEvent("cell_focus", {
+                                detail: {
+                                    cell_id: barrier_cell_id,
+                                    line: 1, // 1-based to 0-based index
+                                },
+                            })
+                        )
+                }}
+                class="runcell"
+                title=${disable ? "Please remove the barrier" : "Run"}
```
ederag, line 37: Perhaps "Please enable the cell" would be closer to the "Enable cell" menu item ?

ederag:
> I added a function to interrupt the current cell if the barrier is activated for it.

It works, on linux, good !

When the cell is running and has to be interrupted, 
too many cells display `InterruptException`, instead of only being grayed out, but it's probably related to #151,
and some have to be restarted manually (this is *not* new). None of this seem related to this PR.

Thinking about that, from a user point of view, interrupting and disabling have a lot in common:
Disable > Interrupt if running > Gray out cell and all dependent cells, showing their last clean state
When the "Interrupt" icon is pressed, the "Disable" step is skipped.
Just a hunch: the mechanism this PR has introduced might be a way to solve the interrupt issue as well ?
(When all cells have been given time to finish, 
it is perfect: on a Disable, parents are not affected, and after a Re-enable, all children run automatically)

> I think it makes sense to store barrier information in the notebook file in some way, 

Indeed, that's what I had in mind: Pluto-specific tag + comment or `if...end`.

> We are planning to first merge this feature and then think about the next steps.

Sounds fine !

lungben:
> When the cell is running and has to be interrupted,
> too many cells display `InterruptException`, instead of only being grayed out, but it's probably related to #151,
> and some have to be restarted manually (this is _not_ new). None of this seem related to this PR.

I think this is due to this code part:

```
interrupt_remote: (cell_id) => {
                // TODO Make this cooler
                // set_notebook_state((prevstate) => {
                //     return {
                //         cells: prevstate.cells.map((c) => {
                //             return { ...c, errored: c.errored || c.running || c.queued }
                //         }),
                //     }
                // })
                this.client.send("interrupt_all", {}, { notebook_id: this.state.notebook.notebook_id }, false)
            },
```
The interruption is a global event which interrupts all running cells. There is currently no way for the execution barrier code to interrupt only the current cell without interrupting the dependent cells, but instead deactivate the dependent cells. This would require a refactoring of the interruption code.

frontend/components/RunArea.js
lungben: done

ederag:
> This would require a refactoring of the interruption code.

Then it might be better to postpone 812fd489f505680be141d6fbddac4a36a0ec5c91;
meanwhile the experience will be less responsive, but a lot cleaner.

lungben:
The issue is not only for parent cells, but also for all other cells which are scheduled for running when a cell is deactivated, even if they are completely unrelated. Therefore, I also think that 812fd489f505680be141d6fbddac4a36a0ec5c91 should be reverted.

lungben:
From my side, this PR is ready to merge.
@fonsp any points from your side?

ederag:
I have used this for a few hours, with a few limitations that can be lifted later, it seems solid, 
and is already an improvement, thanks !

fonsp:
Looks good! Thanks again for your patience and for working together on getting this done!

Before we merge, I would like to:
- [ ] Tweak the CSS and JS
- [ ] Try to break it :)

I will work on it right now!

fonsp:
I reused the existing `topological_order` function to find the indirectly disabled cells: [`47934ce` (#985)](https://github.com/fonsp/Pluto.jl/pull/985/commits/47934ce38b5d095363a9d3d7f25ef525e58946bc) This makes the PR much smaller :)

fonsp (approved):

lungben:
Cool, thanks!

oschulz:
Yay!!

rajithv:
Does the temporary execution barrier work in 0.14.8? I'm trying to find how to use this feature. Thanks!

lungben:
Yes. You can activate it by pressing the (...) button at the right side of the cell and then "Disable Cell".

PR #988, opened 2021-03-10
🌈 Interactive @bind on a static preview

fonsp:
Also see the bind server: https://github.com/fonsp/PlutoBindServer.jl/tree/main

https://user-images.githubusercontent.com/6933510/103956853-a8b79c80-5149-11eb-86bf-e26e5e66fabd.mov

These requests are **stateless**, which means that they can be **cached** (by the browser and by cloudflare/ingress/thingy). In the video below, you see that the first couple of request take some time to process (due to network lag), but they start building up a cache. After a while, all requests are served straight from the cache, and the bind server is idle:

https://user-images.githubusercontent.com/6933510/106898023-f8ab6400-66f3-11eb-868c-0897afdfd70a.mov


frontend/components/Editor.js
```diff
@@ -651,17 +654,107 @@ patch: ${JSON.stringify(
...
+        }
 
         this.on_disable_ui = () => {
             document.body.classList.toggle("disable_ui", this.state.disable_ui)
             document.head.querySelector("link[data-pluto-file='hide-ui']").setAttribute("media", this.state.disable_ui ? "all" : "print")
-            //@ts-ignore
-            this.actions = this.state.disable_ui ? fake_actions : real_actions //heyo
+            if (use_slider_server) {
+                this.actions = this.state.disable_ui ? fake_actions : real_actions //heyo
+            }
```
fonsp, line 753: why did i do this

fonsp:
TODO: test

PR #990, opened 2021-03-10
let `JULIA_NUM_THREADS` have higher priority

Moelf:
prioritize user-set `JULIA_NUM_THREADS`


fonsp:
Thanks! Can you send a screenshot demonstrating that this works?

Moelf:
image
then:
image

PR #1000, opened 2021-03-15
fix anonymous functions with keyword args

### LINKED ISSUE #997: UndefVarError: a function reference inside of structure
image

code
```julia
import Pkg
Pkg"add https://github.com/mmikhasenko/AlgebraPDF.jl"
#
using AlgebraPDF
#
f1(x) = x^2
d = pdf((x;p)->f1.(x), ∅, (2,6))
d.f(1.1; p=∅)
```
Something known?

Pangoraw:
Interestingly, the parameter tuple of `(x;p) -> x` gets parsed as a block so `begin x; p end -> x` is a valid syntax for an anonymous function :exploding_head: 

Closes #997


fonsp:
Wow, that's bizarre! In one of your previous PRs you linked to Julia's devdocs about AST structure. Is this documented? Maybe we should write a PR to those docs?

Pangoraw:
It does not seem documented, there is nothing about anonymous functions :thinking:

PR #1010, opened 2021-03-19
Ignore assignment to underscores

mcabbott:
It's common assign throw-away values to `_`; this changes ExpressionExplorer to ignore them:

Screenshot 2021-03-19 at 12 46 22

In Pluto right now, this example gives a multiple assignment error, since two cells appear to write to the same variable. But they don't really, since `_` cannot be read from.

~~(Not sure why tests seem to see a dependence on `_`, even though they correctly don't think it's being written to.)~~

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -175,8 +175,8 @@ function get_assignees(ex::Expr)::FunctionName
     end
 end
 
-# e.g. x = 123
-get_assignees(ex::Symbol) = Symbol[ex]
+# e.g. x = 123, but ignore _ = 456
+get_assignees(ex::Symbol) = ex === :_ ? Symbol[] : Symbol[ex]
 
 # When you assign to a datatype like Int, String, or anything bad like that
 # e.g. 1 = 2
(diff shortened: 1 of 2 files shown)


pankgeorg:
I understand correctly that this is the expected julia behaviour right?
image

Can you add a reference to the docs, if any, just for completeness?

Pangoraw:
> (Not sure why tests seem to see a dependence on _, even though they correctly don't think it's being written to.)

To account for special cases like this:
```julia
x[a+b] = ... # will reference a, b and x, will assign to x
```
Pluto explores the left side of the assignment like any other expression. It then removes from the references any assignments, see the `setdiff` in `ExpressionExplorer.jl`:

https://github.com/fonsp/Pluto.jl/blob/a454b61a14e9116e3c7b04dc9bea70f41ed019ea/src/analysis/ExpressionExplorer.jl#L317

If `_` cannot store values between cells, perhaps it would be simpler to not consider them in the references as well.

mcabbott:
Thanks, that's what I couldn't find. I think the last commit should remove them from being seen as references too. 

I'm not actually sure where this is documented, it's discussed e.g. [here](https://discourse.julialang.org/t/usage-rules-for-bare-underscores/28858).

fonsp (changes requested):
>  I think the last commit should remove them from being seen as references too.

Can you add a test?

image

Can you extend this PR to any "all-underscore" variable name?

mcabbott:
There are some tests, could expand? 

Didn't think of multiple underscores, but have now expanded it to handle these too.

fonsp:
Can you add a test for this remark:

> I think the last commit should remove them from being seen as references too.

mcabbott:
This is the change that went with that remark: 

Screenshot 2021-03-22 at 08 53 17

fonsp (approved):

fonsp:
Thanks @mcabbott !

PR #1032, opened 2021-03-26
Analyse macros

Pangoraw:
See https://github.com/fonsp/Pluto.jl/issues/196#issuecomment-800485127 for more information

<details>
<summary>Previous update</summary>

pluto_macros_demo

</details>

##### Update 2
With the help of #1120:

macro_update

To see what works and what doesn't, take a look at the tests in [MacroAnalysis.jl](https://github.com/Pangoraw/Pluto.jl/blob/macros/test/MacroAnalysis.jl)

<details>
<summary>Support for <code>@odedef</code></summary>

ode_def
</details>

<details>
<summary>Plays nicely with the Queryverse</summary>

queryverse
</details>

TODOs:
 - [x] Update tests
 - [x] Integration with [PlutoSliderServer.jl](https://github.com/JuliaPluto/PlutoSliderServer.jl), currently `load_notebook()` uses [`static_resolve_topology()`](https://github.com/Pangoraw/Pluto.jl/blob/macros/src/evaluation/Run.jl#L204) so only a few base macros (like `@bind`, `@enum`,...) will be reactive.
 - [x] Investigate clunky `result not defined` in above gif
 - [x] Macro calls inside function scopes
 - [x] make sure error line numbers are consistent


fonsp:
(I think the frontend tests will pass when you merge main into your branch)

fonsp:
Amazing! I'll review it more closely this week, but so far it looks great!

Could you quickly summarise again which things do _not_ work yet? I noticed that some tests run the notebook cells one by one, is a batched reactive run not yet supported there?

Pangoraw:
> Could you quickly summarise again which things do not work yet? I noticed that some tests run the notebook cells one by one, is a batched reactive run not yet supported there?

Sure! There are a few cases where the batched reactive update would not work (with a fallback to the previous static behavior) or not do exactly what is expected.

1. When batching both the usage of a macro and its definition. Example:
 https://github.com/fonsp/Pluto.jl/blob/2a792f900a069074816c9725fbb52ff11aff3390/test/MacroAnalysis.jl#L35-L40
  In this case, Pluto will not have the definition of `my_macro` so it will not be able to macro expand. The second cell will count as referencing `x`, not assigning to it. However, if you define `my_macro` and then execute the call cell, it will work.
  The same is true for `import Blabla: @my_macro` instead of `macro my_macro()` unfortunately since this also counts as a assignment but it could be improved by collecting the imports as the usings in the ReactiveNodes.
2. Related to 1., if you submit an update to both the macro definition and its calls, the old version of the macro will be executed. However, if you update only either one it will work. This can be confusing, but it is easy to detect.
3. Because macros that fail to expand are executed in the previous workspace, if you use `@__MODULE__` and a user defined macro in the same cell, the `@__MODULE__` will be the previous one.

If you are using macros with `using MyPackage` then you should be fine, even for the PlutoSliderServer where the cases above might fail or create an invalid topology.

fonsp:
To solve 1 & 2, perhaps we can switch workspaces (by deleting 0 variables) after evaluating a cell that contains an import or a macro definition, during the reactive run? (Does this make sense?)

Pangoraw:
> To solve 1 & 2, perhaps we can switch workspaces (by deleting 0 variables) after evaluating a cell that contains an import or a macro definition, during the reactive run? (Does this make sense?)

For imports we could extract them and run them as `using` in the new module:
https://github.com/fonsp/Pluto.jl/blob/2d423bad6f6d51cd7f9c8d8653ab32870fb85b30/src/evaluation/Run.jl#L197-L202

For macro definition, extracting the macro definition and running it in the previous module could work but if there are other new variables as part of the reactive run that are used in the macro, the macro definition+call will not work.

fonsp:
@Pangoraw  What if we add a line below https://github.com/Pangoraw/Pluto.jl/blob/b7b441f0c9fdb1f33bdc3a283a209fc49587c780/src/evaluation/Run.jl#L101

(pseudocode)

```julia
remaining = to_run[i+1:end]

if (
	any(has_unresolved_macrocall, remaining) && 
	(
		# a new macro definition might be available if:
		defines_macro(cell) ||
		imports_package(cell)
	)
)
	put!(notebook.executetoken)

	new_new_topology = resolve_topology(session, notebook, new_topology, current_workspace_name)
	# use remaining cells as roots for the next reactive run, and exit the current one.
	return run_reactive!(session, notebook, new_topology, new_new_topology, remaining; deletion_hook=deletion_hook, persist_js_state=persist_js_state)
end
```

fonsp:
We will make the PR work for `import`ed macros and then it's ready to merge!

Pangoraw:
Support for `imports` added in https://github.com/fonsp/Pluto.jl/pull/1032/commits/94aecd54466df40e858e5e5a2de977d0ef225fe8 :+1:

src/evaluation/Run.jl
```diff
@@ -171,19 +172,104 @@ end
...
 "Do all the things!"
 function update_save_run!(session::ServerSession, notebook::Notebook, cells::Array{Cell,1}; save::Bool=true, run_async::Bool=false, prerender_text::Bool=false, kwargs...)
 	old = notebook.topology
-	new = notebook.topology = updated_topology(old, notebook, cells)
 
-	update_dependency_cache!(notebook)
+	old_workspace_name, new_workspace_name = WorkspaceManager.bump_workspace_module((session, notebook))
+
+	unresolved_topology = updated_topology(old, notebook, cells)
+	new = notebook.topology = resolve_topology(session, notebook, unresolved_topology, old_workspace_name)
```
fonsp, line 264: To avoid race conditions, we have a `notebook.executetoken` to ensure that all `run_reactive!` calls are consecutive. `bump_workspace_module` and `resolve_topology` should be called _inside_ `run_reactive!`, when the token is taken, to ensure that they happen within the same run.

Right now, it is possible that `update_save_run!` is called while another call is busy, and it will switch workspaces and expand macros during another run.

fonsp (changes requested):

src/evaluation/Run.jl
fonsp: let me take a shot a this...

fonsp: pretty much done now, some cleanup left

fonsp:
Initial bond values are not set:

image

fonsp:
I added tests on Julia 1.7.0-beta2, and they are failing. I am not sure whether this is because of one of my other changes or not.

fonsp:
@Pangoraw  I made some changes:
- Run `resolve_topology` inside `run_reactive!` to fix race condition: https://github.com/fonsp/Pluto.jl/pull/1032#discussion_r657034015
- Fixed the `is_just_text` mechanism for running markdown cells on the server before running the notebook
- Cleaned up `delete_vars` vs `move_vars`

fonsp:
Does this mean that we can delete https://github.com/fonsp/Pluto.jl/blob/0fe8bf681fa062bc427585d9b2c24964846f1f9a/src/analysis/ExpressionExplorer.jl#L776-L813 ?

Pangoraw:
I fixed the problem with 1.7, it was because the errors are no longer wrapper in a `LoadError`.

> Does this mean that we can delete the mock macros?

Good catch!

fonsp:
I did a simple notebook with Symbolics.jl but it did not work, I evaluated cell #3 two times:

image

I added this as a test case, which is now failing. @Pangoraw can you take a look?

fonsp:
I added another (failing) test where defining a macro _after_ using it creates a problem, can you take a look?

Pangoraw:
The `Symbolics` problem is due to serialization issue between the pluto and notebook process, I will make sure they are better reported so they are all detected. However, the test is still failing when the serialization is fixed because the `Pkg` instructions have not run yet so `Symbolics` is not installed yet :thinking: I will look into the other user defined macro test

fonsp:
Ah that's right, we would also need a 'batched run' to handle the `begin Pkg.add using ... end` cells. Though notebooks that just call `using Symbolics` (with the Pluto package manager) will work just fine. That's another reason to wait with merging until we have implemented a batched run system. Maybe you can take another crack at implementing it?

fonsp:
For the test that I added with the user-defined macro, I think the problem is:
- We define the macro in cell 2 and run cell 2, this will do a reactive run of cells `[2, 1]`. 
- During `run_reactive!`, but _before_ running the cells, hence _before_ running the macro definition, it will "resolve" cell 1, which is now marked as "resolved", even though the macro was not yet defined.
- Since we don't explore inside macros in the first pass, `f` is not detected as reference.
- Running 3 does to reactively trigger 1.

Solution: 'batched run' or: only mark macros as "resolved" if macroexpansion was successful?

fonsp:
image

Pangoraw:
> So that's another reason to wait with merging until we have implemented a batched run system. Maybe you can take another crack at implementing it?

I think you are right and a batched run system is the only solution that will work in most cases. I will take a crack at it

Pangoraw:
@fonsp I started a first implementation of your suggestion and it seems good so far 🎉. I still have a bit of cleanup and extra code to remove.

fonsp:
Awesome work! Can you pass the `token` to the sub-run, or something similar, to avoid race conditions?

src/evaluation/Run.jl
```diff
@@ -99,12 +116,18 @@ function run_reactive!(session::ServerSession, notebook::Notebook, old_topology:
...
+			return run_reactive!(session, notebook, new_topology, new_new_topology, to_run[i+1:end]; deletion_hook=deletion_hook, persist_js_state=persist_js_state, already_in_run=true)
+		end
 	end
-	
+
 	notebook.wants_to_interrupt = false
 	flush_notebook_changes()
 	# allow other `run_reactive!` calls to be executed
-	put!(notebook.executetoken)
+	already_in_run || put!(notebook.executetoken)
```
Pangoraw, line 130: I don't think there need to be a condition here, since there is only one call to reactive_run! which will arrive here because the previous calls return early:
https://github.com/fonsp/Pluto.jl/blob/a208d840d567efe1f431fb5f8dfde98a25b5689e/src/evaluation/Run.jl#L120-L124

fonsp:
Can we make `@macroexpand` depend on the macro definition?

https://user-images.githubusercontent.com/6933510/124311302-d0724500-db6d-11eb-8412-d249df4d69b6.mov

Pangoraw:
> Can we make `@macroexpand` depend on the macro definition?

The problem is not really with macro analysis. We can add cases in the ExpressionExplorer for calls to `Base.macroexpand` and `Core.eval` to explore the second argument.

```julia
julia> @macroexpand @macroexpand @enum A a b c
:(Base.macroexpand(Main, $(QuoteNode(:(#= REPL[1]:1 =# @enum A a b c))), recursive = true))

julia> @macroexpand @eval Main 1+1
:(Core.eval(Main, $(Expr(:copyast, :($(QuoteNode(:(1 + 1))))))))
```

fonsp:
How is the PR going @Pangoraw ? Can you remind me what the TODO list is?

Pangoraw:
> How is the PR going @Pangoraw ? Can you remind me what the TODO list is?

I just added reactivity to `@macroexpand`. I have used the PR in this state a bit and have not found other issues for now. Maybe we should proceed like with PlutoPkg by inviting people to test on `main` once merged ?

fonsp:
Cool! @Pangoraw feel free to merge yourself and post the message

fonsp (approved):

PR #1036, opened 2021-03-27
Trailing throttle syncs to front-end

pankgeorg:
I'm using a slightly modified version of `Flux.jl`'s throttle: https://github.com/FluxML/Flux.jl/blob/8afedcd6723112ff611555e350a8c84f4e1ad686/src/utils.jl#L662

### DIFF
--- a/src/evaluation/Run.jl
+++ b/src/evaluation/Run.jl
@@ -48,7 +48,7 @@ function run_reactive!(session::ServerSession, notebook::Notebook, old_topology:
 
 	# Send intermediate updates to the clients at most 20 times / second during a reactive run. (The effective speed of a slider is still unbounded, because the last update is not throttled.)
 	# flush_send_notebook_changes_throttled, 
-	send_notebook_changes_throttled, flush_notebook_changes = throttle(1.0/20; leading=false, trailing=true) do
+	send_notebook_changes_throttled, flush_notebook_changes = throttled(1.0/20; leading=false, trailing=true) do
 		send_notebook_changes!(ClientRequest(session=session, notebook=notebook))
 	end
 	send_notebook_changes_throttled()
@@ -195,7 +195,7 @@ update_run!(args...) = update_save_run!(args...; save=false)
 
 
 """
-		throttle(f, timeout; leading=true, trailing=false)
+		throttled(f, timeout; leading=true, trailing=false)
 
 Return a function that when invoked, will only be triggered at most once
 during `timeout` seconds.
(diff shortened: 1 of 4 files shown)


pankgeorg:
Converting this to draft as I would like to check whether we can add a lock to the function call; I don't trust concurrency 🎈✨⭐

pankgeorg:
Added 'thread safety', which means that the function _cannot run two times simultaneously_.
The caveat is that (in tightly called sequences) the timeout time gets extended by the time the function is running.

The last thing to consider is, do mutations that may happen 'simultaneously' matter? should every throttled function call get its own copy of the state??

fonsp:
In the first paragraph of your comment, aren't you saying that the second paragraph is not possible?

fonsp:
Can you add some super basic tests with `sleep` to make sure that the basic throttling is working? The code got a bit too complicated to just review it

pankgeorg:
> In the first paragraph of your comment, aren't you saying that the second paragraph is not possible?

Not exactly - I'm saying that if `f` is throttled by the function, and has `argument1`, `argument2` as arguments, then, `f` will never run when `f` runs. (first paragraph)

BUT if `g` is mutating `argument1`, `argument2` there is no guarantee that `f` and `g` don't run such as they corrupt each other's data. (second paragraph)

I'm not even sure if that's possible by Julia's thread model (since our `f` doesn't yield, but it may do some network stuff that counts as 'idle'??)

The thing is that we'll probably need to be locking the resource (`argument1`, `argument2` = `session`, `notebook`).

fonsp:
Why does `f` take arguments?

src/evaluation/Run.jl
```diff
@@ -47,7 +47,8 @@ function run_reactive!(session::ServerSession, notebook::Notebook, old_topology:
 	end
 
 	# Send intermediate updates to the clients at most 20 times / second during a reactive run. (The effective speed of a slider is still unbounded, because the last update is not throttled.)
-	send_notebook_changes_throttled = throttled(1.0/20, 0.0/20) do
+	# flush_send_notebook_changes_throttled, 
+	send_notebook_changes_throttled, flush_notebook_changes = throttle(1.0/20; leading=false, trailing=true) do
 		send_notebook_changes!(ClientRequest(session=session, notebook=notebook))
```
pankgeorg, line 52: This code will run at most once per 1/20 sec, `@async`. 
BUT there is no guarantee I can think of that [X] code will not run right before/right after/inbetween

fonsp:
>  locking the resource (argument1, argument2 = session, notebook).

We already lock `notebook`:

https://github.com/fonsp/Pluto.jl/blob/7005c6bb8d236e297962386e3e5d53ed4c9b378e/src/evaluation/Run.jl#L10

pankgeorg:
> Why does `f` take arguments? It makes more sense to drop that feature

The feature is not related; the `() -> send_notebook_changes!(ClientRequest(session=session, notebook=notebook))` holds references to `session` and `notebook`. 

> > locking the resource (argument1, argument2 = session, notebook).
> 
> We already lock `notebook`, right?
> 
> https://github.com/fonsp/Pluto.jl/blob/7005c6bb8d236e297962386e3e5d53ed4c9b378e/src/evaluation/Run.jl#L10

all these functions all hold the lock, I think.

fonsp:
> The feature is not related

I see, I can try to think about the thread safe stuff tomorrow with a CLEAR brain :) It does seem like the code would be simpler if you deleted the args, kwargs, and result features. `later` could just be a boolean, for example.

PR #1046, opened 2021-04-01
Fix ServiceWorker

fonsp:
I always wondered why this page was empty:

Schermafbeelding 2021-04-01 om 23 20 09

It turns out that the serviceworker was never actually working! `caches.match` does not do anything, you should first get the cache, then match the found cache.

Schermafbeelding 2021-04-01 om 23 20 22

I am also undoing https://github.com/fonsp/Pluto.jl/commit/0b62cdc79c13532147d42eea5725d2f9b4b188d6


pankgeorg:
It seems like adblock was also interfering with the service worker; I had to disable it locally to work. Other than that, seems to work like a charm!

frontend/sw.js
```diff
@@ -1,55 +1,71 @@
...
     event.waitUntil(
         caches.open(CACHE_NAME).then(function (cache) {
-            console.log("Opened cache")
+            logger.log("Opened cache")
         })
     )
 })
 
-const whiteList = ["www.gstatic.com", "fonts.gstatic.com", "fonts.googleapis.com", "cdn.jsdelivr.net", "cdnjs.cloudflare.com", "unpkg.com"]
+const allowList = ["www.gstatic.com", "fonts.gstatic.com", "fonts.googleapis.com", "cdn.jsdelivr.net", "cdnjs.cloudflare.com", "unpkg.com"]
```
pankgeorg, line 23: Updated this

pankgeorg (approved):

frontend/sw.js
```diff
@@ -1,56 +1,67 @@
 // mostly based on https://developers.google.com/web/fundamentals/primers/service-workers
 
-var CACHE_NAME = "pluto-cache-v1"
+const DEBUG = true
```
pankgeorg, line 3: @fonsp shouldn't this be "false" for "production"?

fonsp:
oops

fonsp:
fixed

PR #1052, opened 2021-04-02
What you see is what you REST

ctrekker:
Unlike the previous merge request, this one should only include the commits I authored.

### DIFF
--- a/src/Pluto.jl
+++ b/src/Pluto.jl
@@ -45,6 +45,7 @@ include("./webserver/Static.jl")
 include("./webserver/Dynamic.jl")
 include("./webserver/REPLTools.jl")
 include("./webserver/WebServer.jl")
+include("./webserver/REST.jl")
 
 if get(ENV, "JULIA_PLUTO_SHOW_BANNER", "1") !== "0"
 @info """\n
--- a/src/Pluto.jl
+++ b/src/Pluto.jl
@@ -12,6 +12,7 @@ module Pluto
 project_relative_path(xs...) = normpath(joinpath(dirname(dirname(pathof(Pluto))), xs...))
 
 import Pkg
+export PlutoNotebook, @resolve
 
 const PLUTO_VERSION = VersionNumber(Pkg.TOML.parsefile(project_relative_path("Project.toml"))["version"])
 const PLUTO_VERSION_STR = 'v' * string(PLUTO_VERSION)
--- a/src/webserver/REST.jl
+++ b/src/webserver/REST.jl
@@ -1,0 +1,59 @@
+module REST
+import ..Pluto: ServerSession, Notebook, NotebookTopology, Cell, FunctionName, WorkspaceManager, where_assigned, where_referenced, update_save_run!
+import UUIDs: UUID
+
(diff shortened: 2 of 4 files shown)


lungben:
Awesome!

With this, one could build powerful micro-services with very little effort. Thinking about use cases e.g. in financial industry, it would be important to protect the notebooks, so that the user of the API cannot modify the notebook (i.e. the running microservice) itself.
For this, the following features would be great in my opinion:

1. The notebook should define which variables are modifyable with the API and which variables should be requestable. Possible syntax:
```
@listen a, b
@publish y, z
```
Variables not explicitly defined should not be accessible from the API. Maybe the default could be to listen / publish all variables, unless `@listen` / `@publish` is explicitly written in the notebook?

2. It should be possible to allow users access of the API without allowing them to open the notebook itself. This could e.g. be archieved by using different secrets.

ctrekker:
Your first point makes a lot of sense! Not only would explicitly defining which variables users can change via the API (in production/deployment settings) enhance notebook protection, but it would also allow for further optimization since execution plans can be computed beforehand and only once.

For your second point, I was thinking about adding a configuration option which would allow users to run Pluto in an "API only" mode which would disable notebook editing and access features, restricting use to the API only. Perhaps this mode will also default to only serving "endpoints" defined by the `@listen` and `@publish` macros mentioned earlier and will otherwise fall back to serving all variables but give a warning message in the console about notebook security.

lungben:
Would it be possible to put the functionality for accessing the Pluto REST API (in file REST.jl, like PlutoNotebook) into a separate little package, e.g. called PlutoRESTClient?
When it is in Pluto, you have to import the complete Pluto package in your client just to use the REST API access functions.

ctrekker:
@lungben It's definitely overkill to have to import all of Pluto, so putting it in a separate package that main Pluto would import inside `PlutoRunner` might work better. Perhaps the REST functions live in a different package, `PlutoRESTClient.jl`, and is exported from `Pluto.jl` as well?

Roger-luo:
+1 for a seperate package, I'd like to use this as a workspace backend for CLI scripts etc. too, see also #1073

src/Configuration.jl
```diff
@@ -21,6 +21,7 @@ Base.@kwdef mutable struct ServerOptions
     disable_writing_notebook_files::Bool = false
     notebook::Union{Nothing,String} = nothing
     simulated_lag::Real=0.0
+    enable_rest::Bool = false
```
lungben, line 24: Is there a significant performance impact of having the REST API enabled?
Otherwise, I would suggest to enable it as default.

ctrekker: There isn't any performance impact, but right now there are zero security features at all. It's probably not a big deal though, so I will update that

ctrekker:
> Would it be possible to put the functionality for accessing the Pluto REST API (in file REST.jl, like PlutoNotebook) into a separate little package, e.g. called PlutoRESTClient?

The client functionality is now in a separate package https://github.com/JuliaPluto/PlutoRESTClient.jl. I haven't published it nor removed the client code from the PR yet though

fonsp:
@ctrekker frontend tests are fixed on `main`!!! ✅✅✅

PR #1055, opened 2021-04-03
📂 Auto-save when the Expr did not change

fonsp:


fonsp:
@Pangoraw this can be achieved using our new execution model

Pangoraw:
Can we do this on the frontend using the lezer tree?

PR #1024, opened 2021-03-25
Support syntax highlight for lots of languages in Markdown

schneiderfelipe:
This adds support for syntax highlight for many languages, as suggested/commented [here](https://github.com/fonsp/Pluto.jl/issues/235#issuecomment-794362160) (issue #235 is probably related).

Captura de tela de 2021-03-25 11-41-17

It uses CodeMirror, so no new dependencies here, except for the [`loadmode` addon](https://codemirror.net/doc/manual.html#addon_loadmode) (~1.6 kB) for lazily loading new languages as they appear. Julia is always preloaded, of course.

PR #1061, opened 2021-04-06
Add aliases for syntax highlighting

schneiderfelipe:
This is a continuation of #1024. It adds the following language aliases to syntax highlighting ([comment](https://github.com/fonsp/Pluto.jl/pull/1024#issuecomment-814257610)):

- `"html" => "htmlmixed"`
- `"jl" => "julia"`
- `"js" => "javascript"`

Captura de tela de 2021-04-06 15-43-37

### DIFF
--- a/frontend/components/CellOutput.js
+++ b/frontend/components/CellOutput.js
@@ -333,8 +333,19 @@ export let RawHTMLContainer = ({ body, persist_js_state = false, last_run_timest
                 for (let code_element of container.current.querySelectorAll("code")) {
                     for (let className of code_element.classList) {
                         if (className.startsWith("language-")) {
+                            let aliases = {
+                                "html": "htmlmixed",
+                                "jl": "julia",
+                                "js": "javascript",
+                            }
+
+                            let language = className.substr(9)
+                            if (language in aliases) {
+                                language = aliases[language]
+                            }
+
                             // Remove "language-"
-                            highlight(code_element, className.substr(9))
+                            highlight(code_element, language)
                         }
                     }
                 }


frontend/components/CellOutput.js
```diff
@@ -333,8 +333,19 @@ export let RawHTMLContainer = ({ body, persist_js_state = false, last_run_timest
                 for (let code_element of container.current.querySelectorAll("code")) {
                     for (let className of code_element.classList) {
                         if (className.startsWith("language-")) {
+                            let aliases = {
+                                "html": "htmlmixed",
+                                "jl": "julia",
+                                "js": "javascript",
+                            }
```
schneiderfelipe, line 340: More aliases could be added here (e.g., `jldoctest`).

(Another thing: maybe this variable should be 'const` and outside of the local scope?)

fonsp:
Really nice, thank you!

It would be great if we can contribute this to codemirror itself. Maybe a good first step is to open an issue to 'offer' the PR before spending time on it, maybe they don't want the aliases to live on the codemirror side.

frontend/components/CellOutput.js
fonsp: Style-wise, declaring it here seems like the best option.

Performance-wise, that seems like over-optimization, and the JS compilers might actually be clever enough to do it automatically.

schneiderfelipe: Great, I'll leave it like that.

schneiderfelipe:
> It would be great if we can contribute this to codemirror itself. Maybe a good first step is to open an issue to 'offer' the PR before spending time on it, maybe they don't want the aliases to live on the codemirror side.

Yes, that would be nice to have, especially `jl`. But I just realized that the language code is associated with a file name in <https://cdn.jsdelivr.net/npm/codemirror@5.58.1/mode/> as well, so that would imply creating a symlink in the server too:

https://github.com/fonsp/Pluto.jl/blob/2ebdaaf615c9491ba0b04271a9214665b5a3d5b5/frontend/components/CellOutput.js#L364

So, even if they accept an alias, I think we would still have to check for the right filename.

schneiderfelipe:
> So, even if they accept an alias, I think we would still have to check for the right filename.

@fonsp I can confirm that this is the case (the problem was the filename), as `js` is already an alias for `javascript` ([see here](https://github.com/codemirror/CodeMirror/blob/7e2466f1653faf3832302fdee508ad886d5431e7/mode/meta.js#L73-L74)). In all cases,  codemirror/CodeMirror#6644 will hopefully add the `jl` alias.

fonsp:
Hm but `html` is an extension for `htmlmixed` but that did not work...

schneiderfelipe:
> Hm but `html` is an extension for `htmlmixed` but that did not work...

I think I know what's going on. We have to use the facilities in `mode/meta.js` (such as [CodeMirror.findModeByName](https://github.com/codemirror/CodeMirror/blob/bd1b7d2976d768ae4e3b8cf209ec59ad73c0305a/mode/meta.js#L212-L221)). This way, there's no need to hardcode stuff as in https://github.com/fonsp/Pluto.jl/blob/bac224bf655a30d7b9ccaa9d66a53c23f0f33012/frontend/components/CellOutput.js#L336-L340

PR #1063, opened 2021-04-06
Copy new state-based export from PlutoSliderServer

fonsp:
and remove the weird DOM-manipulation export code

i also removed the PDF-on-firefox warning because firefox has had some updates and their PDF export is great now

@pankgeorg


fonsp:
@Ellipse0934 exporting notebooks in now built-in to the _julia_ side of Pluto, so you no longer need PlutoUtils#export (deprecated) or PlutoSliderServer.jl to generate the html files, you just do:

```julia
julia> import Pluto

julia> s = Pluto.ServerSession();

julia> nbfile = "./vintage.jl"
"./vintage.jl"

julia> nb = Pluto.SessionActions.open(s, nbfile; run_async=false);

julia> html_contents = Pluto.generate_html(nb);
┌ Info: Using CDN for Pluto assets:
└   cdn_root = "https://cdn.jsdelivr.net/gh/fonsp/Pluto.jl@0.14.1/frontend/"

julia> write("export.html", html_contents);

julia> Pluto.SessionActions.shutdown(s, nb)

```

fonsp:
@ctrekker You can combine that with PDF generation right?

ctrekker:
Yes, it will require a couple of changes, but this means I no longer have to start a Pluto server and can instead just render the static HTML inside Puppeteer

PR #1074, opened 2021-04-09
Demo of using reactivity info

lungben:
Demo of @fonsp with the visualization examples commented out in Cell.js.

I will use this for my PlutoCon talk today. Is it OK to have this commented out code also in Main until someone implements a nice frontend for the cell dependency visualization?


lungben:
this PR is obsolete now due to #1144 and #985

PR #1089, opened 2021-04-13
Saving file to non-existent directory

### LINKED ISSUE #1080: Saving a file in a folder which doesn't exist doesn't create a new folder
I was trying to save a Pluto notebook in a path that doesn't exist. This is how the directory looked before:

image

Now I wanted to create a new `submission` sub-directory inside the `julia` directory and save the file there.

image

However, neither a new directory was created nor I got any errors. I think we can create the directories just like `mkdir -p` does in linux.

This feature is supported by GitHub web dashboard while adding a new file.

wintersm1th:
Creating path before saving the file
Close #1080

### DIFF
--- a/src/notebook/Notebook.jl
+++ b/src/notebook/Notebook.jl
@@ -120,7 +120,7 @@ end
 
 function open_safe_write(fn::Function, path, mode)
     file_content = sprint(fn)
-    mkpath(splitdir(path)[1])
+    mkpath(dirname(path))
     open(path, mode) do io
         print(io, file_content)
     end
--- a/src/notebook/Notebook.jl
+++ b/src/notebook/Notebook.jl
@@ -120,13 +120,14 @@ end
 
 function open_safe_write(fn::Function, path, mode)
     file_content = sprint(fn)
+    mkpath(splitdir(path)[1])
     open(path, mode) do io
         print(io, file_content)
     end
 end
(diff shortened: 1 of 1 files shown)


pankgeorg:
Hi @wintersm1th! kudos for your first PR!

src/notebook/Notebook.jl
```diff
@@ -120,13 +120,14 @@ end
 
 function open_safe_write(fn::Function, path, mode)
     file_content = sprint(fn)
+    mkpath(dirname(path))
```
greimel, line 123:
```suggestion
    dir = dirname(path)
    isdir(dir) || mkpath(dir)
```

fonsp:
@greimel Isn't the purpose of `mkpath` that we don't need to do that?

greimel:
yes, that seems right. Sorry for the noise! (I thought it behaves like `mkdir` but that it creates multiple nested folders if necessary.)

fonsp:
@greimel No worries, thanks for contributing!

fonsp:
@wintersm1th deleted the branch

hungpham3112 (commented):
I think it should be `mkpath(path)` see #2708. If already exist `~/dir1/` but user want to create `~/dir1/dir2/dir3/` then error  appear.

PR #1090, opened 2021-04-13
GR performance: SVG, not HTML, and use PNG for large plots

### LINKED ISSUE #577: Blank frames make reactive Pluto plots look choppy
In the Julia REPL, you can use Plots.jl to make a plot on one line and update it in the next like
```julia
p = plot()
plot!(rand.(3,3), st=:heatmap)
```

Any additional calls to `plot!` will update the figure based on it's current state. However long it takes to evaluate a `plot!()` call, the old version of the figure will stay up until the new one replaces it. This is very convenient for interactive data exploration and for making animations.

plot_REPL

In Pluto, however, users are encouraged to initialize plots in the same  (see the sample notebook [`Plots.jl`](https://github.com/fonsp/Pluto.jl/blob/master/sample/Plots.jl.jl) which says "I strongly recommend that you only use `plot!()` to alter plots you initialised in the same cell."). The problem is that the reactivity introduces some choppiness, which looks like a flash of empty white frames in between frames actually showing the plots.

plot_Pluto

(Here's a snapshot of one of these white frames from my Gif-making app)
Screen Shot 2020-10-15 at 4 30 18 PM

Up to this point, it kind of makes sense because I can imagine there is a lag between when the empty plot in `plot()` is displayed up and when the updated results of `plot!(...)` are displayed. However, the same problem occurs even if you split the initialization of the figure `plot()` and updating `plot!()` into two separate cells.

plot_Pluto_separate

### LINKED ISSUE #911: SVG output from Plots can cause browser tab to consume large amounts of memory
If you try to plot thousands of points, the browser tab can accumulate massive amounts of memory for each plot, using the default SVG output. This can lead to a poor experience for people trying to naively plot data.

One idea could be to try to globally set the default for Plots to PNG instead of SVG. I don't know if that's possible.

Another is that Pluto could warn when evaluating a cell leads to massive allocations on the browser side and could slow down the notebook; and/or notice that specifically a large svg is being returned and Pluto could suggest the format be changed to raster for better performance.

fonsp:
Implemented:

- For `Plots.Plot{Plots.GRBackend}` objects, disable HTML to render as SVG instead: fix #577 using idea 1 in https://github.com/fonsp/Pluto.jl/issues/577#issuecomment-717811253
- For `Plots.Plot{Plots.GRBackend}` objects, when there are more than 8000 data points ([count function](https://github.com/fonsp/Pluto.jl/blob/950df6578ac69f752366fd77cac1310f353fa8ff/src/runner/PlutoRunner.jl#L933-L940)), also disable SVG to render to PNG: fix #911. Yes that number is arbitrary; no it won't be configurable ❤️

Hide whitespace in the diff: https://github.com/fonsp/Pluto.jl/pull/1090/files?diff=split&w=1

# Try it out!

```julia
julia> ]
pkg> activate --temp
pkg> add Pluto#plots-performance
```

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -927,9 +927,6 @@ const integrations = Integration[
     Integration(
         id = Base.PkgId(UUID("91a5bcdd-55d7-5caf-9e0b-520d859cae80"), "Plots"),
         code = quote
-            makeint(x::Integer) = x
-            makeint(x) = parse(Int, x)
-
             approx_size(p::Plots.Plot) = try
                 sum(p.series_list) do series
                     length(series[:y])
@@ -938,10 +935,8 @@ const integrations = Integration[
                 @warn "Failed to guesstimate plot size" exception=(e,catch_backtrace())
                 0
             end
-
             const max_plot_size = 8000
             pluto_showable(::MIME"image/svg+xml", p::Plots.Plot{Plots.GRBackend}) = approx_size(p) <= max_plot_size
-
             pluto_showable(::MIME"text/html", p::Plots.Plot{Plots.GRBackend}) = false
         end,
     )
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -939,7 +939,7 @@ const integrations = Integration[
                 0
             end
 
-            const max_plot_size = 20000
+            const max_plot_size = 8000
             pluto_showable(::MIME"image/svg+xml", p::Plots.Plot{Plots.GRBackend}) = approx_size(p) <= max_plot_size
 
             pluto_showable(::MIME"text/html", p::Plots.Plot{Plots.GRBackend}) = false
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -927,6 +927,21 @@ const integrations = Integration[
     Integration(
         id = Base.PkgId(UUID("91a5bcdd-55d7-5caf-9e0b-520d859cae80"), "Plots"),
         code = quote
+            makeint(x::Integer) = x
+            makeint(x) = parse(Int, x)
+
+            approx_size(p::Plots.Plot) = try
+                sum(p.series_list) do series
+                    length(series[:y])
+                end
+            catch e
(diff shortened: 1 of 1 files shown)


fonsp:
@daschw @jheinen FYI: Pluto will now skip the HTML MIME type for outputs of type `Plots.Plot{Plots.GRBackend}`. I improvised a function to count the 'complexity' of a given plot, and when there are too many datapoints, SVG is also disabled.

@jheinen the interactive GR mode still works, since that is not the Plots.jl type. What I was not sure about is whether you can get the Pluto-interactive GR mode as part of a `Plots.Plot`? If so, is there a way for Pluto to check whether that is being used for a given `Plots.Plot`, so that we can re-enable HTML for that plot?

jheinen:
Right now, you can't use the GR JS backend within Plots.jl. We (only) have an idea, how this could be managed - but there's still a lot of things to do.

fonsp:
Okay! @jheinen please let me know when that happens, I will need to re-enable HTML output for GR JS plots. Important is that Pluto can use a function:
```julia
prefers_html(::Plots.Plot)::Bool
```
to decide whether or not HTML should be enabled. Pluto-managed SVG and PNG output has significant performance benefits over SVG-inside-HTML or PNG-inside-HTML.

pankgeorg:
#1111

The size of the surface plot in the above issue is calculated to be 201, while the SVG size is about 700MB, and it crashes the browser
An idea is to default to PNG for _all_ surfaces or for all plots in general (and allow opt-in SVG using DisplayAs)

Another idea is to default, but respect an explicit Plots.jl option, if it exists.

jheinen:
To which surface plot do you refer? MWE?

pankgeorg:
> To which surface plot do you refer? MWE?
From the issue I've linked: #1111 

```julia
x_range = -10:0.01:10
surface(x_range, x_range, (x, y)->sin(x)+cos(y), fillalpha=0.5, color=:coolwarm)
```

jheinen:
It seems that `Plots` doesn't use `GR`'s software renderer to create 3d surface plots. The SVG size of the following plain `GR` example is about 400K.
```julia
using GR
x_range = -10:0.01:10
surface(x_range, x_range, (x, y)->sin(x)+cos(y))
```
Screen Shot 2021-04-22 at 10 09 24

PR #1093, opened 2021-04-14
Make .pluto.jl the default file extension

fonsp:
`.jl` will still work of course (the file extension does not actually matter)


fonsp:
On MacOS, for a file `hello.pluto.jl`, only `.jl` is detected as the "extension", which means:
- Renaming the file to "world" will give you `world.jl` (`hello.pluto` was replaced)
- Downloading a notebook for the second time will give `hello.pluto (1).jl`

PR #1100, opened 2021-04-16
toplevel QuoteNode fix

### LINKED ISSUE #1039: QuoteNode global variables become Symbols (metaprogramming)
The problem is described by this example (you should add this as a test!):

image

The cause seems to be our function wrapping mechanism (read #720). Specifically, I think the problem lies in the way we turn the function evaluation results into global variables. If I disable this feature by doing a macro call, the problem disappears:

image

wintersm1th:
We must wrap all top-level defined metaprogramming units (Symbol, Expr, QuoteNode) in user code.
Close #1039

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -109,7 +109,7 @@ function register_computer(expr::Expr, key, input_globals::Vector{Symbol}, outpu
 end
 
 quote_if_needed(x) = x
-quote_if_needed(x::Union{Expr,Symbol}) = QuoteNode(x)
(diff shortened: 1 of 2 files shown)


fonsp:
Awesome, thank you for contributing @wintersm1th !

Can you add a test? A good place would be to add cells to this testset: https://github.com/fonsp/Pluto.jl/blob/badc2c92ea25f37fc08abd9c267574531ace575d/test/React.jl#L965

wintersm1th:
Ready

fonsp:
Yayooo thank you!

PR #1118, opened 2021-04-23
State file export endpoint

ctrekker:
Exports a notebook to plutostate MessagePack binary.

Exposes a single endpoint `/statefile?id=<session-id>` which works identically to `/notebookfile` and `/notebookexport`

### DIFF
--- a/src/webserver/Static.jl
+++ b/src/webserver/Static.jl
@@ -228,7 +228,7 @@ function http_router_for(session::ServerSession)
     ) do request::HTTP.Request
         try
             notebook = notebook_from_uri(request)
-            response = HTTP.Response(200, MsgPack.pack(Pluto.notebook_to_js(notebook)))
+            response = HTTP.Response(200, Pluto.pack(Pluto.notebook_to_js(notebook)))
             push!(response.headers, "Content-Type" => "application/octet-stream")
             push!(response.headers, "Content-Disposition" => "inline; filename=\"$(without_pluto_file_extension(basename(notebook.path))).plutostate\"")
             response
--- a/src/webserver/Static.jl
+++ b/src/webserver/Static.jl
@@ -222,6 +222,22 @@ function http_router_for(session::ServerSession)
     end
     HTTP.@register(router, "GET", "/notebookfile", serve_notebookfile)
 
+    serve_statefile = with_authentication(; 
+        required=security.require_secret_for_access || 
+        security.require_secret_for_open_links
+    ) do request::HTTP.Request
+        try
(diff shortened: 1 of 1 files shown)


src/webserver/Static.jl
```diff
@@ -222,6 +222,22 @@ function http_router_for(session::ServerSession)
     end
     HTTP.@register(router, "GET", "/notebookfile", serve_notebookfile)
 
+    serve_statefile = with_authentication(; 
+        required=security.require_secret_for_access || 
+        security.require_secret_for_open_links
+    ) do request::HTTP.Request
+        try
+            notebook = notebook_from_uri(request)
+            response = HTTP.Response(200, MsgPack.pack(Pluto.notebook_to_js(notebook)))
```
fonsp, line 231: This should be `pack` (i.e. `Pluto.pack`) instead of `MsgPack.pack`. Our packing and unpacking is slightly different from vanilla msgpack because:
- we added some extensions and 
- we don't convert `Int64` to `Int32`, `16` or `8` if it is small enough

https://github.com/fonsp/Pluto.jl/blob/main/src/webserver/MsgPack.jl
and the JS side:
https://github.com/fonsp/Pluto.jl/blob/main/frontend/common/MsgPack.js

This is actually something that you should look into for the WYSWYR project! You might want to:
- make sure that the client libraries have matching extensions, or
- use the unextended 'vanilla' MsgPack when sending responses, so that clients can use vanilla MsgPack. I think the first option is better

EDIT: actually my suggestion makes no difference, since `Pluto.pack(args...) = MsgPack.pack(args...)`, and our extensions to MsgPack are global (unfortunately). But still it makes sense 'semantically' to change it to `Pluto.pack`.

fonsp (changes requested):

src/webserver/Static.jl
```diff
@@ -222,6 +222,22 @@ function http_router_for(session::ServerSession)
...
+            push!(response.headers, "Content-Type" => "application/octet-stream")
+            push!(response.headers, "Content-Disposition" => "inline; filename=\"$(without_pluto_file_extension(basename(notebook.path))).plutostate\"")
```
fonsp, line 233: Cool! Can you add a `Content-Disposition` header to our other endpoints as well?

ctrekker: It's already set for `/notebookexport` and `/notebookfile` as well, do you want it set for everything else too?

fonsp: I see, thanks!

PR #1120, opened 2021-04-23
Skip topology update on set bond value

Pangoraw:
The topology will be unchanged because the update will only affect the bind
value. Therefore, we can skip the topology update to speed up the process. :zap:

### DIFF
--- a/src/evaluation/Run.jl
+++ b/src/evaluation/Run.jl
@@ -97,18 +97,17 @@ function run_reactive!(session::ServerSession, notebook::Notebook, old_topology:
 	return new_order
 end
 
-run_reactive_async!(session::ServerSession, notebook::Notebook, to_run::Vector{Cell}; kwargs...) =
-  run_reactive_async!(session, notebook, notebook.topology, notebook.topology, to_run; kwargs...)
+run_reactive_async!(session::ServerSession, notebook::Notebook, to_run::Vector{Cell}; kwargs...) = run_reactive_async!(session, notebook, notebook.topology, notebook.topology, to_run; kwargs...)
 
 function run_reactive_async!(session::ServerSession, notebook::Notebook, old::NotebookTopology, new::NotebookTopology, to_run::Vector{Cell}; run_async::Bool=true, kwargs...)
-  run_task = @async begin
-    run_reactive!(session, notebook, old, new, to_run; kwargs...)
-  end
-  if run_async
-    run_task
-  else
-    fetch(run_task)
-  end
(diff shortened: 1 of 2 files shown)


Pangoraw:
hmm looks like something is failing on the frontend test, i will take a look :eyes:

src/evaluation/Run.jl
```diff
@@ -146,7 +146,7 @@ will_run_code(notebook::Notebook) = notebook.process_status != ProcessStatus.no_
 ###
 
 "Do all the things!"
-function update_save_run!(session::ServerSession, notebook::Notebook, cells::Array{Cell,1}; save::Bool=true, run_async::Bool=false, prerender_text::Bool=false, kwargs...)
+function update_save_run!(session::ServerSession, notebook::Notebook, cells::Array{Cell,1}; save::Bool=true, prerender_text::Bool=false, kwargs...)
```
pankgeorg, line 149: Hey!! I may be missing something, but where is `run_async` now?

Pangoraw: > Hey!! I may be missing something, but where is `run_async` now?

It was part of the `kwargs` array, which are then passed to `maybe_run_async!`. 
I [put it back](https://github.com/fonsp/Pluto.jl/compare/f9bad8ff21d61afe146a4e368a636df7bf85b152..682d085d86648c827622b510881449364d7cc2d8) to be more explicit :smiley:

src/evaluation/Run.jl
```diff
@@ -181,21 +181,26 @@ function update_save_run!(session::ServerSession, notebook::Notebook, cells::Arr
...
-		end
+		maybe_run_async!(session, notebook, old, new, to_run_online; run_async=run_async, kwargs...)
 	end
 end
 
 update_save_run!(session::ServerSession, notebook::Notebook, cell::Cell; kwargs...) = update_save_run!(session, notebook, [cell]; kwargs...)
 update_run!(args...) = update_save_run!(args...; save=false)
 
 
+maybe_run_async!(session::ServerSession, notebook::Notebook, to_run::Vector{Cell}; kwargs...) =
```
fonsp, line 192: Can we change this to `run_maybe_async!`?

fonsp (changes requested):
-

src/evaluation/Run.jl
Pangoraw: :heavy_check_mark: Renamed and moved just after `run_reactive!`

fonsp (approved):

PR #1125, opened 2021-04-26
Download and proxy assets

fonsp:
right now all assets are in the Pluto respository, and i dont like it. Can someone use artifacts to store the assets?

Also:
- this approach relies on my regex skills, which is not reliable
- the same could be achieved by bundling editor.html, which might be the more stable idea
- the first time you launch Pluto, the service worker won't be loaded yet. There are tricks for that

# Changes

https://github.com/fonsp/Pluto.jl/pull/1125/files#diff-2347e4dd7b83b5a18e82e7b51f7de4bf529260b488a9d211b22b49e763d8507f

https://github.com/fonsp/Pluto.jl/pull/1125/files#diff-a03dce75298a3650af3eb42e93a90a05f512eb6e1982c9b3df27b53c0230865f


wulpuqu:
Based on your notebook [download_assets.jl](https://github.com/fonsp/Pluto.jl/blob/bdfa7832a10c64d23bc0146ffb922d1ed28f8123/frontend/download_assets.jl) that scraps dependencies from `editor.html`, I drafted this [script ](https://github.com/wulpuqu/Pluto.jl/blob/41f7aa85deccbc843605df0114aa89b963e35d4a/frontend/artifacts_draft.jl) that initiates an artifact depot with all the scraped assets.
It is still based on the initial regex matchers, and as you said it might not be the most robust way to go. I looked for Web-Scraping packages in Julia and only found Gumbo.jl but it might be overkill for this task (and requires the CGumbo library).

We could hard-code all url dependencies into a (.csv) file that is kept in sync with the actual dependencies of editor.html via some state of the art (bash or some .jl) script not shipped with Pluto.jl.

So my question is: once the artifact directory is instantiated, do you have an idea of how we can pass to the service worker in [sw.js](https://github.com/fonsp/Pluto.jl/blob/bdfa7832a10c64d23bc0146ffb922d1ed28f8123/frontend/sw.js) an artifact directory like "$JULIA_DEPOT_PATH/artifacts/somehash/" ?

I think it needs to be dynamically done during the launch of the webservice in case the user has different julia environnements or depot paths.

fonsp:
@wulpuqu Awesome, thanks! @Pangoraw can you take a look and tell us what you think?

> do you have an idea of how we can pass to the service worker

This is already implemented in this PR, take a look :)

Pangoraw:
Looks great! 

> This is already implemented in this PR, take a look :)

I think the missing part is moving the assets to `/proxied_lib/`, because with artifacts will be located in `.julia/artifacts/<artifacts_hash>/`. Should we create a new repository to host the generated tarballs in Github releases?

wulpuqu:
> This is already implemented in this PR, take a look :)

> I think the missing part is moving the assets to `/proxied_lib/`

I thought of changing the `./proxied_lib/` directory to the actual artifact directory, it avoids copying or moving stuff around. I don't know if it is possible for the service worker to read from a file the hash in order to form the path `.julia/artifacts/<artifacts_hash>/`

fonsp:
We register a new GET endpoint on the pluto server for `/proxied_lib/*` that serves files from the folder with downloaded assets. That way we don't move files, and we don't implement hash logic in the sw.

wulpuqu:
If I understood correctly :slightly_smiling_face: : 
During the installation of Pluto.jl, you want to initiate an artifact directory at `$JULIA_DEPOT_PATH/artifacts/<artifact_hash>/`(either by `Pluto` itself or by a dedicated `Pluto_jll` package dependency) which will contain all the assets and then copy all the assets to `/proxied_lib/*` such that `sw.js` correctly fetch the assets when offline (and cache emptied).

In that case, should be there a `build.jl` script that move the assets in `/proxied_lib/*` ?

Pangoraw:
> In that case, should be there a build.jl script that move the assets in /proxied_lib/* ?

There is no need to move anything. We can make the http server respond to `/proxied_lib/<file>` (the url) with the content of the file located `.julia/artifacts/<hash>/<file>` (the actual file location on disk). Something similar exists for the sample notebooks: https://github.com/fonsp/Pluto.jl/blob/main/src/webserver/Static.jl#L200

uPvtkBlpiucvDUf:
I just came across this PR after filing issue #1315 . Please verify the integrity of .js and .css files whenever you download them from an external source.

You could do this for the stuff in `editor.html` by adding an `integrity` attribute to all `<script>` and `<link>` tags, then scraping not only the URL but also the integrity hashes of each .js and .css file, and then checking this hash whenever it is downloaded in `download_assets.jl` or via the fallback `fetch` in `sw.js`. For the fallback `fetch`, doing this is simple. You can just pass the hash as `fetch(URL, {integrity: hash})`. In `download_assets.jl` I'm not sure what the easiest way would be. But anyway, this does not work so easily for the `import {...} from "https://esm.sh/preact@10.5.13?target=es2020"` in `Preact.original.js`.

Actually, I guess it make sense to use some standard js package manager like `npm` to manage js dependencies inside `frontend/`, no? This should make downloading everything you need simple and you get integrity verification for free. You can then either commit the downloaded files into the Pluto repo or ship them as a separate artifact and only commit the `package-lock.json` (the `Project.toml` analog).

As a "cheap" solution you could also just download the .css and .js files like you do with `download_assets.jl` and distribute them as an artifact. If you correctly [verify the artifact after it is downloaded](https://pkgdocs.julialang.org/v1/api/#Pkg.Artifacts.verify_artifact), this would already be much better than the situation now. (But you should then probably remove the fallback `fetch` in `sw.js` to avoid accidentally downloading things without any integrity check after all.)

dralletje:
Superseded by https://github.com/fonsp/Pluto.jl/pull/1561?

PR #1126, opened 2021-04-26
📈 Pluto object viewer as embeddable element

fonsp:
You can now use Pluto's object viewer inside custom HTML output, making things possible like: interactive table viewer on the left half of a cell, plot on the right. The embedded viewer is available inside notebooks as:
```julia
embed_display(x)
```
and inside packages as:
```julia
if isdefined(Main, :PlutoRunner) && isdefined(Main.PlutoRunner, :embed_display)
	# if this package is used inside Pluto, and Pluto is new enough
	Main.PlutoRunner.embed_display(x)
else
	x
end
```

image

```julia
let
	left = DataFrame(rand(30,3),:auto)
	right = Matrix(left) |> plot
	
	@htl("""

	<div style="display: flex;">
		
	$(embed_display(left))
	$(embed_display(right))
		
	</div>

	""")
end
```

# It's interactive, and you can click on "More"

https://user-images.githubusercontent.com/6933510/116559533-d85d0100-a900-11eb-81cd-83807e566d4d.mov

```julia
let
	left = rand(50)
	right = left |> plot
	
	@htl("""

	<div style="display: flex;">
		
	<div style="max-width: 50%; overflow: auto;">
		$(embed_display(left))
	</div>
	$(embed_display(right))
		
	</div>

	""")
end
```

# You can also use it inside Markdown

image

TODO:
- [x] I had to import preact directly from `esm.sh` to get fields that are hidden in the htm bundle. This probably won't work in China, so either find a way to get it with jsdelivr or #241 . I think the easiest way is to create a new repo containing just https://github.com/fonsp/Pluto.jl/blob/pluto-display-custom-element/frontend/imports/Preact.js and then call `deno bundle` on that, and store the bundle in the same repo
- [x] The preact contexts don't work for custom elements inside preact components, which means that ~bonds don't work~ showing more items doesn't work
- [x] Add helper function to `PlutoRunner`
- [x] Scripts are executed twice?
- [x] How to reuse the preact state?

This is achieved in combination with #1124


fonsp:
This will fix https://github.com/Roger-luo/Configurations.jl/pull/34

fonsp:
@andreypopp I was having a quick look at your fork and this PR might be of interest to you

schneiderfelipe:
@fonsp Do you think this can help improve fonsp/PlutoUI.jl#102?

PR #1153, opened 2021-05-11
Some additions to shortcuts list

i-apellaniz:
Added the shortcut for toggling between code and markdown to the list.
I also completed the sentence "The notebook file saves every time you run" as it appears on [https://github.com/fonsp/Pluto.jl/wiki/💻--UI](https://github.com/fonsp/Pluto.jl/wiki/%F0%9F%92%BB--UI)


fonsp:
Thanks! On MacOS it's also Ctrl+M, because Cmd+M is already bound to minimizing the window :)

Issue #488, opened 2020-09-25
SymPy formulas are not rendered correctly


holomorphism:
After the commit 1aa22b28bd48e1dd4dda3fdc1e7491c1f6cb94dd (#381), SymPy formulas are not rendered correctly.

before:
![ok](https://user-images.githubusercontent.com/3076585/94217599-60b79580-ff1d-11ea-9f23-1170b4704cf2.png)

after:
![ng](https://user-images.githubusercontent.com/3076585/94217605-64e3b300-ff1d-11ea-95b9-3bacf7306a51.png)

Thank you for the amazing software!

fonsp:
Can you post the result of:
```julia
repr(MIME"text/latex"(), Sym("x"))
```

holomorphism:
Sure.
![ng1](https://user-images.githubusercontent.com/3076585/94279474-1e2aa300-ff87-11ea-83b9-32e7b99304df.png)

stillyslalom:
I think the underlying issue is MathJax's inability to mix display-mode equations and text. See [this](https://docs.mathjax.org/en/v2.7-latest/tex.html#differences), 

> First and foremost, the TeX input processor implements only the math-mode macros of TeX and LaTeX, not the text-mode macros. MathJax expects that you will use standard HTML tags to handle formatting the text of your page; it only handles the mathematics.

Given that, probably best to roll back #381 and provide good documentation on how to properly mix TeX and text.
![image](https://user-images.githubusercontent.com/4324760/95646279-6a4c0c00-0a8c-11eb-99aa-87d7bb456e0f.png)

fonsp:
MathJax is able to mix text and math, as your screenshot demonstrates

stillyslalom:
Yes, but the amsmath `equation*` environment (used by SymPy.jl for LaTeX repr) can't be wrapped with `\text`. It can be worked around by overriding https://github.com/JuliaPy/SymPy.jl/blob/b9028048dc2d8e8681ef3837c8232b8b262d0f3d/src/types.jl#L76 to change the output style from `equation*` to `inline`.
![image](https://user-images.githubusercontent.com/4324760/95659249-5e406880-0ae5-11eb-85b2-0adedec0ffde.png)

fonsp:
> can't be wrapped
Are you referring to LaTeX spec or is this a limitation of MathJax?

stillyslalom:
[...]

PR #1164, opened 2021-05-16
No more wrapping `text/latex` with `\text{}`

zmoon:
Essentially reverts #381, which definitely seems to have caused more issues (breaking display of complex equation blocks for various packages, etc.) than it solved (fixing display for `LaTeXString`s with mixed text/math). Much discussion in #488.

I opened an issue at LaTeXStrings (https://github.com/stevengj/LaTeXStrings.jl/issues/50). Else, users can override the display for `LaTeXStrings` if they want like suggested [here](https://github.com/fonsp/Pluto.jl/issues/488#issuecomment-841655163)/[here](https://github.com/fonsp/Pluto.jl/issues/488#issuecomment-841687837), e.g. something like
```julia
Base.show(io::IO, ::MIME"text/html", s::LaTeXString) = 
    print(io, Markdown.LaTeX("\\text{$(s.s)}"))
```
seems to work ok. Or
```julia
function Base.show(io::IO, ::MIME"text/html", s::LaTeXString)
    texed = repr(MIME"text/latex"(), s)
    html(io, Markdown.LaTeX("\\text{$texed}"))
end
```


fonsp:
Thanks! Is this PR still a draft? Can you add a TODO list?

zmoon:
Hi @fonsp. I guess I wanted to wait a bit for LaTeXStrings to see if they respond. Also wondering, is it in the scope of Pluto to include this special case for LaTeXStrings within the Pluto codebase? If no, then I wouldn't necessarily add anything else to this PR unless you have something in mind.

Not in the scope of this PR, but what about some parsing of `text/latex` in order to split text and math? Jupyter seems to do this. For example, in JupyterLab:
image
:point_up: The text part is rendered as normal HTML text, only the math part is passed to MathJax.

fonsp:
> is it in the scope of Pluto to include this special case for LaTeXStrings within the Pluto codebase?

Great question! Indeed, at the start of the discussion about LaTeXStrings, I would have said no, but recently we were able to easily add "integration" support code like the one you suggestion, with a system similar to Requires.jl. Let's wait to see what the LaTeXStrings friends decide to do, and otherwise we can add it.

Thanks for opening the issue at LaTeXStrings!

fonsp:
Thanks! Let's continue waiting for LaTeXString (see my comment on your issue) and implement this in the meantime

PR #1170, opened 2021-05-17
Add tests for authentication

rikhuijzer:
In https://github.com/fonsp/Pluto.jl/pull/529, authentication was added. This PR adds some basic tests.

The benefit of these tests is that they are much easier to verify than the implementation in `src/webserver/Static.jl`.


test/Configuration.jl
```diff
@@ -38,4 +40,42 @@ end
...
+        "notebookfile",
+        "notebookexport",
+        "open",
+        "sample",
+        "sample/Interactivity.jl",
+        "statefile",
+        "foo"
+    ]
+    for suffix in routes
+        @test access_denied(url("edit"))
```
Pangoraw, line 77: I think you meant to use `suffix` here :wink:

rikhuijzer: Thanks! Well spotted

rikhuijzer:
The failing test is due to a front end test being flaky. The back end tests all pass.

fonsp:
Nice! Can you try to shut down the HTTP server afterwards?

test/Configuration.jl
```diff
@@ -38,4 +40,40 @@ end
     ["--compile=min", "--project=test", "--startup-file=no", "--history-file=no"]
 end
 
+@testset "authentication" begin
```
fonsp, line 43: Just to make sure that the tests are really testing authentication, and not a bad server/tests, could you change this test line to:

```julia
@testset "Authentication: $(auth ? "on" : "off")" for auth in [true, false]
```

and then enable/disable authentication accordingly, and use two different ports.

fonsp: Alternatively, you could leave authentication on for all tests, and instead add a second test where the authentication secret is provided in the URL, and check that this gives you access.

fonsp:
My first reaction is to try removing the trailing slashes, i.e. `host/new?secret=asdf` instead of `host/new/?secret=asdf`, etc. Any remaining failing URLs are probably because they only work if you provide an existing notebook ID. e.g. the `notebookfile` endpoint returns the file of the given notebook. 

Feel free to take a look, but I can also finish this myself! I think you want to add a small notebook to the session before testing the routes, see our other tests on how to do that.

fonsp:
The sample URL should be URL-encoded. Open the Pluto main menu, go to samples and right click on one of the links to see what it looks like. (Same for the `/edit` url and such.) Note that you won't see the secret there because normal Pluto usage keeps it in cookies, instead of in each URL.

rikhuijzer:
@fonsp, thanks for the feedback. I've applied most of it.

> I think you want to add a small notebook to the session before testing the routes, see our other tests on how to do that.

Could you pick this one up? I didn't figure it out and it's then probably quicker if you pick it up in a commit then we talking about it some more :)

fonsp:
@rikhuijzer dankjewel!

PR #1184, opened 2021-05-20
Draft: Front-end tests for JavaScript API

pankgeorg:
This PR adds some tests for the JS functionality as defined by the JavaScript sample notebook.

- [X] If you return an HTML node, it will be displayed.
- [X] The observablehq/stdlib library is pre-imported, you can use DOM, html, Promises, etc.
- [X] When a cell re-runs reactively, this will be set to the previous output (with caveat, see the later section)
- [ ] The variable invalidation is a Promise that will get resolved when the cell output is changed or removed. You can use this to remove event listeners, for example.

Furthermore, now tests close the notebook they use, to save resources.

Some gotchas about the JS API:

- The bonds updates are throttled. That means that when you have two bond updates to be done very close to each other, they won't both trigger a dependent Cell reflow
- Puppeteer is fast. It's really fast because it's not bounded by the reflow/sleep/whatever cycles of the browser. Remember to wait for th right thing - in this PR it's the cell that shows the updated value to show the updated value, before re-triggering a bond update event.
- After every cell rerun, the cell output is _removed from the browser DOM_. Not deleted (we test that it's the same node reference). It's just _not in the DOM_, for a little while, even if it's the same unchanged node. After a bit it gets back. That's why (in conjunction with the puppeteer being _fast_, it's not enough to wait for `waitForNoUpdateOngoing`. After the update is done, we check that the 'affected' output nodes got remounted! (highly likely, when the test was class, it was so slow it was sufficient - as it was apparently triggering some nasty reflows/repaints)


fonsp:
Nice! You can also merge this PR right now and add more things in later commits

PR #1191, opened 2021-05-24
🥐 Bake jl file into static export, binder button on exported HTML files

fonsp:
A generated HTML file now contains the contents of the original notebook file, as a base64-encoded data URL. This means that HTML files generated from the export menu now have a binder button and a "download notebook" button built-in!

The goal was to also add the ability to open HTML files from the main menu, but we still need to think about where to store the jl file after it is loaded. What about storing it back into the HTML file, skipping a .jl file entirely? 👀 With embedded state? 👀👀 Would be neat!

Example of a new export file:

https://gist.github.com/fonsp/28b2480990372f001d356aa00644119b

View it in your browser: https://htmlview.glitch.me/?https://gist.github.com/fonsp/28b2480990372f001d356aa00644119b

### DIFF
--- a/frontend/binder.css
+++ b/frontend/binder.css
@@ -165,6 +165,7 @@ body.wiggle_binder #binder_launch_help {
     background-image: url(https://cdn.jsdelivr.net/gh/ionic-team/ionicons@5.0.0/src/svg/close-outline.svg);
 }
 
+.download_div,
 .copy_div {
     padding: 4px 8px;
     display: flex;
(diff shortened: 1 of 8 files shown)


fonsp:
@ctrekker Instead of extracting a jl file from the statefile, I decided to use the existing config variable `notebookfile` which normally contains a URL pointing to the notebook file, to store it as a data URL. No conversion needed! Take a look at the sample html to see what I mean.

PR #1208, opened 2021-05-31
Remove Ctrl-d shortcut from pluto for mac os users

### LINKED ISSUE #1200: Ctrl + d  keyboard shortcut interferes with macOS functionality
I prefer to use Ctrl-P/N/F/B/D/K/Y/D/A/E on macOS to control the cursor on text. On Pluto notebook, It seems Ctrl-P/N/F/B/K/A/E works as described in `Document shortcuts` section of Mac keyboard shortcuts: https://support.apple.com/en-us/HT201236.

Unfortunately `Ctrl-D` does not work as `Control-D: Delete the character to the right of the insertion point.`. 

Since line in `KeyboardShortcuts.js` tries to replace `Ctrl-blah-blah` in with `Cmd-blah-blah` in `keymap`, one may think this is fine. 

https://github.com/fonsp/Pluto.jl/blob/cd7c123e5dc9cb5c2d28e119aa66b83cdb0371f7/frontend/common/KeyboardShortcuts.js#L14

However, this lines assigns `Ctrl-D` shortcut to do something (selecting some words ???)

https://github.com/fonsp/Pluto.jl/blob/28615ad0f053f505bc478f814bb69c4ccb522e8d/frontend/components/CellInput.js#L151-L162

If I remove this lines above, I can use Ctrl-D to delete the character to the right, which is what I expect on macOS.

Note that if one would like to use `Ctrl-D` functionality that Pluto.jl provides on macOS, one only have to do use `Cmd+D` (due to this line https://github.com/fonsp/Pluto.jl/blob/cd7c123e5dc9cb5c2d28e119aa66b83cdb0371f7/frontend/common/KeyboardShortcuts.js#L14) .

Do we need insert conditional branch something like ?

```javascript
if (!is_mac_keyboard){
  keys["Ctrl-D"] = () => {
  // do something
  }
}
```

Any thoughts about this issue ?

terasakisatoshi:
This PR resolves #1200 . It would make texting more comfortable. for macOS users.

### DIFF
--- a/frontend/common/KeyboardShortcuts.js
+++ b/frontend/common/KeyboardShortcuts.js
@@ -12,6 +12,10 @@ export let map_cmd_to_ctrl_on_mac = (keymap) => {
     let keymap_with_cmd = { ...keymap }
     for (let [key, handler] of Object.entries(keymap)) {
         keymap_with_cmd[key.replace(/Ctrl/g, "Cmd")] = handler
+        // remove Ctrl-D from Pluto.jl keybind for MacOS
+        if (key == "Ctrl-D") {
+            delete keymap_withcmd[key]
+        }
     }
     return keymap_with_cmd
 }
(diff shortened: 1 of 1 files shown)


terasakisatoshi:
thank you 🙏

Issue #421, opened 2020-09-13
Idea for richer notebook format


j-fu:
Hi, just a thought here for a possible modificatioon of the notebook format --  in case you are considering to extend the current format with more stored information:

I can imagine to use a special comment prefix and  some standard way to store richer config information - Julia or  (preferably) toml which you can  parse to dicts  using TOML.jl  E.g. like this:
````
#═╡ [PlutoNotebook]
#═╡ format_version = "2.0"
#═╡ pluto_version  = "1.20"



#═╡ [PlutoCell]
#═╡ uuid = "7f904dfe-b85e-4ff6-b463-dae2292396a8"
#═╡ hidden = false
#═╡ show_output=true
#═╡ results_on_top=false
hack hack hack

#═╡ [CellOrder]
#═╡ order = [ 
#═╡	      "uuid1",
#═╡	      "uuid2"
#═╡	   ]
````
(not sure if the Julia toml parser supports arrays).

This would give you the possibility to add richer meta information to the notebook in an extendable way.

Edit: this could be consistent also with some of the ideas in e.g. #142 .:
````
#═╡ [PlutoProject] 
#═╡   project.toml stuff here
#═╡
````

Edit: I think an API for reading and writing notebooks could go into an extra package, thus allowing tools
like Literate to write notebooks through this API and encourage the development of a large tooling ecosystem.

fonsp:
Thanks! #311

(I'll point out that you can use `;` as the last character of a cell to toggle showing output.)

fonsp:
I've put this on the notion page:
https://www.notion.so/Flexible-notebook-format-61749c09aa1b450fa79f025d725bef4c

j-fu:
On Mon, 14 Sep 2020 22:56:11 +0200,
Fons van der Plas wrote:
> 
> (I'll point out that you can use ; as the last character of a cell to toggle showing output.)

... I had in mind the @with_output stuff, also with regard to
https://julialang.zulipchat.com/#narrow/stream/243342-pluto.2Ejl/topic/print.20inside.20in.20the.20notebook.3F/near/209725305



Best regards
Jürgen

-- 
Jürgen Fuhrmann         Numerical Mathematics and Scientific Computing
            Weierstrass Institute for Applied Analysis and Stochastics
Mohrenstr.39 10117 Berlin   phone:+49 30 20372560  fax:+49 30 20372317
[...]

Issue #132, opened 2020-05-31
Can't import a .jl file


dpsanders:
I tried to open a standard .jl file in Pluto (from the Pluto startup page) and got

```
Failed to load notebook:

MethodError(union, (), 0x0000000000006c87)

<a href="/">Go back</a>
```

It would be great to be able to import a normal `.jl` and have it Plutified. This should e.g. take blocks of code and make them a single cell.

fonsp:
I like your optimism in using the Open button and seeing what happens!

I guess that this is easy to add in, but hard to do right. With that I mean that there is lots of new UI needed to explain to the user: what they are doing, what the result will be (no two-way conversion), what went wrong (if the import fails) and where they want to save the notebook after importing (overwrite?).

Still, this might be worthwhile to add in, since it makes it easier to get started with Pluto.

fonsp:
To anyone taking this on: see also #14, maybe this can be combined with Weave.jl and pandoc and Literate.jl and friends

j-fu:
I think a possible way would be through Literate.jl, which has an option to output to Juypter, that means there is already a way to  define markdown and Julia cells. By no means I think that a workflow .jl -> pluto should pass through .ipynb, though. 

But may be it is even simpler: IMHO it would be sufficient to have syntax to mark cell boundaries, e.g. by `# ===` which is recognized by Pluto on the first load and converted into the hash voodoo separators.  Then we would have the notion of an "empty", "untouched" "proto-" notebook or however we call it and a notebook with recognized cells and  manifest cell order.
This could even go into a Pluto.plutoify() method outside of the browser which takes in a proto notebook and spits out a real one (and complains about errors/ambiguities), very much like Literate itself works. This then could even be used for having a cli tool for this purpose.

fonsp:
Support for writing notebook files by hand is not of this project, and making conversion part of Pluto is requires some carefully designed UI (https://github.com/fonsp/Pluto.jl/issues/132#issuecomment-636748506). But just like Jupyter conversion (https://observablehq.com/@olivier_plas/pluto-jl-jupyter-conversion and https://github.com/vdayanand/Jupyter2Pluto.jl), this can be an external tool, which is very simple to write! Maybe someone could try making it themselves?

If you run `Meta.parse(code_string, 1)`, then it parses as much as it can into a single expression, and it returns **the index of the next expression**:
```julia
julia> code="s = 1\nr=2"
"s = 1\nr=2"

[...]

PR #1209, opened 2021-05-31
saving / loading disable status of cells

lungben:
implements #1205 

- [x] Saving and loading directly and indirectly disabled cells
- [x] Make sure that indirectly disabled cells are updated before saving the notebook


pankgeorg:
Is hiding the `#`s in the plan for the frontend?

lungben:
I have not changed the way how deactivated cells look in the frontend.
The addition is that the cell deactivation information is now stored persistently in the notebook.jl file as commenting out the cell code with a special syntax. 
Furthermore, the downstream dependencies are commented out, too (with a different marker) so that the notebook.jl file still runs stand-alone.

ederag:
Upon reloading, the `#` are not visible in the frontend, good.

For now, after disable, quit and reopen, it seems to work fine, congratulations !
And re-enabling brings back exactly the previous file, perfect.

I really like how the manually disabled cells are darker than their dependencies.

A detail: in the file, the comments are dim, which is good (as in the frontend).
But some characters are not displayed here (kate editor, hack font):
Screenshot_20210602_151801
This might be a kate deficiency; "liberation mono" and "droid" have the same issue.
Another proposition, using the double line symbols that are used in other places by Pluto:
```
# ╔═╡ d5c3ed28-8854-4381-bcf3-9f764af94643
#=╠═ disabled
sol = solve(prob, Tsit5(), reltol=1e-8, abstol=1e-8)
  ╠═ disabled =#
```
I left the last tee symbol (and not a corner) just in case it would be necessary to append other pluto-specific comments later.

lungben:
🛑 is a unicode symbol, it seems that your editor or font does not support it (it works for me in VSCode). 
However, it is trivial to change the symbol - it this version working better for you?

src/evaluation/Run.jl
```diff
@@ -177,6 +190,8 @@ function update_save_run!(session::ServerSession, notebook::Notebook, cells::Arr
 
 	update_dependency_cache!(notebook)
 
+	indirectly_deactivated = disable_dependent_cells!(notebook, new)
+
 	session.options.server.disable_writing_notebook_files || save_notebook(notebook)
```
lungben, line 195: I have to determine the indirectly deactivated cells before saving (in the original implementation it was done afterwards).
In order to avoid double work, `run_reactive!` gets the list of deactivated cells as (optional) parameter.

src/evaluation/Run.jl
```diff
@@ -35,13 +36,8 @@ function run_reactive!(session::ServerSession, notebook::Notebook, old_topology:
...
 
-	# find (indirectly) deactivated cells and update their status
-	deactivated = filter(c -> c.running_disabled, notebook.cells)
-	indirectly_deactivated = collect(topological_order(notebook, new_topology, deactivated))
-	for cell in indirectly_deactivated
-		cell.running = false
-		cell.queued = false
-		cell.depends_on_disabled_cells = true
+	if indirectly_deactivated === nothing
+		indirectly_deactivated = disable_dependent_cells!(notebook, new_topology)
```
lungben, line 40: `run_reactive!` is called both from `update_save_run` where the indirectly deativated cells are calculated before (in order to save this information to the notebook.jl file) and from `Dynamic.jl`, where this has not been done yet. This check keeps generality of this function while avoiding double calculations.

ederag:
> However, it is trivial to change the symbol - it this version working better for you?

With a94685f172fdb80e76e87b9c1cbe899f148ad39d, the symbols are visible, thanks !
```julia
# ╔═╡ d5c3ed28-8854-4381-bcf3-9f764af94643
#= ╠═ disabled ═╣
sol = solve(prob, Tsit5(), reltol=1e-8, abstol=1e-8)
╠═ disabled ═╣ =#
```
Just in case it would be an overlook, I find it a bit cleaner with the tee aligned vertically
(with `╠═╡` and no closing symbol, for consistency with other Pluto comments),
```julia
# ╔═╡ d5c3ed28-8854-4381-bcf3-9f764af94643
#=╠═╡ disabled
sol = solve(prob, Tsit5(), reltol=1e-8, abstol=1e-8)
  ╠═╡ disabled =#
```

Otherwise, disable/re-enable dependent cells works as expected.

Good: if someone mess with the comments in the file, by inadvertence,
in Pluto the code just appears inside an enabled cell, together with the messed-up comments, no loss of information.

Up to now it looks robust.

lungben:
Thanks, is corrected.

ederag:
In the file, there are two empty lines after a disabled cell, but only one for enabled cells.
Removing one of these empty lines in the file, and reopening in Pluto leads to syntax error: 
the last character of the valid code has been lost.

Current PR works, but it might be nice to have a consistent layout,
since editing is sometimes used as a workaround, and there are some Pluto files producers in the works (e.g. #132) ?

While at it: why not include the trailing `\n` in `_running_disabled_prefix` ?

lungben:
Good catch, is corrected!

lungben:
@fonsp from my side this PR is ready for merging now.
Or do you have any further suggestions?

ederag:
The format seems good to me, thanks !

Another detail, maybe not related to this PR:
the file is saved just after loading.
If there is something Pluto does not expect in the file (e.g. two blank lines between cells instead of one),
the file is overwritten and corrupted.
I would expect the file to be overwritten only upon a manual change on the Pluto side.

fonsp:
Thanks! Since this is a (non-breaking) file change, I would like to wait a bit before merging.

One thought is that we might need to better communicate "disabled cells" when opening a notebook, since the feature will now be exposed to people who have never used Pluto before and opened someone else's file.

lungben:
Maybe show a popup window, similar to the undo window for cell deletion, when at least one cell is deactivated?
This popup window should give the option to reactivate all cells.

A bit off-topic, but a keyboard shortcut for toggeling the current cell between enabled and disabled would be helpful, too.

lungben:
Another small point: currently, if a cell with a higher `cell_precedence_heuristic` (e.g. a cell containing a `using` statement) is deactivated, dependent cells are not indirectly deactivated because the dependencies are not known.
One way to deal with it would be to deactivate all cells with higher precedence if a cell with non-default (highest) precedence is deactivated.

ederag:
That's a valid concern, but I'd rather disable *only* cells that are *known for sure* to error due to the disable, 
and wouldn't risk to deactivate cells that are actually independent on this "`using`" cell.
The cells really depending on the "`using`" cell would just error with a meaningful message, so it looks good enough for now ?

lungben:
Yes, deactivating based on preceedance can have problematic side effects. So I agree to keep the logic like it is now and only disable known dependencies.

src/notebook/Notebook.jl
```diff
@@ -211,10 +229,21 @@ function load_notebook_nobackup(io, path)::Notebook
...
+            # remove the disabled on startup comments for further processing in Julia
+            code_normalised = replace(replace(code_normalised, _running_disabled_prefix => ""), _running_disabled_suffix => "")
+            code_normalised = replace(replace(code_normalised, _depends_on_disabled_cells_prefix => ""), _depends_on_disabled_cells_suffix => "")
+
             # remove the cell suffix
             code = code_normalised[1:prevind(code_normalised, end, length(_cell_suffix))]
 
             read_cell = Cell(cell_id, code)
+            read_cell.running_disabled = running_disabled
+            read_cell.depends_on_disabled_cells = depends_on_disabled_cells || running_disabled
```
fonsp, line 246: We don't need to _read_ this behaviour, right? We only need to write it.

src/evaluation/Run.jl
```diff
@@ -108,6 +104,23 @@ function run_reactive!(session::ServerSession, notebook::Notebook, old_topology:
...
+	deactivated = filter(c -> c.running_disabled, notebook.cells)
+	indirectly_deactivated = collect(topological_order(notebook, topology, deactivated))
+	for cell in indirectly_deactivated
+		cell.running = false
+		cell.queued = false
+		cell.depends_on_disabled_cells = true
+	end
+	return indirectly_deactivated
+end
+
```
fonsp, line 123: Why are there changes to Run.jl? We already compute the indirectly disabled cells, right?

src/notebook/Notebook.jl
lungben: When I remove the line
```
read_cell.depends_on_disabled_cells = depends_on_disabled_cells || running_disabled
```
I get the error message in the console:
```
┌ Warning: Old Pluto notebook might not have loaded correctly. Backup saved to: 
│   backup_path = "C:\\Users\\benlu\\.julia\\pluto_notebooks\\cell_deactivation_demo backup 2.jl"
└ @ Pluto c:\Users\benlu\.julia\dev\Pluto\src\notebook\Notebook.jl:339
```

src/evaluation/Run.jl
lungben: I have to determine the indirectly deactivated cells before saving (in the original implementation it was done afterwards).
In order to avoid double work, run_reactive! gets the list of deactivated cells as (optional) parameter.

See https://github.com/fonsp/Pluto.jl/pull/1209#discussion_r644156923 and https://github.com/fonsp/Pluto.jl/pull/1209#discussion_r644159254

lungben:
> Maybe show a popup window, similar to the undo window for cell deletion, when at least one cell is deactivated?
> This popup window should give the option to reactivate all cells.
> 
> A bit off-topic, but a keyboard shortcut for toggeling the current cell between enabled and disabled would be helpful, too.

For these parts I would need help from a JS expert.

ederag:
> > Maybe show a popup window, similar to the undo window for cell deletion, when at least one cell is deactivated?
> > This popup window should give the option to reactivate all cells.

Those popups are perhaps useful for deletion that otherwise would be irreversible,
but I find them a bit distracting. And disable is easily reversible.
Not strong opinion here, just unsure it is worth the time.

> A bit off-topic, but a keyboard shortcut for toggeling the current cell between enabled and disabled would be helpful, too.
> Should we put these points in this PR or rather in a separate one?

Would be nice, but quite separate indeed.

src/notebook/Notebook.jl
fonsp: 👍

fonsp:
I think we should:
- move the depends_on_disabled_cells field from the Cell struct to the CellDependencies struct (update other code to match)
- move the calculation of that field to update_dependency_cache!
- leave the Run.jl file mostly unchanged, computing the field twice is not a problem. Keeping Run.jl simple is more important
- add a test for indirect deps using the jl_file_is_runnable function in test/helpers.jl

lungben:
Points 1-3 done, this point is still tbd:

> * add a test for indirect deps using the jl_file_is_runnable function in test/helpers.jl

pankgeorg:
@pankgeorg

- [ ] Keyboard shortcut to enable/disable cells
- [ ] Some GUI to enabled all disabled cells at one

ederag:
@pankgeorg
do you mean they prevent merging this PR (as the checkboxes imply),
or are they (good) ideas that came to mind while reviewing ?

pankgeorg:
it's very nice to have @ederag. let's discuss it next Thursday on the pluto devs call !! - [see here](https://julialang.zulipchat.com/#narrow/stream/257776-Pluto.2Ejl-development/topic/Weekly.20Pluto.20meetups)!

dralletje:
Heyyy, I have another request here: Can we wait with this until we have a more generic way to store extra metadata with cells?

OR???

Let a disabled cell be a cell that is commented out fully?
Just like in the Chrome style explorer, when you disabled a style with the checkbox, it "just" comments it (if you comment a style out, it will show up in the style explorer, but unchecked).
This way we also have a shortcut for disabling cells directly, `ctrl + /`!

ederag:
> Let a disabled cell be a cell that is commented out fully?

That is already the case with this PR, in the .jl file
```julia
# ╔═╡ d5c3ed28-8854-4381-bcf3-9f764af94643
#=╠═╡ disabled
sol = solve(prob, Tsit5(), reltol=1e-8, abstol=1e-8)
  ╠═╡ disabled =#
```

But I find it good that it is only dimmed out in the Pluto display, 
and not simply commented out,
since the **last output is shown** (dimmed out also).
It feels more consistent, preserves **more information**,
and makes it **easier to locate the right cell** to reenable, later.

Conversely, if you merely comment out the cell content, as I understand your suggestion, then
1) the **syntax highlighting** is lost
2) the output is lost
3) there is no way for Pluto to tell the difference between a disabled cell (whose children are disabled)
and a commented out cell (whose children should error), so the metadata you talk about is needed anyway.

Reading again, it looks like you figured that out in the end, so please take this only as a clarification for others.

dralletje:
> That is already the case with this PR, in the .jl file

I'm so not used to seeing `#=`, hahaha.
I was thinking prepending all the lines with `#` but this works just as well.
Isn't prepending every line with `#` a bit more error-prone? Now we also need to check cells for not having a dangling `=#` D:

> But I find it good that it is only dimmed out in the Pluto display,
> and not simply commented out,

Yeah!! Visually, very much!! I'm thinking more the other way, so when you select a whole cell and comment it out, it becomes disabled and dimmed. I like the dimmed version way better than this huge pink block of comments :P

> Reading again, it looks like you figured that out in the end, so please take this only as a clarification for others.

Yeah... 😅 Thanks!
So what about commenting, but storing the `disabled = true` (and `disabled = dependent` or something) using https://github.com/fonsp/Pluto.jl/issues/1482 (If I get this is in before the end of the week 😁)?
So it'd be like

```Julia
# ╔═╡ d5c3ed28-8854-4381-bcf3-9f764af94643
# ╠═╡ disabled = true
#=
sol = solve(prob, Tsit5(), reltol=1e-8, abstol=1e-8)
=#
```

or 

```Julia
# ╔═╡ d5c3ed28-8854-4381-bcf3-9f764af94643
# ╠═╡ disabled = true
# sol = solve(prob, Tsit5(), reltol=1e-8, abstol=1e-8)
```

lungben:
Another cool emergent feature of cell disabling is that you can use variables defined in a disabled cell in a new cell without an error, the new cell gets automatically disabled.

I like the idea of using TOML meta-data!

Regarding commenting out by line (with `#`) or block-wise (with `#=`) I do not have a strong preference. I originally used block-wise because it was easier this way to distinguish between "normal" comments and disabled cells (using the metadata after the block comment sign), but with the TOML meta-data this would not be an issue anymore.

postylem:
Where does this PR stand in terms of getting merged?  I'm working on using Pluto to create assignments and being able to disable cells which the students must modify before they will work, or cells whih will run for an extremely long time is a really nice feature I was hoping to use.  I didn't realize the disabledness state doesn't currently get saved.

lungben:
The issue is that this PR would change/enhance the Pluto file format, which is not reversible (however, the change is backward-compatible).

Thus, if #1482 should be introduced, it is better to do it before this PR and store the cell disabling status in that new metadata.
@dralletje what is the status of that PR?

ederag:
_Since this is stalled anyway, here is an answer to https://github.com/fonsp/Pluto.jl/pull/1209#issuecomment-924817755:_

Prepending `# ` on every lines is simpler,
but leads to much larger diffs.
With block comments, cell disable is a two lines change
with clear purpose.

With the nice toml idea #421 or #1482 it would be tempting to add a `# ╟─ disabled = true` line,
but that would either 
- [_prepend strategy_] save a single line in the source, but add a lot of `+ # ...` and `- # ...` lines in the diffs, or
- [_block quote strategy_] add another line in the source
(no big deal, but labelled block quote keeps explanation closer to the comments).

Yet it's true that executing a `.jl` with dangling `#=` would syntax error as
`incomplete: unterminated multi-line comment #= ... =#`,
and a `.jl` with a dangling `=#` would syntax error as
`unexpected "="`.
Neither affect Pluto, and both are easy to fix in the source if needed, so it does not look so bad to me ?

lungben:
How to proceed here:

1. First implement the cell metadata functionality !1482 (@fonsp already started with it, I can continue)
2. Then reimplement this PR based on the new Pluto main and !1482 (there have been so many changes in Pluto that it is easier to re-implement this PR than merging main into this PR).

lungben:
This functionality is implemented in #1895, closing this PR now.

PR #1220, opened 2021-06-05
Better detect function assignments to disable function wrapping

### LINKED ISSUE #1199: function parsing bug on Pluto v0.14.7
for some reason, this function written in short-form is not understood as a function:
```julia
named_tuple(obj::T) where {T} = NamedTuple{fieldnames(T), Tuple{fieldtypes(T)...}}(ntuple(i -> getfield(obj, i), fieldcount(T)))
```
Screenshot 2021-05-27 at 23 46 51

In long form, though, it's understood fine:
```julia
function named_tuple(obj::T) where {T}
	NamedTuple{fieldnames(T), Tuple{fieldtypes(T)...}}(ntuple(i -> getfield(obj, i), fieldcount(T)))
end
```

Screenshot 2021-05-27 at 23 48 18

Pluto v0.14.7

Pangoraw:
This is a small refactor to build the `ExprAnalysisCache` and the `ReactiveNode` together as part of a topology update. This allows using the Reactive node data to enable/disable function wrapping without re-writing all the function transforms rules from `explore!` in `can_be_function_wrapped`. The Reactive nodes are also not re-computed if the code of the cell does not change, avoiding extra work.

Fixes #1199

### DIFF
--- a/src/analysis/TopologyUpdate.jl
+++ b/src/analysis/TopologyUpdate.jl
@@ -6,22 +6,32 @@ import REPL: ends_with_semicolon
 function updated_topology(old_topology::NotebookTopology, notebook::Notebook, cells)
 	
 	updated_codes = Dict{Cell,ExprAnalysisCache}()
+	updated_nodes = Dict{Cell,ReactiveNode}()
 	for cell in cells
 		if !(
 			haskey(old_topology.codes, cell) && 
 			old_topology.codes[cell].code === cell.code
 		)
-			updated_codes[cell] = ExprAnalysisCache(notebook, cell)
+			parsedcode = parse_custom(notebook, cell)
+			new_node = parsedcode |>
+				ExpressionExplorer.try_compute_symbolreferences |>
+				ReactiveNode
+			using_imports = ExpressionExplorer.compute_usings_imports(parsedcode)
+			new_code = ExprAnalysisCache(
+				code=cell.code,
+				parsedcode=parsedcode,
+				module_usings_imports=using_imports,
+				function_wrapped=isempty(filter(funcname -> !startswith(string(funcname), "anon"), new_node.funcdefs_without_signatures)) &&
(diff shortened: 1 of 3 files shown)


fonsp:
Hmmmm does this mean that cells with `do` or `->` can no longer be function wrapped? Is there a way to fix the issue without making that restriction?

src/analysis/TopologyUpdate.jl
```diff
@@ -6,22 +6,32 @@ import REPL: ends_with_semicolon
...
+				ExpressionExplorer.try_compute_symbolreferences |>
+				ReactiveNode
+			using_imports = ExpressionExplorer.compute_usings_imports(parsedcode)
+			new_code = ExprAnalysisCache(
+				code=cell.code,
+				parsedcode=parsedcode,
+				module_usings_imports=using_imports,
+				function_wrapped=isempty(filter(funcname -> !startswith(string(funcname), "anon"), new_node.funcdefs_without_signatures)) &&
+					:eval ∉ new_node.references && :include ∉ new_node.references &&
+					isempty(using_imports.usings) && isempty(using_imports.imports) &&
```
fonsp, line 28: the eval, include, using, import checks are already done by `can_be_function_wrapped`, right? I think the `can_be_function_wrapped` is super fast, so the optimization is probably not necessary.

fonsp:
Can you add the issue as a test?

Pangoraw:
> Hmmmm does this mean that cells with `do` or `->` can no longer be function wrapped? Is there a way to fix the issue without making that restriction?

That's why I did the `!startswith("anon")` check but it was not the most elegant way. I moved [the condition used in `explore!`](https://github.com/fonsp/Pluto.jl/pull/1220/files#diff-04c7252262a0e6de2db2cded7aba9b1130c1e5ba6adff25bf8ec3a34e36a4788R288-R295) to its own function and used that in `can_be_function_wrapped` instead.

Issue #1204, opened 2021-05-31
direction of expansion of Pluto cells containing a <details> html elements


disberd:
I originally posted this as a question on zulip but @fonsp asked me to submit an issue so here it is

I was looking at the [notebook](https://plutocon2021-demos.netlify.app/fonsp%20%E2%80%94%20javascript%20inside%20pluto) from the fonsp talk on using Javascript in Pluto
and I found out about the nice use of `<details>` html tag to create blocks of text that can be expanded by user click.

One issue I have with simply having the current way this works is that pluto output cells are "anchored" at the bottom, in the sense that if the cell expands the point on the window browser stays at the bottom of the output cell.
This is the opposite of what I would like to achieve for cells containing a `<details>` element as if the hidden piece of text is quite big, you are directly skipping the beginning of it and just see the bottom part of the hidden text when "showing" the element.

I uploaded a video to try and clarify better what I mean.

https://user-images.githubusercontent.com/12846528/120194880-76a0f700-c21e-11eb-9b96-fcce8e81bb2e.mp4

From the video it's clear that opening/closing the details element breaks the reading continuity.
Ideally one would want the point in the page to remain fixed on the line containing the `<summary>` html field.

I am sure there must be an easy way using html/css/js to simply change this behavior for cells that include `<details>` elements, but I have not much clue yet as to how to achieve this.

Is this something easy to fix?

fonsp:
I believe that this is caused by Pluto's scroll compensation: when a cell changes height, Pluto will scroll the page to compensate the height difference, **if** 

> the cell output that changed height appears above the currently focused cell

The strange scrolling behaviour with `detail` might be caused by a faulty implementation, or we might need to change the compensation rule above. It could also be the case that clicking `detail` will focus the element, but clicking it a second time blurs it, causing two different behaviours.

To investigate:
1. Is this caused by scroll compensation? Disable scroll compensation in source code to figure it out
2. What should be changed to fix this behaviour?

disberd:
@fonsp If you can point me to the part of the Pluto code where this scroll compensation is happening I can try start having a look and maybe provide some comments to help understanding this issue.
I am quite new to the whole HTML/CSS/JavaScript ordeal but I love Pluto and plan on using it extensively for most of my future work/research so I am more than happy to have some useful study cases to dig deeper into it.

fonsp:
[...]

PR #1231, opened 2021-06-08
Fix scrolling when opening/closing <details>

disberd:
Fixes [#1204 ](https://github.com/fonsp/Pluto.jl/issues/1204)


fonsp:
simple but effective!

Issue #996, opened 2021-03-14
Output cell is too small for the DataFrame and shows scrollbar


jonasjonker:
When I output a dataframe, the output cell is to small for the DataFrame and therefor has a scrollbar.  I tried it in two browsers (Firefox and Brave). I assume this is not a feature. What is the best way for me to help find the cause? 

![2021-03-14-124518_728x372_scrot](https://user-images.githubusercontent.com/24474645/111086984-7a768480-851f-11eb-9513-e106a13d420e.png)

fonsp:
This is a feature :)

jonasjonker:
Oke :)

GiggleLiu:
![image](https://user-images.githubusercontent.com/6257240/123518648-d9967a00-d674-11eb-82bf-d769d9521165.png)

Hi, is there an easy solution to remove the scrollbar here?

```julia
md"""
When $m<b$, the mapping is $(m, b) \rightarrow (b, b)$. Let `m` and `b` be 64 bit integers.
If all states have equal probability, their information entropies are $128\log 2$ and $64 \log 2$
"""
```

Pluto@v0.14.7

holomorphism:
@GiggleLiu
I was also wondering about the same issue.

When I checked the MathJax formula, the height of the element "mjx-assistive-mml" was exceeding the element "pluto-output", so I think this might be the cause.

![image](https://user-images.githubusercontent.com/3076585/123520050-afcb6a80-d6e9-11eb-98ee-69692aa0f90d.png)

I wasn't sure if this was a Pluto issue or a MathJax issue, so I hadn't reported this issue yet.


After adding the following custom css, the extra scrollbars are no longer displayed, so I have adopted this workaround when publishing notebooks as html.

```julia
"""
<style>
mjx-assistive-mml { display: none !important; }
</style>
""" |> HTML
```

However, I am not sure if hiding "mjx-assistive-mml" will cause any problems.

GiggleLiu:
@holomorphism Thanks for your magic spell, it solves my problem immediately.

PR #1261, opened 2021-06-22
 prevent scrollbar on dataframe output

jonasjonker:
Any dataframe with more than 12 rows automatically has a scrollbar in Pluto since it is more than 400px high. 450px is enough to fit dataframes with more than 15 rows since dataframes automatically hides rows. 502px is enough to show up to 15 rows without scrollbar.

I mentioned the scrollbar before [here](https://github.com/fonsp/Pluto.jl/issues/996), but was told is was a feature. I'm understand that there might be good reasons not to increase the max output size that I'm not aware of. But I thought I give this PR a try anyway.

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -349,7 +349,7 @@ pluto-output div.footnote p:last-child {
 
 pluto-output.scroll_y {
     max-height: 80vh;
-    max-height: 400px;
+    max-height: 502px;
     overflow: auto;
 }
 


fonsp:
Can you use screenshots to explain what you mean?

jonasjonker:
Current situation with scrollbar:
image

Increasing max-height removes scrollbar for both 15 row and 15+ row dataframes:
image
image

jonasjonker:
Of course I agree that scrollbars should kick in when you show more rows of the dataframe. But I don't like that they are there in the default view. But that might be just me

fonsp:
and why 502?

jonasjonker:
Yes 502 was kind of a magic number. I did did some further tests with different column types. Types that are displayed as single line of alphanumeric characters per row fit within the 502px on my system. However I found that that zooming my browser to 110% made the scrollbar reappear again. 

In general I would sat that alphanumeric characters seem to be quit uniform in size, even when the extended Latin Alphabet is used. This isn't necessarily true for other alphabets. I checked some strings of random Chinese characters, which rarely fit within the 502px I mentioned...

Types represented on multiple lines such as matrices, and types not represented as characters such as colors don't (necessarily) fit either.

I can't tell you if the 502px works across system, or that it happens to be the magic number on my screen.

I'm no css wizard, but if you agree that any dataframe with rows of _just single line alphanumeric characters_ should be displayed without scrollbar I would be willing to spend some time to try find a good way to achieve this. Or you could change it to something like 550px and hope that covers most cases.

### single line per row alphanumeric characters.
image
image

### subset of Chinese characters
image

### non-single line per row and non-character  representation

image

image

Issue #1255, opened 2021-06-16
Notebook location of pasted cells


disberd:
When pasting copied cells in the input of an empty cell, I was expecting the pasted cells to appear in the position of the notebook where the paste command was sent.

At the moment regardless of where the paste command is sent, the pasted cells are appended to the end of the notebook.
One can always move them where needed afterwards, but it can be a bit inconvenient when dealing with pretty long notebooks.

In fact, while the `add-deserialized-cells` accepts the index of paste as argument, the value of -1 is currently given in the Cellinput.js file:
https://github.com/fonsp/Pluto.jl/blob/2475e293b49c1824675af7b7d97752f3e86811f0/frontend/components/CellInput.js#L582

Is this intended for some reason or could this be fixed to have the pasted cell appear at the point of the notebook where you send the paste command?

PR #1263, opened 2021-06-22
Paste cells after currently selected input cell

disberd:
Initial attempt to fix [#1255 ](https://github.com/fonsp/Pluto.jl/issues/1255)

This simply add the pasted cells after the input cell where the paste command has been issued:

https://user-images.githubusercontent.com/12846528/122967186-7feb3280-d38a-11eb-862e-fff8477b2ea4.mp4

Paste events outside of input cells are still put at the end of the notebook and I don't know whether there is a better way to get the position of the current cell

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -476,7 +476,7 @@ export const CellInput = ({
             const topaste = e.clipboardData.getData("text/plain")
             const deserializer = detect_deserializer(topaste, false)
             if (deserializer != null) {
-                pluto_actions.add_deserialized_cells(topaste, -1, deserializer)
+                pluto_actions.add_deserialized_cells(topaste, editor_state.notebook.cell_order.indexOf(cell_id)+1, deserializer)
                 e.stopImmediatePropagation()
                 e.preventDefault()
                 e.codemirrorIgnore = true
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -476,7 +476,7 @@ export const CellInput = ({
             const topaste = e.clipboardData.getData("text/plain")
             const deserializer = detect_deserializer(topaste, false)
             if (deserializer != null) {
-                pluto_actions.add_deserialized_cells(topaste, editor_state.notebook.cell_order.indexOf(cell_id)+1, deserializer)
+                pluto_actions.add_deserialized_cells(topaste, cell_id, deserializer)
                 e.stopImmediatePropagation()
                 e.preventDefault()
                 e.codemirrorIgnore = true
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -262,13 +262,18 @@ export class Editor extends Component {
                     code_folded: false,
                     running_disabled: false,
                 }))
-                /* Try parsing the input as a number */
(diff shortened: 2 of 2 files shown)


fonsp:
Can you try to do it without `editor_state`? That would break out of the React logic and might cause problems later.

fonsp:
Also thank you @disberd ! This issue also bothers me and I am happy to see that you are contributing!

disberd:
Sure, I just did the first thing that came to mind (I also don't know React).
I started yesterday evening to see how React works and when I learn a bit more will try a modification

disberd:
@fonsp 
I tried an alternative implementation on the line you suggested, modifying the add_deserialized_cells to accept either a number or a cell id.
The behavior is slightly changed as now the cells are pasted *above* the input cell rather than below as this leads to *cleaner* and smaller code change.
If the original behavior is the preffered way one could add another if inside the `isNan` check to add 1 to index just when it is not -1 (To keep the nice fallback to the original behavior)

fonsp:
It should come after the focused cell, and use `typeof(x) === 'number'` instead of conversion.

The tests are failing on main too, should be fixed soon

disberd:
@fonsp 
I did as you asked and switched to use typeof and enforce paste after the current cell

Just for my own education, what is the reason to avoid conversion and go for the `typeof(x)`?

disberd:
I also added a small additional feature in the last commit.
If the cell where the paste event is fired is empty, the first `deserialized` cell is directly pasted inside of it instead of beging created after it.
I did as usually my use case is to create an empty cell before the past event in order to specify the location of the notebook where I want to paste my cells.

fonsp:
Thanks! The code you added for updating an empty codemirror uses a couple of hacks, making it hard for us to maintain. Can you remove it? You can continue working on it in a new PR if you like.

fonsp:
> Just for my own education, what is the reason to avoid conversion and go for the `typeof(x)`?

No technical reason, but my style preference is for code to be _descriptive_, not _imperative_. With the new code, you _describe_ your intention using code, instead of using code to implement your intended behaviour.

disberd:
> Thanks! The code you added for updating an empty codemirror uses a couple of hacks, making it hard for us to maintain. Can you remove it? You can continue working on it in a new PR if you like.

I just reverted the last commit.

In order to prepare a new PR for improving on my solution, what is the thing that is considered a hack there?

Is it checking if the cell is empty with `cm.getValue()`?

The indexOf I also used in the previous commit and the rest of the cell manipulation I took from:
https://github.com/fonsp/Pluto.jl/blob/71753aae76309e62f10f3d335b797223c7bc6b62/frontend/components/Editor.js#L315-L319
at the bottom of `add_deserialized_cells`

fonsp:
Thanks!

The 'hack' is that we wrap the CodeMirror in a react component: https://github.com/fonsp/Pluto.jl/blob/main/frontend/components/CellInput.js to isolate its _statefulness_ from the rest of our codebase. To read/write the codemirror input, you should go through react, setting the state of the notebook, which in turn updates the codemirror.

In your case, this is probably more difficult than setting the codemirror value directly, but it prevents maintenance headaches later.

PR #1280, opened 2021-07-01
Remove `project` configuration

fonsp:
Because:
- I noticed that this system did not work as expected: it uses the `pwd` of the Pluto server to resolve the environment for notebook processes. We want these to be disconnected.
- Since #844, the environment is now either embedded in the notebook file (default), or set up using `Pkg.activate` inside notebook code. 

Finally, this option conflicts with the new Pkg logic, and no longer works. Let's see if the need for such an option arises again, but the new Pkg behaviour might cover all use cases.

### DIFF
--- a/src/Configuration.jl
+++ b/src/Configuration.jl
@@ -105,7 +105,6 @@ These options will be passed as command line argument to newly launched processe
 - `banner::Union{Nothing,String} = nothing`
 - `optimize::Union{Nothing,Int} = nothing`
 - `math_mode::Union{Nothing,String} = nothing`
-- `project::Union{Nothing,String} = "@."`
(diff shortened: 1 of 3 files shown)


fonsp:
To add to this:

> I noticed that this system did not work as expected: it uses the pwd of the Pluto server to resolve the environment for notebook processes. We want these to be disconnected.

I removed this behaviour because it was confusing. For the same reason that launching `julia` in a project dir does not activate it: _"starting ~julia~ Pluto in a directory is not a sufficiently strong signal from the user that they trust the content of that directory and want to let it control what code they run"_ (from Stefan Karpinski https://github.com/JuliaLang/IJulia.jl/pull/820#issuecomment-470960416)

PR #1317, opened 2021-07-14
Allow assignment by destructuring

### LINKED ISSUE #1312: New property destructuring `(; a, b) = (a = 5, b = 6)` doesn't work
Property destructuring (https://github.com/JuliaLang/julia/pull/39285) might be the most exciting new feature of Julia 1.7. 

```julia
julia> struct Foo
           a
           b
       end

julia> x = Foo(1, 2)
Foo(1, 2)

julia> (; b) = x
Foo(1, 2)

julia> b
2
```

(Previously this was achieved by `using UnPack; @unpack b = x `)

This new syntax doesn't seem to be supported by Pluto right now.

https://user-images.githubusercontent.com/6280307/125408157-e5937300-e3ba-11eb-8819-ddd939594437.mov

1. `b` doesn't update after changing `x`.
2. Changing the destructuring expression leads to an error

greimel:
Fixes #1312.

The destructuring PR  (JuliaLang/julia#39285) allows many more things than this, but this is the feature that I am using.

This was the first time I've looked into `Pluto.jl/src` so I cannot be sure if this is the correct way to go.

https://user-images.githubusercontent.com/6280307/125588197-ea52620a-bc05-4b72-89e7-9023351faab6.mov

(this feature needs Julia 1.7)

### DIFF
--- a/test/ExpressionExplorer.jl
+++ b/test/ExpressionExplorer.jl
@@ -115,6 +115,7 @@ Some of these @test_broken lines are commented out to prevent printing to the te
         @test testee(:((x, y), a, (b, c) = z, e, (f, g)), [:z, :e, :f, :g], [:x, :y, :a, :b, :c], [], [])
         @test testee(:((x[i], y.r), a, (b, c) = z, e, (f, g)), [:x, :i, :y, :z, :e, :f, :g], [:a, :b, :c], [], [])
         @test testee(:((a[i], b.r) = (c.d, 2)), [:a, :b, :i, :c], [], [], [])
+        VERSION > v"1.7.0-0" && @test testee(:((; a, b) = x), [:x], [:a, :b], [], [])
     end
     @testset "Broadcasting" begin
         @test testee(:(a .= b), [:b, :a], [], [], []) # modifies elements, doesn't set `a`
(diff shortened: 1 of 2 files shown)


Pangoraw:
Hi!

I think you need to explicitly check that the first arg is a `:parameter` because for example the following expression is also valid in an assignment context: `(a[1], b)`

You can also add tests in `test/ExpressionExplorer.jl`:

https://github.com/fonsp/Pluto.jl/blob/28e452bc0582bb53c25768e02169a65ffa436e99/test/ExpressionExplorer.jl#L103

greimel:
Thanks for the hint. Tests pass now locally.

test/ExpressionExplorer.jl
```diff
@@ -115,6 +115,7 @@ Some of these @test_broken lines are commented out to prevent printing to the te
         @test testee(:((x, y), a, (b, c) = z, e, (f, g)), [:z, :e, :f, :g], [:x, :y, :a, :b, :c], [], [])
         @test testee(:((x[i], y.r), a, (b, c) = z, e, (f, g)), [:x, :i, :y, :z, :e, :f, :g], [:a, :b, :c], [], [])
         @test testee(:((a[i], b.r) = (c.d, 2)), [:a, :b, :i, :c], [], [], [])
+        VERSION > v"1.7.0-0" && @test testee(:((; a, b) = x), [:x], [:a, :b], [], [])
```
greimel, line 118: not sure if the version check is needed.

I'd try what happens on CI, but GHA workflows are still disabled for me.

greimel:
I removed the version check and tests pass locally for 1.6 and 1.7

(would you mind approving the workflow once more @fonsp)

fonsp:
I'm on holiday, assigning @Pangoraw

Pangoraw:
Thanks @greimel ! :raised_hands:

Issue #1278, opened 2021-06-29
Open multiple notebooks at start and without browser


lungben:
You can open a default notebook with the Pluto option
`Pluto.run(notebook="test.jl")`

Especially when using Pluto as a REST Server (with https://github.com/fonsp/Pluto.jl/pull/1052 ) it would be good to have the possibility to start multiple notebooks from the command line.

Suggested syntax:
`Pluto.run(notebook=["test.jl", "test2.jl"])`

Currently, the opening of a notebook works by passing the path to the browser (e.g. http://localhost:1234/open?secret=r6d46b55&path=test.jl). However, this only works when a browser is actually opened, which may not be the case if Pluto is used as a REST server. It would be good to have a way to open a notebook on server side only, without browser.

fonsp:
Currently, `notebook` is part of our configuration, but that is a bit strange (since it depends on the location where you launch pluto).

We should remove it from the Configuration and instead add it as kwarg to `Pluto.run`.

Then we add two kwargs, `notebook` and `notebooks` (for backwards compat) and we do:

`[notebook, notebooks...]`

or multiple dispatch :))

PR #1324, opened 2021-07-15
Open multiple notebooks at start and without browser

lungben:
#1278 

- [ ] support opening single notebook without need to start browser
- [ ] remove notebook to open from Configuration and add it as kwarg to `Pluto.run()`
- [ ] support opening multiple notebooks

### DIFF
--- a/src/webserver/WebServer.jl
+++ b/src/webserver/WebServer.jl
@@ -74,6 +74,12 @@ end
 
 function run(options::Configuration.Options)
     session = ServerSession(;options=options)
+
+    notebook_to_start = options.server.notebook
+    if notebook_to_start !== nothing
+        SessionActions.open(session, notebook_to_start; compiler_options=options.compiler)
+    end
+
     run(session)
 end
 


lungben:
I kept the `notebook` in `ServerOptions` for backward compatibility. This field is still used for opening the defined notebook in the browser.
If the keyword argument `notebook` is given to `Pluto.run()`, this notebook is now opened independently from the browser, so that this also works if `launch_browser=false`. If `launch_browser=true`, the same notebook is also opened in the browser.
The keyword argument `notebook` now supports an AbstractArray of notebook paths, all these notebooks are opened now in the backend. If `launch_browser=true`, the first notebook in this Array is additionally opened in the browser.

fonsp:
This will add 10 seconds to our total test time because of the sleeps. Do we need the test?

src/webserver/WebServer.jl
```diff
@@ -68,12 +68,32 @@ This will start the static HTTP server and a WebSocket server. The server runs _
...
+# open notebook(s) on startup
+function run(options::Configuration.Options, notebook:: Nothing = nothing)
+    session = ServerSession(;options=options)
+    run(session)
 end
 
-function run(options::Configuration.Options)
+function run(options::Configuration.Options, notebook:: AbstractString)
     session = ServerSession(;options=options)
+    SessionActions.open(session, notebook; compiler_options=options.compiler)
```
fonsp, line 88: Do you need to manually pass along compiler_options? Are they not taken from the session?

src/webserver/WebServer.jl
```diff
@@ -68,12 +68,32 @@ This will start the static HTTP server and a WebSocket server. The server runs _
 Pluto notebooks can be started from the main menu in the web browser.
 """
 function run(; kwargs...)
-    options = Configuration.from_flat_kwargs(; kwargs...)
-    run(options)
+    notebook = get(kwargs, :notebook, nothing)
+
+    # for backward compatibility - Configuation can only contain a single notebook for startup
```
fonsp, line 73: Why not allow the notebook config to be a vector? Then we don't need to rewrite here

src/webserver/WebServer.jl
lungben: Right - I'll remove this.

lungben:
> This will add 10 seconds to our total test time because of the sleeps. Do we need the test?
> 
> Maybe you can poll instead. Search for 'poll' in our source code, we already have a function to do it

I reduced the wait time to 3s.
How is poll working in this context? I could not find an example usage in the test cases.

src/webserver/WebServer.jl
lungben: done

PR #1328, opened 2021-07-19
Remove scrollbar from in-line latex in markdown

### LINKED ISSUE #1309: Markdown cell output overflows when last row of contains LaTeX
Whenever I create a markdown cell and there is a latex rendered element in the last row of the text, the cell output does not automatically update its height and a small overflow/scrollbar appear:

image

Any tips on how to fix this?

disberd:
Fix #1309

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -398,6 +398,11 @@ kbd {
     border-radius: 3px;
 }
 
+/* Avoid scrollbar in in-line latex with markdown, see https://github.com/fonsp/Pluto.jl/issues/1309 */
+pluto-output mjx-assistive-mml {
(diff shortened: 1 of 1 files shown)


disberd:
Is there a way to run the tests locally to see what is the problem with failing instead of pushing new updated to github every time?

fonsp:
I don't completely understand what the "Assistive MML" is for, but I suspect that it is only there for screen readers, so changing the height to 1px is fine.

fonsp:
The frontend tests are unstable, but you can read instructions here: https://github.com/fonsp/Pluto.jl/tree/main/test/frontend

PR #1351, opened 2021-07-31
Add integrity verification for externally loaded .js and .css

### LINKED ISSUE #1315: Integrity checks for javascript loaded from external sources
Instead of something like this
```
<script
    src="https://cdn.jsdelivr.net/npm/lodash@4.17.20/lodash.min.js"
    defer></script>
```
plase use something like this
```
<script
    src="https://cdn.jsdelivr.net/npm/lodash@4.17.20/lodash.min.js"
    integrity="sha384-EnW5v27oxpPV45dNJxtnA2OWfHpDTXuO/EzFl+c1vzZcB1aWeANNH0pFZLJ6Gsbr"
    defer></script>
```
for all externally loaded css and js files.

This protects against compromised CDNs. This is especially important for Pluto, where a compromised CDN directly allows for arbitrary remote code execution. This is because anyone who controls what javascript code is executed within the browser also controls which julia code is executed outside the browser's sandbox.

At the moment, any admin of jsdelivr.net (or hacker or person who can change the files loaded by Pluto) can execute arbitrary code on the computer of any Pluto user. They can even target only specific users/IP addresses/... to hide an attack from the community's scrutiny etc. With supply-chain attacks becoming more and more common, I think this should be fixed. Especially since in this case it's as easy as adding a few hashes to .css and .js files.

uPvtkBlpiucvDUf:
This is a quick fix for #1315 . Especially for the imports in the js modules, this is only a quick-and-dirty fix using a manually added `checkScriptIntegrity` function.

I guess a proper fix should probably come with #1125 . The correct way would probably be to do proper js and css dependency
management using something like npm. Then one can either bundle everything in Pluto.jl or ship an (integrity-verified) artifact.

Until such a proper fix is merged, I think is is an acceptable short-term solution.

### DIFF
--- a/frontend/components/CellOutput.js
+++ b/frontend/components/CellOutput.js
@@ -401,7 +401,10 @@ export let highlight = (code_element, language) => {
                 code_element.classList.add("cm-s-default")
             },
             {
-                path: (mode) => `https://cdn.jsdelivr.net/npm/codemirror@5.60.0/mode/${mode}/${mode}.min.js`,
+                // Commented out since 
+                // 1) not needed since this is manually loaded in editor.html already
+                // 2) it is loaded w/o integrity verification which we don't want
+                // path: (mode) => `https://cdn.jsdelivr.net/npm/codemirror@5.60.0/mode/${mode}/${mode}.min.js`,
             }
         )
     }
--- a/frontend/components/CellOutput.js
+++ b/frontend/components/CellOutput.js
@@ -10,6 +10,8 @@ import { observablehq_for_cells } from "../common/SetupCellEnvironment.js"
 import { PlutoBondsContext, PlutoContext, PlutoJSInitializingContext } from "../common/PlutoContext.js"
 import register from "../imports/PreactCustomElement.js"
 
+import checkScriptIntegrity from '../imports/check-integrity.js'
+
 //@ts-ignore
 const CodeMirror = window.CodeMirror
 
@@ -379,8 +381,14 @@ export let RawHTMLContainer = ({ body, persist_js_state = false, last_run_timest
     return html`<div class="raw-html-wrapper" ref=${container}></div>`
 }
 
+const modeSRIHashes = {
+    'julia': 'sha384-dDyOA8BSiBRH42Ba2yJjj2j5eS8IiLUZfKmb1ZIFQOrthebHHYr2M8RXTeMk7ibZ',
+    'python': 'sha384-BbNKqfbjrUCjqp31ddcnhy4I0iLdCYELtYKiHaLNTdtlR3Y0uAKqUZHTxssmXMAq',
+    // Add more hashes here...
+};
+
 /** @param {HTMLElement} code_element */
-export let highlight = (code_element, language) => {
+export let highlight = async (code_element, language) => {
     if (code_element.children.length === 0) {
         let mode = language // fallback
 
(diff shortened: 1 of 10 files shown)


fonsp:
Thanks for pointing this out @uPvtkBlpiucvDUf ! I am currently on holiday, but I will be sure to work on getting your PR merged when I get back. So far it looks great, thanks for putting in the effort!

frontend/editor.html
```diff
@@ -23,24 +23,118 @@
...
+    <script
+        src="https://cdn.jsdelivr.net/npm/@observablehq/stdlib@3.3.1/dist/stdlib.js"
+        integrity="sha384-puTWMrPXLOGQnFXpGilbRDiGIrohFMYhoqp9Ot/ZuJpZyZ23hA0ozqEqs75pycVy"
+        crossorigin="anonymous" defer></script>
+    <script
+        src="https://cdn.jsdelivr.net/npm/iframe-resizer@4.2.11/js/iframeResizer.min.js"
+        integrity="sha384-fTp2I0abasco+1TIQhjTbS/v5ihiE09Kbng/klOF2y/DtZCcieGzK+ierZKtNBBx"
+        crossorigin="anonymous" defer></script>
+    <script
+        src="https://cdn.jsdelivr.net/npm/codemirror@5.60.0/lib/codemirror.min.js"
```
fonsp, line 39: Some of our minimized assets, like this one, should not be used with SRI (click on the `.min.js` link to read the comment. So we should use the original, unminimized asset here.

frontend/components/CellOutput.js
```diff
@@ -401,7 +401,10 @@ export let highlight = (code_element, language) => {
                 code_element.classList.add("cm-s-default")
             },
             {
-                path: (mode) => `https://cdn.jsdelivr.net/npm/codemirror@5.60.0/mode/${mode}/${mode}.min.js`,
+                // Commented out since 
+                // 1) not needed since this is manually loaded in editor.html already
+                // 2) it is loaded w/o integrity verification which we don't want
+                // path: (mode) => `https://cdn.jsdelivr.net/npm/codemirror@5.60.0/mode/${mode}/${mode}.min.js`,
```
fonsp, line 407: This code will lazy-load codemirror modes for new programming languages, to highlight code snippets in HTML output. E.g. a `python` code block will trigger loading the Python mode to highlight that block.

It will require some extra work to make this work with the SRI-only restriction. @schneiderfelipe might have ideas

fonsp:
I wrote down two quick comments, but will do a more complete review next month.

frontend/components/CellOutput.js
fonsp: Note that we are planning to switch to codemirror 6 soon, which would require rewriting this mechanism anyways. We might need to switch to a dedicated highlighting library, which would come preloaded with many programming languages. 

We could look into https://highlightjs.org/ as an alternative to this lazy-loading trick.

frontend/editor.html
uPvtkBlpiucvDUf: Good catch! I switched those to use the unminimized version.

frontend/components/CellOutput.js
uPvtkBlpiucvDUf: Ah, I see. I just added SRI verification also for these lazy-loaded files. This requires manually adding the hashes for all mode files we want to support to the `modeSRIhashes` dictionary (see newest commit). I could write a small script to auto-generate all hashes from this directory listing

https://cdn.jsdelivr.net/npm/codemirror@5.60.0/mode/

This should not be too much effort, but I haven't done that yet. If you think this way of doing things is okay, I can write this script. If you decide to switch to a different highlighting library right away, this will not be necessary of course.

fonsp:
The CodeMirror 6 rewrite (https://github.com/fonsp/Pluto.jl/pull/1159) is going faster than expected and should be done within a week, so I think it makes sense to wait with this PR until that one is done.

After the rewrite, I would like to revisit #241 and #1315 together. If no short-term solution comes out then we'll go with this PR.

dralletje:
Sorry for coming with this so late, but this doesn't work on the safari (the top level await part).

Apart from that, I'm really doubtful of how useful this is.
Are we really afraid of jsdelivr getting hacked? 🙃

uPvtkBlpiucvDUf:
Good to know safari doesn't support this yet! I will check how we can work around this in case no other short-term solution is found (see @fonsp's comment).

As to how useful it is: It does protect against at least

- jsdelivr getting hacked
- a rogue jsdelivr admin
- a rogue CA (a la https://www.eff.org/deeplinks/2011/09/post-mortem-iranian-diginotar-attack)
- evil takeover of any of the projects we download from jsdelivr (a la https://blog.npmjs.org/post/180565383195/details-about-the-event-stream-incident)

Any of these otherwise allow to run arbitrary code on your Pluto-running computer outside the browser's sandbox. This is a much worse problem than for most websites.

I agree it's not something that needs an emergency fix in the next few days. But I think this should definitely be fixed. As I said above, I guess the proper solution would be to do JS dependency management using npm or similar and then ship everything with Pluto (either directly or as an integrity-verified artifact). Or something of that sort. But if this is not happening soon, I would like something like this PR to be merged as a short(er)-term solution.

dralletje:
All assets will soon be bundled with Pluto, just need to see how release will work

fonsp:
@uPvtkBlpiucvDUf thanks again for the report and for thinking together about the fix! Very much appreciated ❤️

PR #1029, opened 2021-03-26
📬 Auto reload from file

fonsp:
Watch the notebook file for changes, and load them automatically. New changes are run automatically.

The goal here is to allow:
- Modifying a notebook using git
- Small changes in VS Code

# Try it out

```julia
Pluto.run(auto_reload_from_file=true)
```

# TODO

- [x] Throttle file events, wait until file stops changing
- [x] Handle corrupt notebook file
- [x] Detect changes to embedded Pkg https://github.com/fonsp/Pluto.jl/commit/58989314a2944422ea55ae73b88774c1af21689c https://github.com/fonsp/Pluto.jl/commit/3f65b53b0ddd064586eb6f4f4d773192b22202fb
- [x] Don't trigger when Pluto changes the file
- [x] Detect file deleted and ... ?
- [ ] Test on 3 OSses
- [x] Failing windows 1.6 test?
- [x] Tests
- [x] #1377

# VIDEO

https://user-images.githubusercontent.com/6933510/112561129-e1742300-8dd4-11eb-98cf-952ce4e1b7b6.mov

this PR + https://www.notion.so/malyvsen/Notebook-REST-API-4312a1acf7004806bd33f2eefee327a5 == https://darklang.com ?! @ctrekker

PR #1377, opened 2021-08-09
Update notebook from file with popup

disberd:
Very much inspired in functionality by https://github.com/fonsp/Pluto.jl/pull/1029, but instead of automatically reloading from file, provide a small disclaimer on the notebook (basically copied in style from the one reminding you to save updates when modifying a notebook) that actually reload the notebook from the file upon click.

I did implement this small modification as sometime I had issues with too many updates from file with the greedy update from https://github.com/fonsp/Pluto.jl/pull/1029.

Here is an example functionality:
update_from_file_example

Let me know if this is something that might be considered interesting or if I will simply just keep it on my fork

Still to do:
- [x] Track notebook file movement from Pluto
- [x] Maybe have a configuration option to always do greedy reload without waiting for feedback
- [ ] Implement throttle for greedy update as suggested by @fonsp
- [ ] Handle cell deletion from file
- [ ] Handle cell addition from file

Edit: Added configuration option to default to greedy reload instead of waiting for confirmation. Can be enabled doing
```julia
Pluto.run(;auto_reload_notebook_from_file=true)
```

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -549,7 +549,6 @@ export class Editor extends Component {
                     {},
                     {notebook_id: this.state.notebook.notebook_id}
                 )
-                console.log('message',message)
                 this.setState({last_update_time: Date.now()})
                 this.setState({notebook_file_newer: false})
                 return
@@ -647,9 +646,6 @@ patch: ${JSON.stringify(
                         handle_log(message, this.state.notebook.path)
                         break
                     case "update_notebook_filetime":
-                        console.log(message)
-                        console.log(this.state.last_update_time)
-                        console.log(this.state.notebook_file_newer)
                         /* Only update the state here if the notebook_file_newer is false */
                         if (message.timestamp - this.state.last_update_time > 1000 && !this.state.notebook_file_newer) {
                             this.setState({notebook_file_newer: true})
@@ -1064,10 +1060,6 @@ patch: ${JSON.stringify(
 
         if (old_state.disable_ui !== this.state.disable_ui) {
             this.on_disable_ui()
-        }
-
-        if (old_state.last_update_time !== this.state.last_update_time) {
-            console.log('last_update_time changed from',old_state.last_update_time,'to',this.state.last_update_time)
         }
     }
 
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -201,6 +201,7 @@ export class Editor extends Component {
             desired_doc_query: null,
             recently_deleted: /** @type {Array<{ index: number, cell: CellInputData }>} */ (null),
             last_update_time: 0,
+            last_file_modified: 0,
 
             disable_ui: this.launch_params.disable_ui,
             static_preview: this.launch_params.statefile != null,
@@ -542,6 +543,13 @@ export class Editor extends Component {
                 const { message } = await this.client.send("nbpkg_available_versions", { package_name: package_name }, { notebook_id: notebook_id })
                 return message.versions
             },
+            reload_from_file: () => {
+                return this.client.send(
+                    "reload_from_file",
+                    {},
+                    {notebook_id: this.state.notebook.notebook_id}
+                )
+            }
         }
 
         const apply_notebook_patches = (patches, old_state = undefined) =>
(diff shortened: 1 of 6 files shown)


fonsp:
This is fantastic! Some ideas:
- We should try to keep unsubmitted code changes in the browser. This probably works already?
- I am thinking that we might want to *always* update from file, without asking the user, like in #1029, but use your new UI to tell the user what happened. This fits the "reactive magic" of Pluto, and I think that this is the best option for beginner users.
- What happens when the notebook file changes into something invalid? It would be great if we would tell the user.

We can discuss here or on our call tomorrow!

fonsp:
We could address the greedy update in #1029 using a throttle: when the file changes, 
1. read its new contents
2. wait 200ms
3. read its contents
4. if the last two read contents are not equal, go back to 2
5. update notebook

disberd:
> This is fantastic! Some ideas:
> 
> * We should try to keep unsubmitted code changes in the browser when reloading from file. This probably works already?
> * I am thinking that we might want to _always_ update from file, without asking the user, like in [Watch notebook file for changes #1029](https://github.com/fonsp/Pluto.jl/pull/1029), but use your new UI to tell the user what happened. This fits the "reactive magic" of Pluto, and I think that this is the best option for beginner users.
> * What happens when the notebook file changes into something invalid? It would be great if we would tell the user.
> 
> We can discuss here or on our call tomorrow!

Thanks for the feedback!

Regarding your points:
1. I do not think this already works (I didn't implement the reload from file so I don't remember now but I think i just overwrites the editor with the file contents at the moment).
2. I agree that this is probably the best default approach
3. By invalid you mean from the Pluto file synthax (like making a modification that breaks the Pluto parsing of the file?), I suppose that could be easy to detect and warn about

greimel:
When I modify notebooks in an editor my most common problem is that the uuids at the bottom get out of sync with the code cells. It would be nice if this got handled gracefully. (This also happened frequently when I commited indivual hunks in git and then stashed.)

(Tbh, I haven't this in a while, so I am not sure if this issue is already fixed.)

disberd:
> When I modify notebooks in an editor my most common problem is that the uuids at the bottom get out of sync with the code cells. It would be nice if this got handled gracefully. (This also happened frequently when I commited indivual hunks in git and then stashed.)
> 
> (Tbh, I haven't this in a while, so I am not sure if this issue is already fixed.)

Do you mean when you were deleting cells alltogether without updating the bottom cell order or in general when just changing code from within one cell?
If it's the latter, I think I never experienced this as a problem.

Regarding instead deleting or adding cells all-together within the notebook from the file would be a nice improvement.

- Deleting cells should not be too difficult (I suppose but I'd have to check), we would just have to make sure that one does not break up dependencies with other non deleted cells when doing so but that should be verifiable using the cell dependency tree.
- Adding a new cell is something that instead I found myself wanting to do more often and having to go through the notebook to do is inconvenient. One of the issues here is probably where to put the cell on the front-end once this is inserted from the file, given tht the file cell orders are different from the display order on the notebook

greimel:
> Do you mean when you were deleting cells alltogether without updating the bottom cell order

Yes, that's what I meant.

> Regarding instead deleting or adding cells all-together within the notebook from the file would be a nice improvement.

Yes!

> One of the issues here is probably where to put the cell on the front-end once this is inserted from the file

I think displaying the new cell just below the previous cell would be a good enough default. Don't you think? 

It would be nice to get a pop-up when a new cell is added to jump to this cell in the browser.

Would be nice if the following issues would be handled as well:
* if you copy-paste a cell in the source (leading duplicate uuid)
* if the new uuid is too short or too long or missing

don't error, just assign a new uuid.

fonsp:
These two new features: an easy way to delete or add cells from a text editor, requires some extra thinking and design, and I think we should do that in a separate PR. Let's get this PR ready and merged first, so that Pluto will work more smoothly together with git.

greimel:
Alright. But since you're making the point

>  so that Pluto will work more smoothly together with git.

I want to note that adding/deleting cells is an important feature for working smoothly with git. (In fact, I never edit notebooks in an editor.) 

Committing individual code blocks ("hunks") is very easy with modern git GUIs. And this doesn't play well with Pluto because the notebook doesn't load when the uuid list is out of sync with the code cells.

But I agree that this is a separate issue.

fonsp:
Thanks @disberd ! I used your idea to implement a pop-up message in #1029 

https://user-images.githubusercontent.com/6933510/134787985-259ddaff-3eca-4670-8d8b-c1884bc766f2.mov

disberd:
Thanks for taking the lead @fonsp, I am quite backlogged at work so I didn't have time to proceed yet :D

Issue #1402, opened 2021-08-19
Functions defined with docstrings are not reactive (regression in #main from 0.15.1)


disberd:
I just noticed that on the `main` branch, if one defines a function with a docstring, the function does not seem to be correctly registered among the cell definitions, and hence the reactivity between that function definition and the cells using that function is lost.

This can also be seen from the html on the left where `test1` is listed as definition immediately after the first cell, while `test2` is not in the second cell

![7c9ef307-0f07-4b64-86ed-ce1315c03142](https://user-images.githubusercontent.com/12846528/130088090-d79758bf-f8c8-466c-b7d8-244cc7c45aed.gif)

I tried using the registered 0.15.1 version and the problem didn't seem to be present there.

fonsp:
@Pangoraw I expect that this is caused by macro expansion, docstrings are interally stored as a `@doc` macro call.

Issue #1388, opened 2021-08-12
Base.@kwdef redefinition error with types following Unitful.jl documentation


disberd:
When trying to update a struct created using Base.@kwdef following the guidelines of the [Unitful.jl documentation](https://painterqubits.github.io/Unitful.jl/dev/highlights/#Dimensions-in-a-type-definition), there is an error in Pluto.

![image](https://user-images.githubusercontent.com/12846528/129211598-564a7d08-a35c-489e-ae70-7e6754e6ca19.png)

This is in all likelihood very similar to https://github.com/fonsp/Pluto.jl/issues/1338 so hopefully the solution is also similar.

Edit: Tagging @Pangoraw to confirm if it is indeed similar to https://github.com/fonsp/Pluto.jl/issues/1338

disberd:
It seems that directly putting the variables exported with Unitful.DefaultSymbols does solve the error upon redefinition, though it would still be nice to have this work either way.

Here is the different outputs of the `sanitize_expr`:

![image](https://user-images.githubusercontent.com/12846528/129215636-fb48b913-6c43-42aa-b222-4f65055e16b2.png)

Pangoraw:
Good catch, thanks! 
This is due to a set of macros being expanded on the Pluto process to avoid sending message to the notebook process but the expansion fails because there are unknown macros in the expression. Some macros like `@kwdef` and `@cmd` should probably be removed from this list.

https://github.com/fonsp/Pluto.jl/blob/1dd23a2431b3c3ff56871eca4ada7a425805c196/src/analysis/ExpressionExplorer.jl#L758-L759

PR #1420, opened 2021-08-27
Macro fixes

Pangoraw:
Fixes for #1388 and #1402


fonsp (approved):
Awesome!!

fonsp:
@Pangoraw are we ready to release 0.16 ?

Pangoraw:
Thanks to the many great reports from @disberd, I think most problems have been found ! We can add #1447 to main and then it should be ok :+1:

PR #1446, opened 2021-09-08
Reference outer constructor definition issue

disberd:
As disccused on zulip [here](https://julialang.zulipchat.com/#narrow/stream/257776-Pluto.2Ejl-development/topic/struct.20outer.20constructor.20in.20separate.20cell/near/251200448), adding a comment to ExpressionExplorer.jl to reference the issue about error when trying to define struct outer constructors outside of the struct definition cell


fonsp:
Thanks @disberd! Would love to meet when i'm back from my holiday!

disberd:
> Thanks @disberd! Would love to meet when i'm back from my holiday!

@fonsp sure! I'd love to!

PR #1472, opened 2021-09-17
Draft/WIP: Codemirror6 keymaps

pankgeorg:


dralletje:
I'm integrating this code into my cm6 branch now, and noticing that codemirror is a bit inconsistent with it's capturing of events. On my computer, if I had a cell `aaa`, and I held backspace from the last `a`, it would not give the backspace at the beginning of the cell to us, and it would just keep deleting... so I'm adding in some more checks 😁

Issue #1242, opened 2021-06-12
Some unicode completions don't work


aplavin:
Type `\^2<tab>` in a pluto cell - get `²` as expected. 
Type `\^-<tab>` and get `\^-	`: completion didn't trigger.
Meanwile, julia repl correctly gives `⁻` for the latter.

fonsp:
To newcomers: search for `on_tab_key`

fonsp:
🔥🔥🔥

PR #1485, opened 2021-09-22
Fix unicode completions

dralletje:
And a lot more of completion stuff.
- ctrl + space is normal completion
- Unicode autocomplete works again
- If there is only one match & you used <kbd>Tab</kbd> to trigger autocomplete, it will not even show up
- Pressing `.` while autocomplete is open will, just like with CM5, select the completion and press the `.`.
- Necessary escaping for special symbols happens again (e.g. `Base.:!=` when you query for `Base.!=`)
  - It escapes it bit too much now, like `Base.:π` and `Base.:😁`, that needs fix later on the server
- If you complete on `Module.:` you'll see all the cool symbols!
- That's it for now?

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -48,6 +48,7 @@ import {
     StateField,
     StateEffect,
 } from "../imports/CodemirrorPlutoSetup.js"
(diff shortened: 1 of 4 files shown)


dralletje:
Fixes #1242

PR #1493, opened 2021-09-23
Support running inside VS Code webview

fonsp:
Use vscode extension API instead of websocket when detected

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -29,6 +29,7 @@ import { start_binder, BinderPhase, count_stat } from "../common/Binder.js"
 import { read_Uint8Array_with_progress, FetchProgress } from "./FetchProgress.js"
 import { BinderButton } from "./BinderButton.js"
 import { slider_server_actions, nothing_actions } from "../common/SliderServerClient.js"
+import { available as vscode_available } from "../common/VSCodeApi.js"
 
 const default_path = "..."
 const DEBUG_DIFFING = false
@@ -161,7 +162,7 @@ const url_logo_small = document.head.querySelector("link[rel='pluto-logo-small']
 const url_params = new URLSearchParams(window.location.search)
 const launch_params = {
     //@ts-ignore
-    notebook_id: url_params.get("id") ?? window.pluto_notebook_id,
+    notebook_id: (vscode_available ? null : url_params.get("id")) ?? window.pluto_notebook_id,
     //@ts-ignore
     statefile: url_params.get("statefile") ?? window.pluto_statefile,
     //@ts-ignore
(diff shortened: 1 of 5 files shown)


fonsp:
@pankgeorg can you make that last commit on main, not on this PR?

frontend/components/Editor.js
```diff
@@ -903,6 +907,9 @@ patch: ${JSON.stringify(
                 }
                 e.preventDefault()
             } else if (e.key.toLowerCase() === "s" && has_ctrl_or_cmd_pressed(e)) {
+                // If VSCode is around, we shouldn't 'Set and run all remote changed remote cells.
+                // Control + Save sends VSCode save
+                if (vscode_available) return this.run_selected()
```
fonsp, line 912: I don't understand this change, could you write a more elaborate comment @pankgeorg ?

fonsp: Maybe give a 1.2.3. timeline of what happens when you press Ctrl+S?

fonsp: I think the `Ctrl+S` is pretty much the central TODO that's keeping us from releasing the extension publicly, once we figure this out it's all beautiful 🌟

pankgeorg: I reverted the change and I'm going all in to server-aware un-run changes. wip wip wip wip

src/notebook/Notebook.jl
```diff
@@ -54,10 +54,26 @@ Base.@kwdef mutable struct Notebook
     wants_to_interrupt::Bool=false
     last_save_time::typeof(time())=time()
     last_hot_reload_time::typeof(time())=zero(time())
+    last_serialized_version::String = ""
+    write_out_fs::Bool=true
```
fonsp, line 58: We already have https://github.com/fonsp/Pluto.jl/blob/c1ba792fc154c306129b58eb723621544b91f59c/src/Configuration.jl#L58 for this

src/notebook/Notebook.jl
```diff
@@ -54,10 +54,26 @@ Base.@kwdef mutable struct Notebook
     wants_to_interrupt::Bool=false
     last_save_time::typeof(time())=time()
     last_hot_reload_time::typeof(time())=zero(time())
+    last_serialized_version::String = ""
```
fonsp, line 57: Let's track this in pluto-vscode instead

src/notebook/Notebook.jl
pankgeorg: this is notebook global and it prevents the save function from running at all. this flag is per-notebook and it only prevents the final write-out to the file. Thus, we can run save event listeners and have the latest text representation of the file

src/notebook/Notebook.jl
pankgeorg: The benefits of having this here too are
- we prevent multiple file IO
- we prevent emitting events for file change (through listeners) if the file wasn't changed (that's a weaker argument - it can be part of the listener)
- we can compare this version to the `last_unsaved_serialized_version` (incoming) that is what we want `Shift+Enter` to propagate to the VS Code workspace, before `Control + S` actually does the filesave(!) (can also be tracked in the listener)

pankgeorg: all in all, I like it here for the above reasons - but we don't _really_ need it
What about having a hash of the (string contents of the) file here?

PR #1503, opened 2021-09-25
Fix #1490, suffix column types with `?` with `Union{T,Missing}`.

### LINKED ISSUE #1490: Missing column types in Union{Missing,T}
Screenshot is self-explanatory:

image

MichaelHatherly:
Any particular place in the tests for adding something to cover this?

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -1077,6 +1077,10 @@ function tree_data(@nospecialize(x::Any), context::IOContext)
 
 end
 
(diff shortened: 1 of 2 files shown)


fonsp:
Awesome! Add a new cell here: https://github.com/fonsp/Pluto.jl/blob/77384c69cb8d09e7125a4a4784a840f62bbef85b/test/RichOutput.jl#L217

And for the test, something like `@test occursin("String?", string(notebook.cells[12].output.body))` is good enough

PR #1522, opened 2021-09-29
execute script tags in-place (document.currentScript issue)

cjdoris:
In response to the `document.currentScript` discussion over on Zulip, I've made this PR which executes the script tags in-place. It simply goes through each script tag and replaces it in-place with a duplicate, which causes it to be executed.

This fixes the specific problem I was having, and seems to not break other things - e.g. binding to an input still works.

### DIFF
--- a/frontend/components/CellOutput.js
+++ b/frontend/components/CellOutput.js
@@ -298,6 +298,20 @@ const execute_scripttags = async ({ root_node, script_nodes, previous_results_ma
     return results_map
 }
 
+const execute_scripttags_2 = async ({ root_node, script_nodes, previous_results_map, invalidation }) => {
+    console.log("execute_scripttags_2 v3")
+    let results_map = new Map()
+    for (let old_node of script_nodes) {
+        const new_node = document.createElement("script")
+        for (let attr of old_node.attributes) {
+            new_node.setAttribute(attr.name, attr.value)
+        }
+        new_node.text = old_node.text
+        old_node.parentNode.replaceChild(new_node, old_node)
+    }
+    return results_map
+}
+
 let run = (f) => f()
 
 export let RawHTMLContainer = ({ body, persist_js_state = false, last_run_timestamp }) => {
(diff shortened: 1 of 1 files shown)


cjdoris:
I've just read `JavaScript.jl` more thoroughly and understand the above points better now. That file says that the ObservableHQ-like behaviour is got by giving the script tag an id, but that's not quite true: you get this behaviour for all inline scripts, it's just that the return value isn't cached unless the id is set.

Anyway, there are other legitimate reasons for scripts to have an id, so can I suggest having this opt-in mechanism be a bit more explicit? e.g. an id that starts with "pluto-observable-"? Or some special attribute?

cjdoris:
Hello, sorry for the slow response, I have been on holiday.

I get that it's a breaking change, which is preferable to avoid, but the problem is that Pluto will strive to render anything that can output HTML, but interprets that HTML in a non-standard way (in the sense that it does not follow the HTML standard). This means that anything which implements a `show(::IO, ::MIME"text/plain", ::MyType)` method now needs to consider which display it is writing to (i.e. Pluto or not Pluto) in order to get the right behaviour. This fundamentally breaks the composability of Julia.

As far as I can see, there are two proper ways to fix this. The first is to only support rendering HTML which was specifically written for Pluto to render. The second is to put this Observable behaviour into an undefined part of the HTML spec, which is precisely what using the type attribute that I suggested does.

Another solution could be that the only real difference between an ordinary script and an Observable one is that the latter kind (usually) has a return statement at the top level, which is a syntax error (I think) in JS scripts. So we could use that to distinguish the two cases in a backwards compatible way, but it would require parsing the script to some degree.

It's your project of course and I don't want to twist any arms! But I think the issue is worth considering.

cjdoris:
Oh you also asked if I can change the script. My use case is that I'm wrapping Python objects which can produce HTML representations of themselves, so I'm not writing the HTML myself, it's a generic interface. Hence editing the HTML is not really an option, and if it were it would be very hacky, and anyway requires knowing whether or not the HTML will be rendered in Pluto, cf earlier remarks on composability.

PR #1532, opened 2021-10-03
Backward compatible with CodeMirror5 get/set value object.

pankgeorg:
The following line defines a `getValue` and a `setValue` to use from queries. Experimental!
```
@htl("""
<script>
	const { getValue, setValue } = currentScript.closest("pluto-cell").querySelector(".CodeMirror").CodeMirror
</script>
""")```

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -421,8 +421,11 @@ export const CellInput = ({
 
         // For use from useDropHandler
         // @ts-ignore
-        newcm.dom.CodeMirror = {
+        dom_node_ref.current.CodeMirror = newcm.dom.CodeMirror = {
             getValue: () => newcm.state.doc.toString(),
+            setValue: (value) => newcm.dispatch({
+                changes: {from: 0, to: newcm.state.doc.length, insert: value}
+            })
         }
 
         if (focus_after_creation) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by fonsp** [comment]
We are [already setting the `CodeMirror` on the codemirror6 div](https://github.com/fonsp/Pluto.jl/blob/d94f29ebe480d05a4b84f9c39aa54d263b43f76c/frontend/components/CellInput.js#L424-L426), so let's stick with that and add the `setValue`. It looks like I can't set the `CodeMirror` class on the node, so people will have to rewrite this to a `cm-editor` selector. (It was never public API.)

**Comment by fonsp** [comment]
Maybe we should think about adding future-proof public API?

**Comment by fonsp** [comment]
I added `setValue` https://github.com/fonsp/Pluto.jl/commit/3c59814ac764ecff0b1ebb4846d168b38c1fbd78

**Comment by disberd** [comment]
I have just one question for clarification.
With this PR closed and with just the `setValue` added from  https://github.com/fonsp/Pluto.jl/commit/3c59814ac764ecff0b1ebb4846d168b38c1fbd78, there is still no way to call `setValue` from the front-end right?

Just to understand if I have to keep using this PR code until a public API is implemented

**Comment by pankgeorg** [comment]
> I have just one question for clarification. With this PR closed and with just the `setValue` added from [3c59814](https://github.com/fonsp/Pluto.jl/commit/3c59814ac764ecff0b1ebb4846d168b38c1fbd78), there is still no way to call `setValue` from the front-end right?
> 
> Just to understand if I have to keep using this PR code until a public API is implemented

This should work:

```html
@htl("""
<script>
	console.log(currentScript.parentElement.closest("pluto-cell").querySelector(".cm-editor").CodeMirror.getValue())
</script>
""")
```

pankgeorg:
> I simply do not understand the react structure enough to understand to what those methods were attached to

They are attached to a property of the DOM node called 'CodeMirror'. DOM nodes (HTML elements) are also JavaScript objects which you can access properties of. And they can also have custom attributes, like this one.

PR #1533, opened 2021-10-03
Highlight Markdown, HTML, JavaScript, Python, SQL within cells

pankgeorg:
Very much WIP
image

image
image

TODOs:
- [x] Fix nested `<script>` and `<style>` tags (they [should work](https://github.com/codemirror/lang-html#user-content-htmllanguage) 🤔)
- [X] SQL highlight
- [x] Fix highlight's plugin crash
- [x] Better colors/styles

~[possibly related](https://discuss.codemirror.net/t/parsemixed-javascript-or-into-html-inside-parsedmixed-off-by-something/3596)~ fixed!


pankgeorg:
[possibly related](https://discuss.codemirror.net/t/parsemixed-javascript-or-into-html-inside-parsedmixed-off-by-something/3596)

disberd:
Does this does just highlighting or it also handles formatting (like indenting and auto-close) on its own depending on the context?
In any case very cool addition!

pankgeorg:
> Does this does just highlighting or it also handles formatting (like indenting and auto-close) on its own depending on the context? In any case very cool addition!

It seems to autoclose html tags for sure, along with other nice things, yes. Please test it so we know more! It contains a lot of codemirror magic

disberd:
Cool!
I'll start trying then

Pangoraw:
Very cool! It seems that writing an `html""` value and coming back to Julia code removes the initial Julia highlighting (Firefox 92.0.1).

cm6-mixed-modes

pankgeorg:
The highlight plugin crashes ungracefully for some reason :( 
image

Looking into that

MichaelHatherly:
This is really nice! Is support for `py""` from `PyCall` and `R""` from `RCall` feasible as well?

pankgeorg:
> This is really nice! Is support for `py""` from `PyCall` and `R""` from `RCall` feasible as well?

Thanks! well, it _should_ feasible, yes. I'll test somethings out!

pankgeorg:
Note to self: Remove interpolation highlight from macros (That's a todo in julia-lang codemirror package or something)

disberd:
I just tried pulling again the new colors and it doesn't completely destroy Pluto anymore.

I still have the plugin crashing with `html""""""` cells (no highlighting)

I have also experienced problems with `<script>` tags autoclosure:

As soon as I press the closing `>`, the cell error on the console appears and the closing bracket is not shown in the cell input.

aabfbee6-3eb9-469d-8af6-dc2191c1f467

disberd:
It's nice that in markdown the `-` is added at the beginning of the line when pressing enter in a bullet list.
I noticed though that pressing Tab on a new bullet list line does not increase the indent level but just puts a tab in the text. It would be cool if we could also have this feature like in VSCode.
(It seems that Shift+Tab already decrease indent if you are on an indented line.

pankgeorg:
@Pangoraw Do you remember what was breaking in this? It seems to be working now (let's break it or merge merge merge!!!).

pankgeorg:
@dralletje 
image
I'm removing the syntax highlighting from the interpolation here - can we also stop this underlining? (a pointer plsss :p )

dralletje:
That shouldn't be an interpolation even, no? That's a thing I also need to remove from julia-lezer..

pankgeorg:
> That shouldn't be an interpolation even, no? That's a thing I also need to remove from julia-lezer..

You are right - what I do is to skip the highlight, not to fix the parse tree.

fonsp:
Awesome! Still left is to handle interpolation correctly in `@htl`, but maybe @dralletje should fix julia-lezer first

Schermafbeelding 2021-12-06 om 14 44 04

[Stunning notes 12.txt](https://github.com/fonsp/Pluto.jl/files/7661111/Stunning.notes.12.txt)

fonsp:
The same problem with escaping (I am being annoying I know): 

Escaping rules apply on the Julia side, and the result becomes javascript. This is legal JS code, `also_nice` should have the same color as `nice`:

Schermafbeelding 2021-12-06 om 19 31 04

I updated my last comment to include this example

pankgeorg:
I have a quick and quite dirty idea, WIP

dralletje:
HACKS ON HACKS ON HACKS

pankgeorg:
> HACKS ON HACKS ON HACKS

image
there you go

fonsp:
Not a biggie but JS interpolation is only highlighted if I also do Julia interpolation :)

https://user-images.githubusercontent.com/6933510/147076893-c924ca6b-c926-4d7c-ba14-920d4f6e39fc.mov

frontend/imports/CodemirrorPlutoSetup.js
```diff
@@ -42,9 +42,129 @@ import {
...
+    javascript,
+    javascriptLanguage,
+    sql,
+    PostgreSQL,
+    python,
+    pythonLanguage,
     autocomplete,
 } from "https://cdn.jsdelivr.net/gh/JuliaPluto/codemirror-pluto-setup@0.19.4/dist/index.es.min.js"
 
+const htmlParser = htmlLanguage.parser
```
fonsp, line 59: Can you put all additions to this file in a new file? Right now we don't get typings for the code inside this file and that makes it hard to work with. (Did you write this without typings??)

pankgeorg:
> Not a biggie but JS interpolation is only highlighted if I also do Julia interpolation :)

Ooops that's true! fixing!!

PR #1535, opened 2021-10-05
update tips link to open in new tab

trentfridey:
This is just a small patch to open the link from "Still stuck? Here are some __tips__" of the `LiveDocs` component in a new tab. It's just me scratching my own itch, so feel free to close if it is not helpful. Thanks for making Pluto!


pankgeorg:
Looks good to me! Welcome on board 😄

PR #1541, opened 2021-10-05
Reactive `using Module`

### LINKED ISSUE #24: Reactive response to removing `using Module`
If a Module `export`s some variables, those will be brought into the workspace when a `using Module` call is somewhere, and they will disappear when the `using` call is gone. 

However, removing a `using Module` does not reactively reevaluate cells that depend on its exported variables. Those cells need manual reevalutaion.

Requires Pluto to know exports.

Pangoraw:
Fixes #24


Pangoraw:
Why is there no CI run here ? :thinking:

PR #1544, opened 2021-10-06
Isolated cell view

ctrekker:
Isolates a single cell output to occupy the entire page. Useful for building web applications / dashboards with Pluto


fonsp (approved):
API looks future-proof, we could always change the implementation later

Issue #1540, opened 2021-10-05
Rectangle select


fonsp:
We used to have it with Option/Alt + Click drag

PR #1545, opened 2021-10-06
Add rectangularSelection to list of extensions

ctrekker:
This PR enables rectangular selection in CM6 code inputs, addressing #1540.

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -30,6 +30,7 @@ import {
     indentOnInput,
     defaultHighlightStyle,
     closeBrackets,
+    rectangularSelection,
     highlightSelectionMatches,
     closeBracketsKeymap,
     searchKeymap,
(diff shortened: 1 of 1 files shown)


fonsp:
https://user-images.githubusercontent.com/6933510/136282349-1d4e1352-c7b5-44ec-87f2-4ba3657f388a.mov

Because we moved add-cursor to Opt/Alt instead of Cmd/Ctrl, using Alt/Opt to rectangle select will leave the old cursor....... @pankgeorg any ideas?

ctrekker:
We could switch rectangle select to Opt/Alt + Shift + drag instead of Opt/Alt + drag. This is what rectangle select is bound to in VSCode. Otherwise I do have a solution that makes Opt/Alt work for both rectangle select and multi-select, but it makes the cursor do weird things if you hold the mouse down instead of click.

dralletje:
Fons zie je dit?

ctrekker:
In 0.16.1 a crosshair appears whenever the user holds alt. Is this something worth implementing in CM6?

PR #1561, opened 2021-10-14
Use parcel to bundle

### LINKED ISSUE #241: Artifacts to proxy&cache CDNs

fonsp:
Fix #241 

This PR is by @dralletje 

### TODO
- [x] Fonts seem to now work?
- [x] Import firebase on demand as modules (for the feedback form at the bottom)
- [x] Github Action to automatically deploy to release branch


dralletje:
!!!

uPvtkBlpiucvDUf:
It seems this doesn't work for some .svg files loaded in `frontend/components/PkgPopup.js`, see e.g.

https://github.com/fonsp/Pluto.jl/blob/release/frontend-dist/editor.77d0234b.js#L423

which is loaded from jsdelivr even in `frontend-dist/` in the `release` branch. Or am I missing something?

dralletje:
I knew about the pkg popup svg's (those are inside the preact code and no parsing for that yet in parcel)

Thanks for the iframe resize, didn't think of that... I'll make an issue with all the frontend edge cases

PR #1577, opened 2021-10-19
Fix wrong environment for using in Pkg.activate notebooks

### LINKED ISSUE #1511: Pkg activate fails to load the correct versions of the packages
I am experiencing a very annoying and difficult-to-debug behaviour when updating to 0.16.0 or 0.16.1:
In [my course](https://github.com/jbrea/MLCourse) I do not use the Pluto pkg manager but ship a custom Manifest file.
Each notebook starts with `using Pkg; Pkg.activate("MLCourse")`. Until Pluto version 0.15.1 this works fine.
With the new Pluto versions it seems other versions of the packages are loaded.
The weird thing is: `Pkg.status()` from within the Pluto notebook shows the correct versions (in the REPL),
but `@which somefunction()` points me to other versions of the packages, e.g. [in this notebook](https://github.com/jbrea/MLCourse/blob/main/notebooks/supervised_learning.jl) I expect version v1.1.0 of MLJOpenML to be loaded but `@which MLJOpenML(554)` points to v1.0.0 of MLJOpenML.

With the pop-up saying that there is a new version of Pluto the students are really tempted to update their Pluto version, just to find that after the update the notebooks don't work anymore...

fonsp:
Fix #1511 , by @Pangoraw

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -39,4 +39,4 @@ jobs:
 
             # 🚗
             - uses: julia-actions/julia-runtest@v1
-              continue-on-error: ${{ matrix.julia-version == 'nightly' || matrix.os == 'windows-latest' }}
+              continue-on-error: ${{ matrix.julia-version == 'nightly' }}


fonsp:
This fix was backported to 0.16.2 using this branch: https://github.com/fonsp/Pluto.jl/pull/1576

Registrator commit here: https://github.com/fonsp/Pluto.jl/commit/ca9dc1ff6110ee7a23b12682d849cdb4abcce27a

Omar-Elrefaei:
Thanks @fonsp !

PR #1597, opened 2021-10-25
Update to macro runtime & PlutoHooks.jl

Pangoraw:
## The companion package is [PlutoHooks.jl](https://github.com/JuliaPluto/PlutoHooks.jl)

 - [x] indirectly/directly updating the definition of a macro updates the nodes of all macros calling that macro
 - [x] directly running a macrocall updates the reactive node for that macrocall
 - [x] indirectly running a macrocall does not updates the reactive node for that macrocall
 - [x] Change computer lifecycle to use cell UUIDs instead of expression hash
 - [x] PlutoRunner hooks for cell related cleanups when the computer is removed
 - [x] PlutoRunner hooks to rerun cell implicitely
 - [x] can't use `:(function() end)` as macro outputs :cry: 
 - [x] work without function wrapping

Reactive node update also means update of the underlying function wrapper.

## Function wrapping

Cells with macro calls will be function wrapped. that means at list two things:

#### Stateful cells in Julia (similar to `this` in JavaScript)
wrapped_macrocall

<details>
<summary>Other example</summary>

wrapped_macro_acc

</details>

#### Faster execution because of a single compilation
macro_wrapped_htl

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -144,50 +144,11 @@ function sanitize_expr(expr::Expr)
     Expr(expr.head, sanitize_expr.(expr.args)...)
 end
 
-# a function as part of an Expr is most likely a closure
-# returned from a macro
-# function sanitize_expr(func::Function)
-#     mt = typeof(func).name.mt
-#     GlobalRef(mt.module, mt.name) |> sanitize_expr
-# end
-
-# function sanitize_expr(union_all::UnionAll)
-#     sanitize_expr(union_all.body)
-# end
-
-# function sanitize_expr(vec::AbstractVector)
-#     Expr(:vect, sanitize_value.(vec)...)
-# end
-
-# function sanitize_expr(tuple::Tuple)
-#     Expr(:tuple, sanitize_value.(tuple)...)
-# end
-
-# function sanitize_expr(dict::Dict)
-#     Expr(:call, :Dict, (sanitize_value(pair) for pair in dict)...)
-# end
-
-# function sanitize_expr(pair::Pair)
-#     Expr(:call, :(=>), sanitize_value(pair.first), sanitize_value(pair.second))
-# end
-
-# function sanitize_expr(set::Set)
-#     Expr(:call, :Set, Expr(:vect, sanitize_value.(set)...))
-# end
-
-# function sanitize_expr(mod::Module)
-#     fullname(mod) |> wrap_dot
-# end
-
-# An instanciation of a struct as part of an Expr
-# will not de-serializable in the Pluto process, only send if it is a child of PlutoRunner, Base or Core
-function sanitize_expr(other)
-    # typename = other |> typeof
-    # typename |> parentmodule |> Symbol ∈ [:Core, :PlutoRunner, :Base] ?
-    #     other :
-    #     Symbol(typename)
-    nothing
-end
+sanitize_expr(linenumbernode::LineNumberNode) = linenumbernode
+sanitize_expr(bool::Bool) = bool
+# In all cases of more complex objects, we just don't send it.
+# It's not like the expression explorer will look into them at all.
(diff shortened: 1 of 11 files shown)


rikhuijzer:
> The `@ingredients` macro is working well with #1541 so this can be used to reactively develop source file/notebooks
> 
> reactive_using .

Looks really cool. Does this work with packages too or just source files?

Pangoraw:
> > The `@ingredients` macro is working well with #1541 so this can be used to reactively develop source file/notebooks
> 
> Looks really cool. Does this work with packages too or just source files?

Just with individual source files for now, but something that works on a package level with incremental re-definitions (using the work made on Revise.jl) is something that is definitely on the roadmap of the PlutoHooks "ecosystem". It requires to answer questions about package environment management from macros first.

src/runner/PlutoRunner.jl
```diff
@@ -340,7 +426,14 @@ If the third argument is a `Tuple{Set{Symbol}, Set{Symbol}}` containing the refe
 
 This function is memoized: running the same expression a second time will simply call the same generated function again. This is much faster than evaluating the expression, because the function only needs to be Julia-compiled once. See https://github.com/fonsp/Pluto.jl/pull/720
 """
-function run_expression(m::Module, expr::Any, cell_id::UUID, function_wrapped_info::Union{Nothing,Tuple{Set{Symbol},Set{Symbol}}}=nothing, contains_user_defined_macrocalls::Bool=false)
+function run_expression(m::Module, expr::Any, cell_id::UUID, function_wrapped_info::Union{Nothing,Tuple{Set{Symbol},Set{Symbol}}}=nothing, forced_expr_id::Union{ObjectID,Nothing}=nothing; user_requested_run::Bool=true)
+    if user_requested_run
+        # TODO Time elapsed? Possibly relays errors in cleanup function?
+        UseEffectCleanups.trigger_cleanup(cell_id)
+
+        # TODO Could also put explicit `try_macroexpand` here, to make clear that user_requested_run => fresh macro identity
```
Pangoraw, line 434: no_explicit_rerun

There seems to be a problem with this, explicitely re-running does not regenerate the underlying expression. The same goes for updating the macro definition.

dralletje: Hmm, I don't have that D:

You didn't have that before? Maybe it's time I put the explicit re-macroexpand-ization there anyway then.

Pangoraw: The following condition is alway false in my example:

https://github.com/fonsp/Pluto.jl/blob/5c34b99d756e7d77d624f3155b7e9007ac1b3cc4/src/runner/PlutoRunner.jl#L453

which doesn't trigger a new expansion.

dralletje: Yeah, it should, as that *should* already be handled by macro_expand from the Expression Explorer.. Wonder why it isn't for you though D:

Pangoraw: ah yes you're right

Pangoraw: okay weird, i haven't been able to reproduce

fonsp:
@Pangoraw Can you put the code for [these examples](https://github.com/fonsp/Pluto.jl/pull/1597#issuecomment-953872471) online? I want to know how to make these things!

Pangoraw:
- Self re-rendering example: i don't have the notebook under my belt right now but the important part of the code is still visible in the gif, note that depending on the self-updating cell will re-run downstream cell even when the job is not done so doing progress with #437 will still be easier.
 - File watching: `@use_file` in PlutoHooks
 - Reactive ingredients: `@ingredients` in PlutoHooks
 - Live HTTP server response editing: currently lives in https://github.com/JuliaPluto/PlutoWebServer.jl but i have not updated it to account for the different changes yet so it might not work currently.

PR #1600, opened 2021-10-26
API for widgets to return possible bind values

Pangoraw:
See:
 - https://github.com/JuliaPluto/PlutoSliderServer.jl/pull/29
 - https://github.com/JuliaPluto/PlutoUI.jl/pull/147

(We will merge this after the functionality in PlutoSliderServer is finished)


fonsp:
Can use the new stuff from #1612 !

fonsp:
I just realised: the possible bind values should stay in the notebook process, because they might not be serializable...... You only want the outside process to know the length, and to say e.g. "set bound variable x to the 17th possible value". Will come back to this next week!

Pangoraw:
I think the collected values should correspond to the JavaScript possible values so they should be serializable. If we collect Julia values then the Message pack hashes for the files will be wrong. The Julia can still be found by using the js->jl transform. Also see https://github.com/JuliaPluto/PlutoUI.jl/pull/152.

fonsp:
Totally right! I should change this https://docs.juliahub.com/AbstractPlutoDingetjes/UHbnu/1.1.0/#:~:text=If%20you%20are%20also%20using%20transform_value%20for%20your%20widget%2C%20then%20the%20values%20returned%20by%20possible_values%20should%20be%20the%20values%20after%20transformation.

src/runner/PlutoRunner.jl
```diff
@@ -1610,6 +1612,16 @@ function transform_bond_value(s::Symbol, value_from_js)
     end
 end
 
+function possible_bond_values(s::Symbol)
+    element = get(registered_bond_elements, s, nothing)
+    return try
+        possible_bond_values_ref[](element)
+    catch e
+        @error "Could not collect possible bond values for $s" exception=(e, catch_backtrace())
```
fonsp, line 1620: Why the try catch rethrow?

Pangoraw: i think i tried to much to make it look like `transform_bond_value`, this does not seem necessary

Pangoraw: hmm this can actually be helpful to get the name of the symbol s that failed when developping ?

fonsp:
TODO: there two values still need to be handled properly:
https://docs.juliahub.com/AbstractPlutoDingetjes/UHbnu/1.1.1/#AbstractPlutoDingetjes.Bonds.InfinitePossibilities
https://docs.juliahub.com/AbstractPlutoDingetjes/UHbnu/1.1.1/#AbstractPlutoDingetjes.Bonds.NotGiven

I'll do this myself next week 💆

Issue #1639, opened 2021-11-05
FilePicker does not support paths surrounded by double/single quote


disberd:
The autocomplete on filepicker is neat but sometime it is more convenient to directly paste in a path copied from somewhere to open a notebook.

In Windows, paths copied from shift+right-click on a file are surrounded by double quotes (they are at least for me as I have a HOME directory with a space in the middle that I can't change).

The current implementation requries me to manually delete the double quotes from the beginning and end of the path, otherwise the filename is searched with the double-quotes.

Will removing leading and trailing double or single quote from path break anything or can it be done?

PR #1640, opened 2021-11-05
Remove trailing and leading single/double quotes from path

disberd:
Fixes https://github.com/fonsp/Pluto.jl/issues/1639

Should not break existing valid names containing quotes as it only remove the trailing and leading character if there is a quote after the notebook extension

### DIFF
--- a/frontend/components/Welcome.js
+++ b/frontend/components/Welcome.js
@@ -74,6 +74,11 @@ export const process_path_or_url = async (path_or_url) => {
             path_or_url: u.href,
         }
     } catch (ex) {
+        /* Remove eventual double quotes from the path */
+        const last_char = path_or_url[path_or_url.length-1]
+        if (last_char == '"' || last_char == "'") {
+            path_or_url = path_or_url.replace(/^['"]/,"").replace(/['"]$/,"")
+        } 
         return {
             type: "path",
             path_or_url: path_or_url,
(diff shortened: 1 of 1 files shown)


dralletje:
A bit niche, but sure..

But can you make it very strictly only work when both the first and last character are double quotes?
And also add a comment referring to the issue you made.

disberd:
Done, I also kept single quote as option as I suppose there are cases where the path are pasted surround by them but I can remove them if you think they are breaking something.

dralletje:
It's not about breaking stuff, but about keeping the code we insert for an exception really simple and easy to follow (e.g. `path_or_url[0] === '"' && last_char === '"'` vs whatever there is now). I know the difference seems silly, but having an special handling like this really needn't be something we do often :P

Also can you use `.slice` instead of `.substring`? That way you can write `.slice(1, -1)` 😎

Also also, can you run prettier? Now if I open the file I get a diff instantly because of the formatting

disberd:
Makes sense, I have the tendency of try to take into account too many un-necessary corner cases...

Thanks for the `slice` suggestion, my Javascript expertise always relies on google what I want to achieve :D.

I removed the definition of last_char since probably is clear enough with indexing the string directly. 
Let me know if that is OK.

I didn't have any prettier as I started this edit directly on the web github interface but this time I pulled this branch on VSCode to prettify.

dralletje:
Thanks a lot! Sorry for making you go through 2 edits :P

PR #1644, opened 2021-11-06
Add docstrings to SessionActions

rikhuijzer:
Given that these methods in SessionActions should be relatively stable (source: https://github.com/rikhuijzer/PlutoStaticHTML.jl/issues/5), it would be a good idea to add some documentation. It took me a while to figure out what these methods do.

Feel free to commit directly into this PR if I summarized the method functionality incorrectly.


fonsp:
Thanks!

PR #1666, opened 2021-11-15
make soft edges break on dependencies cycle

### LINKED ISSUE #1579: Another cyclic reference error
Hi, I have a notebook developing some functionalities that started experiencing cyclic references error in the unreleased version of Pluto (current `main` and some previous commits after last release).
Code was working fine in v0.16.1 and still does in latest Pluto version.

Here is the link to the notebook:
https://github.com/disberd/TelecomUtils.jl/blob/fe47aae23675d5bf7c204621c3bf5a4ff2ee5812/src/SatView.jl

This is not a super short notebook as I didn't manage to reproduce this in a very short notebook example.

Apart from a lot of commentend cells in the notebook and some errors due to dependencies on packets that appear commented in the notebook, the main issue is between cells defining the function `_rotation_matrix`, where one method defines a rotation matrix and another method calls the inverse of the rotation matrix through `inv`:
image

and some other cells defining custom transformation structures that add methods to `Base.inv` like this one:
image

I tried putting just those two cells inside a notebook on its own but I don't get the cyclic reference error so there must be something more going on.
The fact is that the same notebook works fine on Pluto 0.16.2 so probably it's something related to the new expression explorer?

### LINKED ISSUE #1622: Cyclic reference for self-recursive A.b methods
Neither should error:

Schermafbeelding 2021-11-01 om 21 18 28

```julia
# ╔═╡ 37e09898-5771-4731-976a-40f6f8dae13a
Base.sqrt(x::Vector) = Base.sqrt("1")

# ╔═╡ 89ff3f12-464a-46dd-a431-a6d13d8f1a81
Base.sqrt(x::String) = Base.sqrt("1")

# ╔═╡ 1d7f81ae-3e57-4227-8a39-78767aeec35d
f(x::Vector) = f("1")

# ╔═╡ 0d506751-811e-4850-a749-a65cb2724979
f(x::String) = f(123)

```

# `Base.show`

It is important that these don't error, because it is common for a `Base.show` overload to use `Base.show` in its definition, for example in PlutoUI:
```julia
function Base.show(io::IO, m::MIME"text/html", slider::Slider)
	x = slider.asdf
	Base.show(io, m, @htl("""
		<input type=range max=$(x)>
	"""))
end
```

The current workaround is to define a method for `Base.show`, but use `show` (without `Base.`) in the function body. Fortunately we were already doing this in the PlutoUI source code :)

Pangoraw:
This prevents soft definitions from creating dependencies cycles by "breaking" the soft definition dependency if a cycle is created.

<details>

soft_cycle

</details>

This currently needs more edge cases tests.

Fixes #1579, fixes #1622

### DIFF
--- a/src/analysis/Errors.jl
+++ b/src/analysis/Errors.jl
@@ -9,8 +9,8 @@ end
 
 function CyclicReferenceError(topology::NotebookTopology, cycle::Cell...)
 	referenced_during_cycle = union((topology.nodes[c].references for c in cycle)...)
-	assigned_during_cycle = union((topology.nodes[c].definitions ∪ topology.nodes[c].funcdefs_without_signatures for c in cycle)...)
-	
+	assigned_during_cycle = union((topology.nodes[c].definitions ∪ topology.nodes[c].soft_definitions ∪ topology.nodes[c].funcdefs_without_signatures for c in cycle)...)
+
 	CyclicReferenceError(referenced_during_cycle ∩ assigned_during_cycle)
 end
 
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -530,24 +530,26 @@ function explore!(ex::Expr, scopestate::ScopeState)::SymbolsState
         innerscopestate.inglobalscope = false
 
         funcname, innersymstate = explore_funcdef!(funcroot, innerscopestate)
-        # Macro are called using @funcname, but defined with funcname. We need to change that in our scopestate
-        # (The `!= 0` is for when the function named couldn't be parsed)
-        if ex.head == :macro && length(funcname) != 0
-            setdiff!(innerscopestate.hiddenglobals, funcname)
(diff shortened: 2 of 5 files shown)


fonsp:
I'm writing a test for #1622 right now by coincidence :)

Pangoraw:
> I'm writing a test for #1622 right now by coincidence :)

Nice! :smiley:

fonsp:
Lots of tests added in https://github.com/fonsp/Pluto.jl/commit/1c202245bfe4d7c8c003e4d77969484d1713c695 , try those. I am also working on https://github.com/fonsp/Pluto.jl/pull/1667 right now, which will make all of these broken tests pass, but it might be too hacky...

fonsp:
@Pangoraw Can you take a look at my PR and if it looks good, combine the two into one super cool PR?

Pangoraw:
#1667 looks super nice! I will combine them since they both cover slightly different scopes.

Pangoraw:
Closed in favor of #1668

Issue #178, opened 2020-06-11
Cyclic references between methods in cells


karlwessel:
At the moment Pluto considers the following example as cyclic references between two cells:

Cell A:
```
begin
	constructor(a) = a
	constructor(a, N) = refine(constructor(a), N)
end
```

and Cell B
```
refine(a, N) = constructor(a) * N
```

According to fonsps comment in issue #177 that should not be a cyclic reference.

karlwessel:
Another MWE where this happen.
First cell:
```
f(x) = g(x+2)
```
Second cell:
```
g(x) = x<10 ? f(x+1) : x
```

Is there some easy fix or workaround for this problem? This basically doesn't allow you to implement certain recursive function calls.

drdozer:
I also hit this, where I have a complex structure where parts need to know the whole during initialization. I get cyclic references between the initialization of the whole and the initialization of the parts.

PR #1667, opened 2021-11-15
😵‍💫 Allow cyclic recursive functions

fonsp:
Pretty simple fix for #178: if all variables in the cycle are the names of functions, then allow the cycle. 

It sounds too simple so maybe I'm missing something...


Pangoraw:
Closed in favor of #1668

PR #1676, opened 2021-11-17
Fix world age issues causing 404 on master

Keno:
The HTTP docs caution:
```
Note that due to being a macro (and the internal routing functionality), routes can only be registered
statically, i.e. at the top level of a module, and not dynamically, i.e. inside a function.
```

Pluto ignores this restriction causing 404 issues on nightly (#1672).
Before 1.8, Pluto was relying on undefined behavior causing tasks
to potentially start in a newer world than their definition point,
making this work by accident. This is the minimal fix to restore
Pluto to good functioning, but it's probably worth taking another
look at whether generating new routes every time this method is
invoked is really the intended behavior.


fonsp:
Hi Keno, that's amazing! How did you find the source of the bug? How come the error messages were hidden?

Keno:
> How did you find the source of the bug? How come the error messages were hidden?

There is no error message - the routes were just too new, so the routing fell back to the default which gives an empty 404.

PR #1683, opened 2021-11-23
Launch parameter to insert HTML into the top of the notebook

fonsp:
This might make the setup for https://computationalthinking.mit.edu/ simpler

Schermafbeelding 2021-11-23 om 14 01 14

https://github.com/fonsp/disorganised-mess/blob/main/custom-preable-demo.jl


fonsp:
beautiful

PR #1687, opened 2021-11-25
PlutoHTMLElement and PlutoHTMLElementList

dralletje:


fonsp:
@dralletje Can you make a TODO list?

dralletje:
From @fonsp during a call: Why are there two primitives rather than one.
This is a good question, I'll ponder it a bit. I still think we need two primitives, but they might be different ones.

fonsp:
Can you make the changes to the preact import on main instead?

PR #1707, opened 2021-11-30
remove time warning

AshtonSBradley:
This no longer seems necessary... ?

On 1.6 it is almost immediate for me, after fresh install (call it 20 seconds on a 2019 iMac). The warning suggests caution that isn't warranted.


fonsp:
Yayyy you're right! Awesome!

PR #1719, opened 2021-12-01
Global storage for published objects

fonsp:
If two cells use `publish_to_js` on `===`-exactly the same object, it will now only be sent to the frontend once :)

This is useful for https://fonsp-disorganised-mess.netlify.app/importing%20local%20js%20modules


fonsp:
why not! yolo

PR #1721, opened 2021-12-02
Don't resend published objects from PlutoRunner if unchanged

fonsp:
Idea from @dralletje :

We diff the published objects between server and frontend, that's great, but we still send it back from PlutoRunner to the server via distributed every time the cell evaluates, making it less suitable for "longish-term storage". Fixed now!

See also #1719 

This is useful for https://fonsp-disorganised-mess.netlify.app/importing%20local%20js%20modules


fonsp (commented):
great stuff wow

PR #1729, opened 2021-12-04
Sync local state (PR to master)

pankgeorg:


pankgeorg:
There seems to be an infinite loop when autocomplete is triggered with this PR, that's why front-end tests fail. Investigating.

fonsp:
What is `time_arrow`? It looks like we only assign to it, but never use the value, so we can remove it?

fonsp:
Can you run this with
```julia
Pluto.run(simulated_lag=1.0)
```

and see if it works?

pankgeorg:
> and see if it works?

It does work - you can "break" it if you start typing in an intermediate state but the fix for that is something much more complicated.

frontend/components/Editor.js
```diff
@@ -318,7 +319,7 @@ export class Editor extends Component {
                         state.selected_cells = []
 
                         for (let cell of new_cells) {
-                            state.cell_inputs_local[cell.cell_id] = cell
+                            state.notebook.cell_inputs[cell.cell_id] = cell
```
dralletje, line 322: Shouldn't this be in an `update_notebook()` call as well, for it to be sent to the other clients?

frontend/components/Editor.js
```diff
@@ -349,10 +351,10 @@ export class Editor extends Component {
 
                 await this.setStatePromise(
                     immer((state) => {
-                        state.cell_inputs_local[cell_id] = {
+                        state.notebook.cell_inputs[cell_id] = {
```
dralletje, line 354: `update_notebook()`?

frontend/components/Editor.js
```diff
@@ -378,7 +381,7 @@ export class Editor extends Component {
                 this.setState(
                     immer((state) => {
                         for (let cell of cells_to_add) {
-                            state.cell_inputs_local[cell.cell_id] = cell
+                            state.notebook.cell_inputs[cell.cell_id] = cell
```
dralletje, line 384: `update_notebook()`?

frontend/components/Editor.js
```diff
@@ -489,8 +493,8 @@ export class Editor extends Component {
                 if (cell_ids.length > 0) {
                     await update_notebook((notebook) => {
                         for (let cell_id of cell_ids) {
-                            if (this.state.cell_inputs_local[cell_id]) {
-                                notebook.cell_inputs[cell_id].code = this.state.cell_inputs_local[cell_id].code
+                            if (this.state.notebook.cell_inputs[cell_id]?.local_code) {
```
dralletje, line 496: Now this is also checking the notebook instead of some separate state, it's better to use the notebook parameter to check against (instead of `this.state` which could possibly have an older/different version of the state once synchronisation becomes more complex)

frontend/components/CellInput.js
```diff
@@ -148,27 +152,29 @@ export const CellInput = ({
     variables_in_all_notebook,
 }) => {
     let pluto_actions = useContext(PlutoContext)
-
+    const client_id = pluto_actions.get_client_id()
```
dralletje, line 155: Can we make this a property instead of a function? Can still be on `pluto_actions` and added in an useMemo (e.g. `pluto_actions_with_client_id = React.useMemo(() => ({ ...pluto_actions, client_id: client_id }), [pluto_actions, client_id])` in Editor.js), feels off to have this part not participate in the React reactivity stuff

frontend/components/CellInput.js
```diff
@@ -55,6 +56,8 @@ import { pluto_paste_plugin } from "./CellInput/pluto_paste_plugin.js"
 import { bracketMatching } from "./CellInput/block_matcher_plugin.js"
 import { cl } from "../common/ClassTable.js"
 
+const change_is_from_remote_annotation = Annotation.define()
```
dralletje, line 59: Nice!

dralletje (changes requested):

fonsp:
Refreshing is glitchy:

https://user-images.githubusercontent.com/6933510/144913250-713bacfd-db28-49f1-bb17-7fd2465758a2.mov


This also looks like the new logic to check if the cell changed (display in dark yellow) is not always correct, because both sides show the cell as unchanged after the refresh

fonsp:
Can you call it `my_author_name` instead of `client_id`? We already have a `client_id` in our websocket logic, but they are not the same (and it's good to not use the same because websocket logic should be as isolated as possible)

frontend/components/Editor.js
pankgeorg: this is actually done a bit below, so I'm just removing this completely

fonsp:
@pankgeorg How ready is this?

pankgeorg:
> @pankgeorg How ready is this?

When you have lag and two people are typing, it is possible to get to a state when you see something different than the other client, as a 'final' state of the document, as the two states don't converge to the 'final' for some reason. I'm not sure if this is fixable by our current approach; I'll investigate more. It works like a charm when only one party is making edits though

fonsp:
https://gist.github.com/fonsp/34dcc52e853baf512341034ede911f3f

pankgeorg:
Issue: When two people edit the same cell, the cursor goes back to the first line.
Options: keep your cursor at the same position as before (if it exists) OR move the cursor along with the last used cursor (which re	quires syncing the cursor too)

fonsp:
> Issue: When two people edit the same cell, the cursor goes back to the first line. Options: keep your cursor at the same position as before (if it exists) OR move the cursor along with the last used cursor (which re quires syncing the cursor too)

I thought we were only going to allow editing by one person at a time? Disabling input for others

fonsp:
First crash (browser freezed):

Schermafbeelding 2022-01-27 om 18 21 58

fonsp:
Second crash (browser freezed):

Schermafbeelding 2022-01-27 om 18 23 04

The error is probably fine, but you're not supposed to get that warning

dralletje:
@pankgeorg looking at this now.. 😏

fonsp:
> https://github.com/fonsp/Pluto.jl/pull/1729/commits/f3fce3e74ac95c4384a52a788c1ff2ff326a11f6

This is great functionality but it really should not have been in this PR... We just want to get this PR done so that we can merge it, this makes that more difficult

dralletje:
Grrrrr

But you make a good point

PR #1736, opened 2021-12-06
Remove `is_first_value` from state

fonsp:


fonsp:
@dralletje I'm merging this but it would be nice if you could take a look anyways!

PR #1745, opened 2021-12-07
Get reverse immer patches from applying patches (immer patches reverse immer applyWithPatches reverse)

fonsp:
Get reverse immer patches from applying patches (immer patches reverse immer applyWithPatches reverse) and apply them to the old state (reverse immer patches) to get the old state (immer patches) to apply the reverse immer patches (immer patches reverse) to get the new state (immer patches)


fonsp:
@dralletje I'm merging this but feel free to take a look at why I need to create the new notebook struct a second time. Set `get_reverse_patches` to `true` by default to test it easily without creating a recording and a local file server.

PR #1748, opened 2021-12-08
Holding a key will stop at cell boundaries again, fixes #1747

### LINKED ISSUE #1747: Holding arrow up jumps between all cells
This used to be debounced, and holding arrow up should not allow you to continue jumping across cells

dralletje:


fonsp:
jaaa een pr yessss xoxo

dralletje:
To be fair, only reason it is an PR is because I don't know how rebasing and such work but I'll take the credit anyway

PR #1749, opened 2021-12-08
Simplify autocomplete with path and dict key support, fixes #1718

### LINKED ISSUE #1718: Autocompletion issues after migration to CM6
As discussed on [zulip](https://julialang.zulipchat.com/#narrow/stream/243342-pluto.2Ejl/topic/Autocomplete.20dict.20keys)

There are some cases where autocompletion used to work before CM6 and does not anymore:
- Dict keys (`dict[<TAB>`)
- apparently also paths `"path/completeion/<TAB>"` (as mentioned by @fonsp)

dralletje:
Removes some, _I think_ unnecessary stuff I thought we needed from how the previous autocomplete worked...
But not sure now if we ever needed it..

This now works for `dict[`<kbd>TAB</kbd> and `"./`<kbd>TAB</kbd>.
(Also because it would make the autocompletes in the list look weird, don't want to have a completion `MyFile.jl"`, so maybe a way to "silently" insert it? 🤷‍♀️)

Haven't tested it extensively with normal autocompletions, so will do that a bit more.

~Ideally I want the dict and path tabs to insert the closing `]` and `"` respectively, but haven't added that yet.~ Shouldn't be necessary as the `]` (and soon the `"`???) gets inserted automatically by bracket matching.

Next up, deep autocomplete (`dict[:subdict][`<kbd>TAB</kbd>)

One more thing I want to add, possibly in this PR: when you autocomplete a path to a directory, the autocomplete should pop up again directly with the entries in that directory. (But this PR is fine to be merged without that too)

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -22,7 +22,6 @@ import {
     indentLess,
     tags,
     HighlightStyle,
-    autocompletion,
     lineNumbers,
     highlightSpecialChars,
     foldGutter,
@@ -36,7 +35,6 @@ import {
     searchKeymap,
     foldKeymap,
     commentKeymap,
-    completionKeymap,
     syntaxTree,
     Decoration,
     ViewUpdate,
@@ -45,6 +43,7 @@ import {
     indentUnit,
     StateField,
     StateEffect,
+    autocomplete,
 } from "../imports/CodemirrorPlutoSetup.js"
 import { pluto_autocomplete } from "./CellInput/pluto_autocomplete.js"
 import { NotebookpackagesFacet, pkgBubblePlugin } from "./CellInput/pkg_bubble_plugin.js"
@@ -190,7 +189,7 @@ export const CellInput = ({
             return true
         }
 
-        let select_autocomplete_command = completionKeymap.find((keybinding) => keybinding.key === "Enter")
+        let select_autocomplete_command = autocomplete.completionKeymap.find((keybinding) => keybinding.key === "Enter")
         let keyMapTab = (/** @type {EditorView} */ cm) => {
             // This will return true if the autocomplete select popup is open
             if (select_autocomplete_command.run(cm)) {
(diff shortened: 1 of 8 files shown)


disberd:
Cool!
The nested dict auto-completion would be quite nice as FuzzyCompletions [does not support it](https://github.com/JunoLab/FuzzyCompletions.jl/issues/9).
If you have to add something for its support, this might probably also used to autocomplete Dictionaries from [Dictionaries.jl](https://github.com/andyferris/Dictionaries.jl) which are quite handy and more suited to some kinds of data manipulation than Dicts.

dralletje:
> If you have to add something for its support, this might probably also used to autocomplete Dictionaries from Dictionaries.jl which are quite handy and more suited to some kinds of data manipulation than Dicts.

This is because `Dictionaries.Dictionairy` isn't a subtype of `AbstractDict` :(
Not sure why, guess they have good reasons, but this should either be fixed by an exception in FuzzyCompletions specifically for Dictionaries.jl, or by having Dictionaries.AbstractDictionary be a subtype of AbstractDict.. but don't think that will happen

disberd:
Yeah indeed the Dictionary <: AbstractDict will most likely never happen as I believe it was a design choice.
I just thought you had to somehow extend some things on top of FuzzyCompletions since that also doesn't support nested dicts, and if an extension was happening maybe it could also work for Dictionary :D

dralletje:
> I just thought you had to somehow extend some things on top of FuzzyCompletions since that also doesn't support nested dicts, and if an extension was happening maybe it could also work for Dictionary :D

Yeah.. after looking at FuzzyCompletions I feel like forking it anyway (I think autocomplete is too important not to be totally Pluto-customised), so might actually be nice to also include Dictionary.jl support then

dralletje:
Okayyy path autocompleting stuff is in!
I think this is fine for merging, @fonsp

dralletje:
Alright keep adding more stuff, will merge this tonight

PR #1755, opened 2021-12-10
JS API to get/set input value of element like Pluto

fonsp:
We have some special logic to get/set the input value of an element:

https://github.com/fonsp/Pluto.jl/blob/2cbc3813f1199795fc8a5b3a58ad60bb5d5111cd/frontend/common/Bond.js#L7-L35

https://github.com/fonsp/Pluto.jl/blob/2cbc3813f1199795fc8a5b3a58ad60bb5d5111cd/frontend/common/Bond.js#L76-L123

The first one (`get`) mostly matches the logic used by `Generator.input` from the observablehq stdlib, but this function can only return an async generator, but the second one (`set`) is not exposed as API.

I want to make some higher-level components, like:
- `combine` that combines multiple inputs into a single JS variable
- `confirm` that holds the value of an element until a button is pressed
- `throttle`
- `debounce`
- `with_value`

For these to work well (with sync across clients) and stable (future proof, in case we change those functions), I need to use the same functions inside JS code that Pluto uses.

I think this API is easy to support forever, because they are just two small, isolated functions.


fonsp:
@dralletje

PR #1761, opened 2021-12-14
Slightly tweak @htl intro in JS sample notebook

icweaver:
As a total noob to web programming, I thought these slight changes helped me follow the logical flow of the text a bit better. Pretty much whenever new topics are introduced, I like following the rule of thumb of just addressing it up front (e.g. with hyperlinks or whatever), and then calling back to it when it's mentioned for the second time.

No idea if that makes sense, but just thought I'd bring this up to see what folks thought!

### DIFF
--- a/sample/JavaScript.jl
+++ b/sample/JavaScript.jl
@@ -102,9 +102,9 @@ md"""
 
 ## Using HTML, CSS and JavaScript
 
-To use web languages inside Pluto, we recomend the small package `HypertextLiteral.jl`, which provides a `@htl` macro.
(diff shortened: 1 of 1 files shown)


dralletje:
Damn I merged the previous pull request so this one now doesn't merge nicely..
Can you re-do this based on current main? 😅

icweaver:
Oh yea, that is my b. Sorry about that, should be fixed now! (I hope)

fonsp:
Thank you! I forgot to mention: we have to create and edit the sample notebooks in Julia 1.6, because of the manifest changes in 1.7. 

A manifest created in 1.6 has the widest compatibility across Julia verisons, although it's not a big deal if the manifest does not work, in that case Pluto will create a new one (but it takes a bit longer to load).

PR #1765, opened 2021-12-16
Make document currentscript work

dralletje:
Instead of eval-ing the code from javascript, I now insert a new script element that gets actually executed (so it becomes the `document.currentScript`). It won't keep `currentScript` after anything async, but that also doesn't work with normal scripts either, so I think this just raises compatibility a bunch.

Maybe even deprecate naked `currentScript` in favor of `document.currentScript`? but I don't care much either way.

Supersedes #1522


fonsp:
Uh oh, I think I wrote some scripts that use `currentScript` in async stuff

dralletje:
Yeah so we might first want to deprecate it if we remove it. You can still do `let currentScript = window.currentScript`.

But this PR doesn't remove `currentScript`.

frontend/components/CellOutput.js
```diff
@@ -282,16 +306,19 @@ const execute_scripttags = async ({ root_node, script_nodes, previous_results_ma
...
-                            invalidation: invalidation,
-                            getPublishedObject: (id) => cell.getPublishedObject(id),
-                            ...observablehq_for_cells,
-                        },
-                        code: node.innerText,
+                    let { node: new_node, result } = await execute_inside_script_tag_that_replace(node, async () => {
+                        return await execute_dynamic_function({
+                            environment: {
+                                this: script_id ? old_result : window,
+                                currentScript: document.currentScript,
```
fonsp, line 313: if we leave this at `node`, it will still work in async things, right?

fonsp: Because right now it would break async stuff?

dralletje: Sorry for the confusion, the "might remove currentScript" in my original message was a suggestion. The `currentScript` variable works exactly as it did before, with and without async.

fonsp: Nice! I guess i should just try it out

fonsp:
Can you add some comments to the code? It's too difficult :)

Start by copy pasting your PR description into a docstring?

fonsp (changes requested):
Can you add some comments to the code? It's too difficult :)

Start by copy pasting your PR description into a docstring?

dralletje:
@fonsp 😏

fonsp:
All the crazy things I could come up with still work, awesome!

```julia
# ╔═╡ b474a054-846a-11ec-0f51-1b8e8c589aae
using HypertextLiteral

# ╔═╡ 3b09d217-42b9-473c-b98c-bcbbc1c856a4
@htl """
<span>
	<button>click meeee</button>
	<script>
		const span = currentScript.parentElement
		const btn = span.querySelector("button")

		btn.onclick = () => {
			console.log(currentScript, currentScript.parentElement, document.currentScript)
		}
	</script>
</span>


"""

# ╔═╡ 6868ed5d-0ca4-4a1a-be17-e8bde8f0f8ef
zz

# ╔═╡ adbcbc43-c88a-469b-8e8b-49a20706b415
@bind zz @htl """<script>
	let value = 1

	let handler = setInterval(() => {
		value += 1
		currentScript.value = value
		currentScript.dispatchEvent(
			new CustomEvent("input")
		)
	}, 500)

	invalidation.then(() => {
		clearInterval(handler)
	})
	
</script>
"""
```

cjdoris:
@dralletje Thanks so much for this! Can confirm that PythonCall objects (such as Altair plots) now display correctly in Pluto 0.18.0.

PR #1774, opened 2021-12-21
Make `publish_to_js` work in Static HTML again

fonsp:
It looks like `publish_to_js` did not work in the static HTML export since #1719 , this fixes it again :)

### DIFF
--- a/frontend/common/SliderServerClient.js
+++ b/frontend/common/SliderServerClient.js
@@ -4,7 +4,19 @@ import { pack, unpack } from "./MsgPack.js"
 import immer from "../imports/immer.js"
 import _ from "../imports/lodash.js"
 
-export const nothing_actions = ({ actions }) => Object.fromEntries(Object.keys(actions).map((k) => [k, () => {}]))
+const actions_to_keep = ["get_published_object"]
+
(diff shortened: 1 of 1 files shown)


fonsp:
@j-fu

j-fu:
Thanks for pinging!

With  "Pluto v0.17.3 `https://github.com/fonsp/Pluto.jl.git#bdc7b45` " this doesn't seem to have been fixed yet.
Neither in #41664ef.


I indeed was wondering  but  suspected everything else besides pluto :)

I made a small notebook for testing: https://gist.github.com/j-fu/3b7292c8766d0ed19f406fd638dcc4dc

fonsp:
This is probably because the static HTML export uses the CDN to get our JavaScript assets, and it will get it from the released version. So 
```
https://github.com/fonsp/Pluto.jl.git#bdc7b45
```

will generate HTML files that use
```
https://cdn.jsdelivr.net/gh/fonsp/Pluto.jl@0.17.3/frontend/components/Editor.js
```
Which still contain this bug

j-fu:
Ah, indeed, this works.

PR #1775, opened 2021-12-21
Make `DivElement` HTML-showable

fonsp:
So that you can use it inside other things! For example:

```julia
function ScrubbableMatrix2(A::Matrix{<:Real})
	# Create a new widgets that combines existing ones
	PlutoUI.combine() do Child
		
		# Create a matrix of scrubbables
		# (Note that each `Scrubbable` is wrapped in the `Child` function!)
		scrubbables = map(A) do x
			Child(Scrubbable(x))
		end

		# use Layout to display them in a grid
		Layout.grid(
			scrubbables;
			scrubbable_matrix_style...
		)
	end
end
```

`Layout.grid` returns a `DivElement`, and without this PR, it can not be rendered to HTML.

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -1965,6 +1965,14 @@ tree_data(@nospecialize(e::DivElement), context::IOContext) = Dict{Symbol, Any}(
 )
 pluto_showable(::MIME"application/vnd.pluto.divelement+object", ::DivElement) = true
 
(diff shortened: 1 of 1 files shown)


fonsp:
@dralletje What do you think?

dralletje:
Shouldn't we go all the way and make it `embed_display` automatically?

fonsp:
I merged #1805, this PR would be a 'performance optimization' or a 'feels better version' of that PR, I'll leave it open until I thought about it some more

PR #1778, opened 2021-12-22
invalidate expanded exprs when dependent macros gets removed

### LINKED ISSUE #1753: Renaming a macro does not cause dependent cells to error
When renaming a macro in one cell that is referenced in other cells the dependent cells do not error as happens when doing the same for a method definition.

Prior to renaming:

image

After renaming and evaluating the first cell:

image

Pangoraw:
Fixes #1753.


dralletje:
Nice

fonsp:
Looks good! Nice to make 'invalidated cells' explicit. I didn't look at the PR carefully but I'll note that we already have this for frontend JS, stored here:

https://github.com/fonsp/Pluto.jl/blob/f881a6f113d1c9c20fecead071e05cb31e938cc8/src/notebook/Cell.jl#L12-L13

maybe it makes sense to unify the two?

fonsp (approved):

fonsp:
@Pangoraw

fonsp:
@Pangoraw can you close this or write a TODO list?

Pangoraw:
I looked into removing `forced_expr_id` altogether but this moved a lot of logic to plutorunner so let's go with this instead!

is the `deletion_hook` in `update_run!` considered public API?

fonsp:
> is the deletion_hook in update_run! considered public API?

What do you mean? Outside of our own code, I would say no, I think `update_run!` also isn't...

Pangoraw:
> I think update_run! also isn't

Oh yeah! i have been confused because the tests are full of `update_run!` :see_no_evil:

fonsp:
Cool! So good to merge?

Pangoraw:
Let's go with this!

Issue #1684, opened 2021-11-23
Cell drag and drop broken in 0.17.2 (linux/chrome)


j-fu:
Hi, with the new release,  instead of the moved cell, Pluto inserts something like this:
```
png_joinpathsplit__FILE__1assets1_q8s7EQC6INcqCHZELbVIDA6png = let
    import PlutoUI
    PlutoUI.LocalResource(joinpath(split(@__FILE__, '#')[1] * ".assets", "1_q8s7EQC6INcqCHZELbVIDA 6.png"))
end
```
The cell to be moved remains on its place. Couriously, on Firefox this problem does not occur.

jcanizales:
Maybe Chrome isn't implementing the API as per the spec this relies on https://github.com/fonsp/Pluto.jl/pull/977/files ? The original "don't do it if dragging started inside the notebook" wording might be the best solution.

jcanizales:
FWIW, I'm seeing this on Pluto 0.16.0, Chrome 96.

fonsp:
Can you try out https://github.com/fonsp/Pluto.jl/pull/1780 to see if that fixes the problem?

j-fu:
Hi, ... well  it seem that this again is not working - my suspicion is chrome v97, because the fix in #1780 ceased to work, and so it doesn't work in 0.17.5 as well. I also tried Pluto 0.16.1 for double checking - the same.

fonsp:
Let's make a new issue

fonsp:
Moved to #1809

PR #1780, opened 2021-12-22
Remove file drag drop 😟

fonsp:
It has not been working well for a long time, and there is no clear path to make it work well. 

Small problems:
- Dragging cells between notebooks was broken
- #1684 
- The CSS looked like a quick hack

Larger problems:
1. Where do we store the file? Right now we create a directory for the notebook assets, but this makes reproducibility tricky. Renaming and relocating was difficult. If we were a cloud service like nextjournal.com, then we could store files online and solve all of these problems, but for a local notebook it is more difficult.
2. The generated code imports a package, but if you import the same file type twice, then the package is simply imported twice, triggering a MultipleDefinitionsError
3. Dropping a file *into* a cell did not make much sense, except when dragging an image into markdown, but that did not work.

Possible solutions are:
1. Embed a DataSets.toml in the notebook file (just like Project and Manifest) and use a base64 driver to store files directly in the notebook file. (Or become a cloud service.)
2. Insert the package import in another cell if it does not exist yet.
3. Smarter UI, maybe a centralized "files" panel with all uploaded files, and some inline CM widget to reference them.

We can always revisit this, and revert this PR to continue working on the feature.


j-fu:
Yeah, now I can move cells in chrome. Tanks!

PR #1783, opened 2021-12-23
Use uname -r instead of accessing a file to detect WSL

pankgeorg:
Tested on (Windows 11, WSL2 Ubuntu 20.04)

In rare cases, the file `"/proc/sys/kernel/osrelease"` seems to be locked and Pluto refuses to run as this read on line 19 is synchronous.

Note that this indicates a more system-wide issue that was solved with system restart.


pankgeorg:
Closing because the suggested way is the implemented one https://github.com/microsoft/WSL/issues/4071#issuecomment-496737093.

PR #1798, opened 2022-01-02
Fix #1018

### LINKED ISSUE #1018: Bold headers not working in Chrome
Chrome:

image

Firefox:

Schermafbeelding 2021-03-22 om 17 05 42

lucifer1004:
This PR fixes #1018 by adding a CSS rule.

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -177,6 +177,10 @@ pluto-output p:first-child {
     margin-block-start: 0px;
 }
 
+pluto-output strong {
+    font-weight: bolder;
+}
+
 /* I actually just want to get rid of the margin-block-start on the p tag,
     but css doesn't allow that. So I have to move the list up, instead of not moving it up */
 pluto-output li p + ul,
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -177,7 +177,9 @@ pluto-output p:first-child {
     margin-block-start: 0px;
 }
 
-pluto-output strong {
+/* https://github.com/necolas/normalize.css/blob/fc091cce1534909334c1911709a39c22d406977b/normalize.css#L96 */
+b,
+strong {
     font-weight: bolder;
 }
 


pankgeorg:
Hey! Thanks for taking the time reviewing this! Quick question, can you provide some context on the source of the issue?
Can you test a CSS library like https://necolas.github.io/normalize.css/? Does it also fix it?
(I'm not saying we should include it per se, just see if it a normalization issue!)

lucifer1004:
Source: [TEXT-BOLD Does Not Work Using HTML5](https://support.google.com/culturalinstitute/partners/answer/10460872)

>  It seems Chrome and Firefox (and other browsers) are rendering this in different ways. Firefox and Internet Explorer are making the bold text even bolder (font-weight: 900) when <strong> is used, while Chrome and Edge are not (font-weight: 700).

> It looks like if you explicitly set font-weight to 900 for the <strong> tags (actually "font-weight: bolder" might be best) then Chrome and Edge will work like Firefox and Internet Explorer.

lucifer1004:
For the `bolder` option: [Meaning of relative weights](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#meaning_of_relative_weights)

lucifer1004:
I had a look at the source code of `normalize.css` and it did exactly the [same](https://github.com/necolas/normalize.css/blob/fc091cce1534909334c1911709a39c22d406977b/normalize.css#L96).

frontend/editor.css
```diff
@@ -177,6 +177,10 @@ pluto-output p:first-child {
     margin-block-start: 0px;
 }
 
+pluto-output strong {
```
pankgeorg, line 180:
```suggestion
/* https://github.com/necolas/normalize.css/blob/fc091cce1534909334c1911709a39c22d406977b/normalize.css#L96 */
b,
strong {
```

pankgeorg:
> I had a look at the source code of `normalize.css` and it did exactly the [same](https://github.com/necolas/normalize.css/blob/fc091cce1534909334c1911709a39c22d406977b/normalize.css#L96).

That is great! Thank you for searching this through!

Issue #300, opened 2020-08-16
Allow using `Distributed`


fonsp:
Pluto uses `Distributed` to create worker processes and to send Julia data structures between them. It works really well! And `Distributed` is pleasant to work with.

However, it means that _Pluto notebooks cannot use `Distributed`_, because your notebook's code is _executed on a slave process_ - you can't create processes, and you can accidentally control other running notebooks.

One solution is to run all your notebooks in the master process by setting the parameter:
```julia
import Pluto; Pluto.run(workspace_use_distributed=false)
```

But this makes the notebook server unresponsive while any notebook is running code, and the stop button is disabled.

## Solutions

What would a solution be? Should Pluto implement its own `Distributed`? This seems silly - we would get the most robust implementation by copying `Distributed`'s source code directly. Maybe there is a way to internally use a copy of `Distributed`? Copy the contents of `julia/stdlib/Distributed` into `/tmp`, rename `Distributed` to `DistributedCopy` and import that? But is the session state completely contained inside the package?

Is there a way to use `Distributed`, but with "multiple global sessions"?

Can we create a thin wrapper around `Distributed` and make sure that the Pluto process is using this instead? For example, `Distributed.add_procs()` would be `RealDistributed.remoteeval(Main, 1, :(RealDistributed.add_procs()))`. Will that also work for the packages that you import inside your notebook, that depend on Distributed?

marius311:
Is there a reason why Pluto's workers need to be processes and can't just be threads? 

(Two other upsides of threads would be reduced memory usage and the ability to work with non-serializable objects.)

fonsp:
Hmm. Can one thread interrupt another thread stuck in `while true end`? What happens when one thread segfaults?

fonsp:
Can different threads use separate package environments? Can they load different versions of the same package?

fonsp:
(Those aren't rhetorical questions 🙃 it's just that I have never used threads in Julia)

Moelf:
jupyter notebook can do it because kernel is one process and jupyter notebook is running on a entirely separate (python) process.

I see two outs:
1. mimic jupyter and make each notebook a separate process and web UI tied to main session (so we can send SIGINT to each notebook still)

2. use main session as a worker broker, i.e when adding worker in a notebook, add at main process and 'shadow' / forward it to notebook worker

carlocab:
> One solution is to run all your notebooks in the master process by setting an environment variable:
> 
> ```julia
[...]

PR #1812, opened 2022-01-07
Restrict notebook communication topology

Oblynx:
Instigated by https://github.com/fonsp/Pluto.jl/issues/300

In case Pluto wants this, some tests would have to change, because they verify that notebooks can talk to each other.
Maybe workers are intended to be able to talk to each other.

### DIFF
--- a/src/evaluation/WorkspaceManager.jl
+++ b/src/evaluation/WorkspaceManager.jl
@@ -180,7 +180,7 @@ function create_workspaceprocess(;compiler_options=CompilerOptions())::Integer
     # run on proc 1 in case Pluto is being used inside a notebook process
     # Workaround for "only process 1 can add/remove workers"
     pid = Distributed.remotecall_eval(Main, 1, quote
-        $(Distributed_expr).addprocs(1; exeflags=$(_convert_to_flags(compiler_options))) |> first
+        $(Distributed_expr).addprocs(1; topology=:master_worker, exeflags=$(_convert_to_flags(compiler_options))) |> first
     end)
 
     for expr in process_preamble


pankgeorg:
Hey!

I have two concerns
1. this feature is listed as experimental
2. it's not clear what we gain from this; Pluto does not make "security" promises at the moment so it feels limiting without a clear upside (We don't protect you from 'adversary notebooks' - notebooks are yours!)

Also I'm not sure if we actually _need_ this to be open (@Pangoraw_ do you remember?)

Nevertheless, thank you for spending time on this! 💪🏾💪🏾

Oblynx:
I agree with your concerns. I'd love to be able to use `Distributed` and would argue that Pluto has to support every part of Julia stdlib if it's to be a Julia IDE. Regarding notebook isolation though, I have no other reason to show this apart from Fons's comment in #300

pankgeorg:
We are working on broad stdlib support! Latest major addition was the macro support 😄!
Until then we don't even claim to be Julia IDE 🙈🙈🙈 just simple reactive notebooks in Julia 🎈🎈🎈focused on education and teaching!

Oblynx:
> Latest major addition was the macro support

I can't express how important that is! Thank you so much for all your good work! I hope I can also help in a small way as a "power user" :)

dralletje:
SHUT UP

This is literally all that is necessary to allow Distributed inside Pluto???
Also why can't workers talk to each other now?

* Throws away Distributed clone 🙃 🥲

Oblynx:
> sadly doesn't make Distributed work

No this isn't about making Distributed work, for that check the last discussion in #300 

> as it still allows worker->master eval even

Really? Then I'm not sure if there's any point in this feature, maybe a bit of performance gain. Why don't you raise this up with Distributed upstream, to see what their use case is, since they mark this flag as "Experimental"

PR #1814, opened 2022-01-08
Add support for JULIA_NUM_THREADS=auto (fixes #1813)

### LINKED ISSUE #1813: Multi-Threading Bug
Hello dear friends.
I encounter a weird issue with **Pluto and Multi-Threading**.

I read the Julia Documentation to use more threads by default every time I launch Julia.
(https://docs.julialang.org/en/v1/manual/multi-threading/)

So, I set a system environment variable with the CMD (ad Admin):
**setx JULIA_NUM_THREADS "auto"**

When I run this command in Julia:
**Threads.nthreads()**
It shows **"8"**, that's correct because it's the number of my local CPU threads.

The problem is when I try to run Pluto (either **"using Pluto"** or **"import Pluto"**), and then **"Pluto.run()"**.
I get this error:

imagen

Thank you for making this awesome library, Pluto is the best !

lucifer1004:
For Julia version 1.7+, add support for `JULIA_NUM_THREADS=auto`.

### DIFF
--- a/src/Configuration.jl
+++ b/src/Configuration.jl
@@ -143,7 +143,14 @@ end
 
 function default_number_of_threads()
     env_value = get(ENV, "JULIA_NUM_THREADS", "")
-    all(isspace, env_value) ? roughly_the_number_of_physical_cpu_cores() : parse(Int, env_value)
+
+    all(isspace, env_value) ?
+        roughly_the_number_of_physical_cpu_cores() :
+        try
+            parse(Int, env_value)
+        catch
+            env_value
+        end
 end
 
 function roughly_the_number_of_physical_cpu_cores()


lucifer1004:
This will fix #1813

fonsp:
Thanks! Just to confirm, it should be the number of logical cores, not the number of physical cores, right? (e.g. an intel i5 has 4 physical cores, 8 "logical cores"). Right now the PR uses the logical core number

lucifer1004:
> Thanks! Just to confirm, it should be the number of logical cores, not the number of physical cores, right? (e.g. an intel i5 has 4 physical cores, 8 "logical cores"). Right now the PR uses the logical core number

Yes, this is according to the behavior of Julia itself.

fonsp:
Maybe we just pass a string value through? So
```julia
function default_number_of_threads()
    env_value = get(ENV, "JULIA_NUM_THREADS", "")
    all(isspace, env_value) ? 
		roughly_the_number_of_physical_cpu_cores() : 
		try
			parse(Int, env_value)
		catch
			# env_value is not a number, but a string like "auto"
			env_value
		end
end
```

lucifer1004:
According to https://docs.julialang.org/en/v1/manual/multi-threading/

> The number of threads can either be specified as an integer (--threads=4) or as auto (--threads=auto), where auto sets the number of threads to the number of local CPU threads.

> Using `auto` together with the environment variable `JULIA_NUM_THREADS` requires at least Julia 1.7.

lucifer1004:
Should we at least keep the version check?

fonsp:
I realise that I don't understand this message in the docs:

> Using `auto` together with the environment variable `JULIA_NUM_THREADS` requires at least Julia 1.7.

Do you know what it means? Does it mean: "Using `auto` **as the value of** the environment variable `JULIA_NUM_THREADS` requires at least Julia 1.7."?

lucifer1004:
I have not tested but I believe the documentation means using `JULIA_NUM_THREADS=auto` causes error for versions `<1.7`.

fonsp:
Schermafbeelding 2022-01-17 om 14 08 29

fonsp:
Thanks @lucifer1004 ! Your PR is now released in 0.17.6 🌟

What do you want to work on next?

PR #1823, opened 2022-01-15
Use newer keyword arg syntax

rikhuijzer:
Since Pluto's lower bound for Julia is now 1.5, the newer keyword argument syntax can be used (https://github.com/JuliaLang/julia/blob/v1.5.0/NEWS.md). I've ran `comby` with `'(:[a]; :[b], :[c]=:[c])' '(:[a]; :[b], :[c])'` and some variations of this pattern to replace keyword arguments such as `x=x` by `x`.


fonsp:
Awesome! I like it

PR #1834, opened 2022-01-21
Cache files from frontend-dist if they have a parcel hash

fonsp:
I noticed that the new version of Pluto loads quite slow, because it is getting all assets (including big fonts) from the binder server (because we have offline support), without caching. This will cache all files that parcel wants us to cache

### DIFF
--- a/src/webserver/Static.jl
+++ b/src/webserver/Static.jl
@@ -8,7 +8,7 @@ function is_pluto_dev()
     if found_is_pluto_dev[] !== nothing
         return found_is_pluto_dev[]
     end
-    found_is_pluto_dev[] = try
+    found_is_pluto_dev[] = get(ENV, "JULIA_PLUTO_FORCE_BUNDLED", "nein") != "ja" && try
         deps = Pkg.dependencies()
 
         p_index = findfirst(p -> p.name == "Pluto", deps)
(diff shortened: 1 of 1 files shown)


src/webserver/Static.jl
```diff
@@ -8,7 +8,7 @@ function is_pluto_dev()
     if found_is_pluto_dev[] !== nothing
         return found_is_pluto_dev[]
     end
-    found_is_pluto_dev[] = try
+    found_is_pluto_dev[] = get(ENV, "JULIA_PLUTO_FORCE_BUNDLED", "nein") != "ja" && try
```
dralletje, line 11: I ~~hate~~ love you

dralletje:
Wouldn't make first launch faster as they'll get a new binder domain/path every time? :(
But enabling caching is good either way!!

fonsp:
Good point! Let's do a 
`<link rel="preload" href="editor.css">`
in the index page? That will start loading the fonts, right?

fonsp:
Should we do something with Etag?

fonsp:
We tried some stuff, Etag seems non-trivial to implement with not so much benefit right now.

pankgeorg:
> We tried some stuff, Etag seems non-trivial to implement with not so much benefit right now.

Since our content has unique names for each file and revision, ETags are not strictly necessary (there will never be another version of the _same_ filename. ETags are useful for REST API where resources get updated and for files that maintain their URI. If we serverd `editor.js`

src/webserver/Static.jl
```diff
@@ -334,4 +334,4 @@ function http_router_for(session::ServerSession)
     HTTP.@register(router, "GET", "/favicon.ico", create_serve_onefile(project_relative_path(frontend_directory(allow_bundled=false), "img", "favicon.ico")))
 
     return router
-end
\ No newline at end of file
+end
```
pankgeorg, line 337: Thanks for this!

fonsp: wasn't me! i dont care of course :)

pankgeorg: It's sooooo annoying when GitHub places the ⛔️

PR #1835, opened 2022-01-21
wider line number gutter (fix #1826)

### LINKED ISSUE #1826: Line number gutter to narrow for long cells
Peek 2022-01-18 11-48

Two options:
- fixed gutter width so the editor content doesn't jump around
- measure text and set gutter width dynamically depending on cell length

lucifer1004:
Increase `min-width` of `.cm-gutter` from `29px` to `31px` to support up to 999 lines without change of gutter width.

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -2618,7 +2618,7 @@ pluto-input .cm-editor .cm-content {
 }
 
 .cm-editor .cm-gutter {
-    min-width: 29px;
+    min-width: 31px;
     min-height: 23px;
 }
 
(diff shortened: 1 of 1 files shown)


fonsp:
It works, awesome!

https://user-images.githubusercontent.com/6933510/150539345-cfac71dc-c47b-4c31-bdcc-3d4a9ffd8c4a.mov

fonsp:
I also reworked the way we switch between number and dot on hover to avoid jumpy layout for >999 lines:

https://user-images.githubusercontent.com/6933510/150539413-02d696cf-ac38-4de7-a82f-44f63e25bfac.mov

Devtools paint flashing tells me that this is a good strategy :)

dralletje:
Awesomeeeee

PR #1853, opened 2022-01-25
Fix #1852, relocatable changes for sysimage usage.

### LINKED ISSUE #1852: Use RelocatableFolders.jl instead of project_relative_dir to allow sysimages
Right now, Pluto is not relocatable when adding it to a sysimage created with PackageCompiler. This is because we use files from Pluto's source code dynamically, notably the `frontend/`, `frontend-dist/` (if it exists) and `sample/` folders.

The package https://github.com/JuliaPackaging/RelocatableFolders.jl/ solves this problem 🎉 and it looks like it will be relatively easy to use in Pluto.

---

To work on this issue, you can get started by finding all the places where we use files from Pluto's source code (you can ignore tests), and use RelocatableFolders instead. 

We then want to verify that the solution works by creating the sysimage using PackageCompiler. Compare results before and after the PR.

MichaelHatherly:
Uses `RelocatableFolders` to store all files that may be accessed
at runtime rather than baking in their file paths, which won't
point to valid file paths once a system image is created containing
`Pluto` and then moved to a different machine/system.


src/Pluto.jl
```diff
@@ -9,12 +9,28 @@ Have a look at the FAQ:
...
+const frontend_dist_exists = FRONTEND_DIR !== FRONTEND_DIST_DIR
+const SAMPLE_DIR = @path(joinpath(ROOT_DIR, "sample"))
+const RUNNER_DIR = @path(joinpath(ROOT_DIR, "src", "runner"))
+function project_relative_path(root, xs...)
+    root == joinpath("src", "runner") ? joinpath(RUNNER_DIR, xs...) :
+    root == "frontend-dist" && frontend_dist_exists ? joinpath(FRONTEND_DIST_DIR, xs...) :
+    root == "frontend" ? joinpath(FRONTEND_DIR, xs...) :
+    root == "sample" ? joinpath(SAMPLE_DIR, xs...) :
+        normpath(joinpath(dirname(dirname(pathof(Pluto))), root, xs...))
+end
```
MichaelHatherly, line 28: These changes are slightly more complex than they could be if all assets that need this treatment were in a single root folder, but that change would have been much more invasive that just doing this.

src/Pluto.jl
```diff
@@ -9,12 +9,28 @@ Have a look at the FAQ:
 https://github.com/fonsp/Pluto.jl/wiki
 """
 module Pluto
-project_relative_path(xs...) = normpath(joinpath(dirname(dirname(pathof(Pluto))), xs...))
+
+import RelocatableFolders: @path
+const ROOT_DIR = normpath(joinpath(@__DIR__, ".."))
+const FRONTEND_DIR = @path(joinpath(ROOT_DIR, "frontend"))
+const FRONTEND_DIST_DIR = let dir = joinpath(ROOT_DIR, "frontend-dist")
+    isdir(dir) ? @path(dir) : FRONTEND_DIR
```
MichaelHatherly, line 17: I'm not 100% sure on the `frontend` and `frontend-dist` stuff, hopefully this logic is actually doing the correct thing.

src/evaluation/WorkspaceManager.jl
```diff
@@ -22,9 +22,9 @@ Base.@kwdef mutable struct Workspace
 end
 
 "These expressions get evaluated whenever a new `Workspace` process is created."
-const process_preamble = quote
+process_preamble() = quote
```
MichaelHatherly, line 25: This needed to change to a function such that the path could be queried at runtime rather than at `using`-time which would pick the wrong path to `include`.

MichaelHatherly:
Testing of relocatablility for these changes is a case of doing the following steps:

- make a system image using `create_sysimage`
- delete/move the folders that need to be available
- start julia with the system image
- check that pluto still runs.

dralletje:
👏🏽 👏🏽 👏🏽

fonsp:
😅 I was going to say: This adds about 1 second of loading time on my computer because it reads and hashes all the files. But yes probably better to do a yolo merge

fonsp:
https://github.com/fonsp/Pluto.jl/issues/1884 would help

dralletje:
>_<

dralletje:
@MichaelHatherly way to only load the files into memory when running with PackageCompiler?

MichaelHatherly:
> This adds about 1 second of loading time on my computer because it reads and hashes all the files.

Just on precompile, or every `using Pluto`? Either is pretty bad, we should try mitigate that.

> way to only load the files into memory when running with PackageCompiler?

May be a way to, would probably be best built into `RelocatableFolders` itself to be able to skip when PackageCompiler isn't what is causing the precomp, I'll take a look at it.

src/evaluation/WorkspaceManager.jl
rikhuijzer: Actually, I just was planning to send in a PR to change this to a function too because it now allows users to override the preamble :rofl: What a coincidence. Thanks!

PR #1854, opened 2022-01-25
Without distributed

dralletje:
It works!

But it is a bit slower than Distributed..
Not so much you notice it while running normally, but you do notice it when dragging a `bind`-ed slider :/


Pangoraw:
Awesome Work! Do you have any idea why it might be slower than Distributed ? Maybe it is the pipes buffering being slower than the socket Distributed is using ?

PS: For anyone wanting to see the code in the notebooks:
- https://pluto-static-viewer.netlify.app/viewer.html?url=https://raw.githubusercontent.com/fonsp/Pluto.jl/without-distributed/src/evaluation/ChildProcessesNotebook.jl
- https://pluto-static-viewer.netlify.app/viewer.html?url=https://raw.githubusercontent.com/fonsp/Pluto.jl/without-distributed/sample/ChildProcesses_benchmark.jl

fonsp:
We don't know *why* it is slower than distributed, but we do know that Distributed uses a TCP server (`Sockets` stdlib) to talk server-worker, while we use stdin/stdout. I think that we should also use TCP, do the benchmark again and reevaluate.

I tried to get a server-worker communication working with the `Sockets` stdlib, but it was surprisingly difficult... Avik mentioned https://github.com/JuliaInterop/ZMQ.jl today (in another context), which is apparently working very well, and I believe that IJulia.jl also uses it. So maybe we should use ZMQ.jl instead of writing our own TCP server/client!

(If you want to give this a try, write a comment so that we don't end up working on it at the same time)

Issue #1800, opened 2022-01-03
Pluto hangs when cell contains `(md" ", md"$t")`


karlwessel:
Pluto notebook does not react anymore if the following MWE is passed to a new empty cell:
```
(md" ", md"$t")
```
Only happens with Pluto 0.17.4. On Firefox and Chrome. I cleared the cache for localhost and used a new temporary julia environment for testing this.

lucifer1004:
This also happens when trying to insert a `$` into `md"a", md"|a"` (`|` denotes the cursor).

fonsp:
@pankgeorg can you take a look?

fonsp:
This is caused by #1533 , and we are working on fixing it, but it seems more difficult than we thought.

We will revert #1533 in the next release (0.17.5), to workaround this issue, and continue working on it...

PR #1855, opened 2022-01-26
Re-enable parseMixed

pankgeorg:
The latest [codemirror-pluto-setup](https://github.com/JuliaPluto/codemirror-pluto-setup) works.
Needs some extensive testing before applying it.

Fixes: #1800 
Related #1533


pankgeorg:
Non-interpolating strings are now ignored by default - let's readjust based on that.!

dralletje:
Playing with it a bit, still need to do dark mode (and CSS (the language)) but I think it looks pretty pretty already.
I've tried to give html and markdown their own color (CSS also but it is a bit much), and javascript and python are "toned down" versions of Julia. 

2BF86449-88D1-4D01-9DBF-4F98EE1EC074
48E575CE-2825-4887-A929-FF3F82413FF2
3EB22212-BCFD-454A-B5C6-81D2BD71994D

dralletje:
Still getting a bunch of errors when working with interpolations :(

pankgeorg:
Thank you for taking a look <3 ! Regarding the interpolation issue, do you think it's an overlay issue, a parser issue or something with CodeMirror? Also, love the `scope` config! 👏🏾👏🏾

dralletje:
> Thank you for taking a look <3 ! Regarding the interpolation issue, do you think it's an overlay issue, a parser issue or something with CodeMirror? Also, love the `scope` config! 👏🏾👏🏾

It's a problem with the markdown parser specifically.
Markdown has a special (non-lezer) parser so I guess that is why it has problems.

frontend/components/CellInput/mixedParsers.js
```diff
@@ -17,14 +17,12 @@ import {
...
 const mdParserExt = markdownLanguage.parser.configure(parseCode({ htmlParser }))
 const postgresParser = PostgreSQL.language.parser
 const sqlLang = sql({ dialect: PostgreSQL })
 const pythonParser = pythonLanguage.parser
 
-const MD_SIMPLE_TAGS = ["md", "mermaid"].flatMap((x) => [x, `@${x}`])
-const MD_EXTENDED_TAGS = ["cm", "markdown", "mdx", "mdl", "markdownliteral"].flatMap((x) => [x, `@${x}`])
+const MD_TAGS = ["md", "mermaid", "cm", "markdown", "mdx", "mdl", "markdownliteral"].flatMap((x) => [x, `@${x}`])
 
 const overlayHack = (overlay, input) => {
```
fonsp, line 27: Can you add a docstring? We probably shouldn't have hacks without comment

frontend/components/CellInput.js
```diff
@@ -56,74 +60,227 @@ import { bracketMatching } from "./CellInput/block_matcher_plugin.js"
...
-    { tag: tags.link, color: "var(--cm-link-color)" },
+        scope: pythonLanguage.topNode,
+        all: {
+            color: "var(--cm-editor-text-color)",
+            filter: `contrast(0.5)`,
+        },
+    }
+)
+
+export const pluto_syntax_colors_css = HighlightStyle.define(
```
fonsp, line 211: We have a lot of `#rrggbb` colors for the CSS syntax, can those all be CSS variables? Maybe just reuse random vars from Julia, even if they are not a sematic match

frontend/imports/CodemirrorPlutoSetup.js
```diff
@@ -56,7 +56,7 @@ import {
     javascript,
     python,
     Text,
-} from "https://cdn.jsdelivr.net/gh/JuliaPluto/codemirror-pluto-setup@0.25.0/dist/index.es.min.js"
+} from "https://cdn.jsdelivr.net/gh/JuliaPluto/codemirror-pluto-setup@0.26.0/dist/index.es.min.js"
```
fonsp, line 59: Can we do this on `main` right now?

pankgeorg: sure!

frontend/components/CellInput.js
dralletje: I'll give CSS another try

dralletje:
I can't get lezer-markdown to even install on my machine, so I say we merge this without markdown interpolations (markdown interpolations that only work with `@markdown` anyway, because `md".."` doesn't have interpolations anyway)

dralletje:
@pankgeorg any idea why Github says 14 files changed? With like a bunch of changes that should already be in master?

pankgeorg:
> @pankgeorg any idea why Github says 14 files changed? With like a bunch of changes that should already be in master?

fixed!

dralletje:
@fonsp Speak now (or in the next hour) or forever hold your peace

PR #1857, opened 2022-01-26
Dark background image but not that loud!

pankgeorg:
Also with the `main` area significantly darker than before (so you can read it as an article!)
image


dralletje:
Hahaa, looks pretty.
Where did you get this image? :o

frontend/dark_color.css
```diff
@@ -189,8 +189,22 @@
     }
     body:not(.disable_ui):not(.more-specificity) {
         /* background by SVGBackgrounds.com */
```
pankgeorg, line 191: From this website @dralletje, it's cool!

PR #1864, opened 2022-01-28
DOI in REAMDE for citation

rikhuijzer:
In response to a discussion on Slack; a very basic citation file. The benefit of these files is that its clear for people what to cite and tooling such as Zotero should be able to interpret it.

I'm not sure how to handle the many contributors, so that's why I've added only the first three.


fonsp:
Let's start by putting the DOI in the readme? Our version-independent DOI is:

https://doi.org/10.5281/zenodo.4792401

[DOI](https://doi.org/10.5281/zenodo.4792401)

This links to our zenodo page, which has download links for bibtex and such.

I'm not entirely happy with giving an "official" shortlist of authors to cite right now. If we do that, I think it should be after discussing with the many other authors, but it doesn't seem super important right now (i think). So let's avoid the problem by using the auto-generating list from zenodo :)

rikhuijzer:
Makes sense. Without authors, the functionality breaks as can be seen in <https://github.com/rikhuijzer/citationexperiment>. There, the cite this repository button is broken when authors are removed.

I've added the badge below Pluto.jl

## Preview

image

PR #1871, opened 2022-01-30
Server-side events as configuration

ctrekker:
As discussed in the [developer call on 1/27](https://juliapluto.github.io/weekly-call-notes/2022/01-27/notes.html), here are a few changes related to server-side event listeners.

### List of changes
- Move event listener to configuration
    - Had to add a `MsgPack.to_msgpack` for functions that returns null so Pluto doesn't try to serialize the event listener function
- Update `Events.jl` docstring with updated instructions
- Rename `event_listener` to `on_event` since its shorter and maybe more descriptive?
- Assign configuration port if none was provided (see [comment below](https://github.com/fonsp/Pluto.jl/pull/1871#discussion_r795117452))

### Example Usage
```julia
import Pluto

# catches unwanted events
function myfn(a::PlutoEvent)
    nothing
end
# catches the notebook `save` event
function myfn(a::FileSaveEvent)
    HTTP.post("https://my.service.com/count_saves")
end

Pluto.run(; on_event=myfn)
```

### Backwards Compatibility
This PR _would_ technically break backwards compatibility. I don't think refactoring everything to kwargs is too bad, but if others think otherwise we can add the following method definition for `ServerSession` and everything should (I didn't test it) be backwards compatible.

```julia
function ServerSession(; event_listener::Function, kwargs...)
    s = ServerSession(; kwargs...)
    s.options.server.on_event = event_listener
    s
end
```

### TODO
- [ ] On server start event that sends port

### DIFF
--- a/src/webserver/WebServer.jl
+++ b/src/webserver/WebServer.jl
@@ -130,6 +130,7 @@ function run(session::ServerSession, pluto_router)
     hostIP = parse(Sockets.IPAddr, host)
     if port === nothing
         port, serversocket = Sockets.listenany(hostIP, UInt16(1234))
+        session.options.server.port = port
     else
         try
             serversocket = Sockets.listen(hostIP, UInt16(port))
--- a/src/webserver/WebServer.jl
+++ b/src/webserver/WebServer.jl
@@ -130,7 +130,6 @@ function run(session::ServerSession, pluto_router)
     hostIP = parse(Sockets.IPAddr, host)
     if port === nothing
         port, serversocket = Sockets.listenany(hostIP, UInt16(1234))
-        session.options.server.port = port
     else
         try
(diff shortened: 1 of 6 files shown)


src/webserver/WebServer.jl
```diff
@@ -130,6 +130,7 @@ function run(session::ServerSession, pluto_router)
     hostIP = parse(Sockets.IPAddr, host)
     if port === nothing
         port, serversocket = Sockets.listenany(hostIP, UInt16(1234))
+        session.options.server.port = port
```
ctrekker, line 133: I added this because otherwise server-side event consumers would have no way of knowing what port Pluto is running on. This would only assign the configured port when it is not explicitly provided.

PR #1874, opened 2022-01-30
fix is_stdlib() for nightly and enable nightly tests

### LINKED ISSUE #1872: Stdlib packages marked as not found
Stdlib packages are indicated as not found in the registry, but they execute correctly.

There's also a LocalRegistry in use, if that matters.

image

Pluto v0.17.7, Julia 1.8-DEV

Pangoraw:
fixes #1872

### DIFF
--- a/src/packages/PkgCompat.jl
+++ b/src/packages/PkgCompat.jl
@@ -216,13 +216,14 @@ end
 # (⚠️ Internal API with fallback)
 _stdlibs() = try
 	stdlibs = values(Pkg.Types.stdlibs())
-	stdlibs = if VERSION > v"1.8.0-"
+	T = eltype(stdlibs)
+	if T == String
+		stdlibs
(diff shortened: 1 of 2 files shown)


src/packages/PkgCompat.jl
```diff
@@ -215,7 +215,14 @@ end
 
 # (⚠️ Internal API with fallback)
 _stdlibs() = try
-	values(Pkg.Types.stdlibs())
+	stdlibs = values(Pkg.Types.stdlibs())
+	stdlibs = if VERSION > v"1.8.0-"
```
fonsp, line 219: Can you replace this with a feature check? e.g.

```julia
if eltype(stdlibs) == String
	stdlibs
elseif eltype(stdlibs)
	first.(stdlibs)
else
	error("Failed to get stdlibs")
end
```

PR #1877, opened 2022-01-30
respect maxlog keyword for logs, fixes #1858

### LINKED ISSUE #1858: Logging doesn't respect `maxlog`
https://user-images.githubusercontent.com/13287213/151267144-54483481-c3fa-4aa5-ae6d-f7100b605e8d.mp4

The new logging feature for Pluto doesn't support the `maxlog` argument for logging. Here is a link to the [notebook](https://gist.github.com/Joel-Dahne/851ad2768a4e3495b7e6edec9846c166).

I can also maybe note that I had performance issues with Pluto when there was a very large number of logs. You can try to increase the counter in the loop to, say, 10000, and at least for me the webpage grinds to a halt. In my case it was only a problem because `maxlog` was not respected though.

Pangoraw:


Pangoraw:
ci failure seems related, i'll investigate

PR #1878, opened 2022-01-30
allow .. as an identifier, fixes #1875

### LINKED ISSUE #1875: EllipsisNotation.jl (`randn((10,10))[1, ..]`) does not work
Hi!

[EllipsisNotation.jl](https://github.com/ChrisRackauckas/EllipsisNotation.jl) does not work properly.

The [notebook](https://gist.github.com/roflmaostc/e2c579a47091c38aff8cd73502769f20).

Attached an image. Julia 1.7.1 and Pluto v0.17.7.

Best,

Felix

Pangoraw:
ultimately, more work has to be done to support any variable name with a dot (like `var"my.var"`) but it should be enough for now

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -271,7 +271,7 @@ end
 
 # this is stupid -- désolé
 function is_joined_funcname(joined::Symbol)
-    occursin('.', String(joined))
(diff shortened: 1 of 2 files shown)


roflmaostc:
Thanks for fixing #1875 :+1:

dralletje:
`..()` derp

PR #1879, opened 2022-01-31
don't error when state contains Ptrs

aplavin:
eg warnings from depwarn()

### DIFF
--- a/src/webserver/MsgPack.jl
+++ b/src/webserver/MsgPack.jl
@@ -12,10 +12,12 @@ MsgPack.msgpack_type(::Type{<:MIME}) = MsgPack.StringType()
 MsgPack.msgpack_type(::Type{UUID}) = MsgPack.StringType()
 MsgPack.msgpack_type(::Type{VersionNumber}) = MsgPack.StringType()
 MsgPack.msgpack_type(::Type{Pkg.Types.VersionRange}) = MsgPack.StringType()
+MsgPack.msgpack_type(::Type{<:Ptr}) = MsgPack.IntegerType()
 MsgPack.to_msgpack(::MsgPack.StringType, m::MIME) = string(m)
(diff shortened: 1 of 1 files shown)


dralletje:
Do you have a quick example for how I can trigger the error you are fixing?

aplavin:
Run any function that throws a depwarn, eg:
```
# ╔═╡ 0328b0ab-faf6-4c80-9d0e-bc06ed90f9bc
using LinearAlgebra

# ╔═╡ f5ecc5ca-7a2b-4c96-ad74-97e01e7803cf
BLAS.vendor()
```
Go back to the pluto notebooks list (index page) and try to open this notebook again (don't restart it). Pluto webinterface hangs with "Loading...", and the terminal shows
```
┌ Warning: Failed to write to WebSocket of xa3u8u 
│   exception =
│    no non-`AnyType` MsgPack mapping found for Ptr{Nothing}; please overload `msgpack_type` for this type.
...
```

dralletje:
I see.
Ideally we don't transfer the `log.id` to the main process/the client at all.. (or we stringify it)
@Pangoraw we're only using it for ProgressLogging right now, right? You know if it safe to "just" `string(...)` it?

Pangoraw:
oh interesting, I totally did not expect log ids to be custom objects. We can use `objectid(log.id) |> string` which gives a consistent hash for ProgressLogging if not serialisable ?

fonsp:
Oh no! I closed by accident and now it says:

Schermafbeelding 2022-02-02 om 13 01 57

fonsp:
Thanks for reporting this! Fixed in https://github.com/fonsp/Pluto.jl/commit/f094e02fe76b670a504a90320c4081c13632fc81

aplavin:
Thanks for a more proper fix, seems to work fine!

PR #1881, opened 2022-02-02
Import PlutoRunner as module in Loader.jl

dralletje:
No idea if this is faster currently, or if it will every be...
But apart from having to do `Pkg.develop()` this feels like a more mature way to load PlutoRunner :P


fonsp:
Nice! I ended up rewriting the whole thing:
- Now using `LOAD_PATH` and `Base.ACTIVE_PROJECT` instead of `pkg_ctx`
- NotebookBootEnvironment.toml no longer needed, since PlutoRunner is the only "dependency"
- Bit cleaner with `try ... finally`

fonsp:
We still need to test:
- [ ] Does this improve launch time?
- [ ] How do different Pluto versions interfere?
- [ ] What about RelocatableFolders.jl? (Was that working before?)

dralletje:
> Nice! I ended up rewriting the whole thing:
> 
> - Now using `LOAD_PATH` and `Base.ACTIVE_PROJECT` instead of `pkg_ctx`
> 
> - NotebookBootEnvironment.toml no longer needed, since PlutoRunner is the only "dependency"
> 
> - Bit cleaner with `try ... finally`


Eyeroll... 😉

dralletje:
> - [ ] What about RelocatableFolders.jl? (Was that working before?)

I'll test it, but I'm pretty sure that still works as RelocatableFolders will "just" create the whole `runner` folder.

src/runner/Loader.jl
```diff
@@ -1,46 +1,62 @@
+# The goal of this file is to import PlutoRunner into Main. 
+# 
+# This is difficult because PlutoRunner uses standard libraries and packages that are not necessarily available in the standard environment. 
+# 
+# Our solution is to create a temporary environment just for loading PlutoRunner. This environment is stored in `.julia/environments/__pluto_book_v*_*/`, and used by all notebook launches. Reusing the environment means extra speed.
```
rikhuijzer, line 5: Great comment. I was already confused about the code below this

rikhuijzer (approved):

rikhuijzer:
> Does this improve launch time? (@rikhuijzer can you use the new benchmark system to test this?)

Hard to tell since this branch isn't up-to-date with `main`. I've manually added the compile tests on top of this branch and it doesn't look off. Some stuff of the `PlutoRunner` is covered via `SessionActions.open`. I should definitely some PlutoRunner benchmarks.

Don't let performance nitpicks come in the way of features :) We can always optimize later.

fonsp:
Let me rebase! Sorry about that

fonsp:
Updated! The main benefit should be in the launch time of the second notebook, because PlutoRunner can be precompiled like a package.

fonsp:
This PR can cause problems when you use Pluto inside Pluto, at a different version than the server. But that problem might already be there because we check the existence of `Main.PlutoRunner`... needs more thinking

Pangoraw:
With 1.9 it might be interesting to revisit this PR. Here are some unscientific measurements:

On main (julia 1.9-rc2):

```julia
julia> @time let
           nb = Pluto.Notebook([Pluto.Cell("1 + 1")])
           sess = Pluto.ServerSession()
           Pluto.update_run!(sess, nb, nb.cells)
           nb.cells[1].output.body
       end
  7.291519 seconds (164.88 k allocations: 12.471 MiB, 0.29% compilation time)
"2"
```

With the loader as a package, this pr (julia 1.9-rc2):

```julia
julia> @time let
           nb = Pluto.Notebook([Pluto.Cell("1 + 1")])
           sess = Pluto.ServerSession()
           Pluto.update_run!(sess, nb, nb.cells)
           nb.cells[1].output.body
       end
  5.720203 seconds (164.12 k allocations: 12.435 MiB, 0.36% compilation time)
"2"
```

Note that there are no precompile statements in PlutoRunner itself so it can probably be improved more?

> This PR can cause problems when you use Pluto inside Pluto, at a different version than the server. But that problem might already be there because we check the existence of Main.PlutoRunner... needs more thinking

This problem will be solved when we enable Malt so that each Pluto instance can drive its own notebook processes independently.

disberd:
I'd like to add that apart from precompilation time this PR (making PlutoRunner its own package with UUID) will opens up a lot of possibilities regarding building package extensions to do Pluto specific stuff.

At the moment one has to rely on AbstractPlutoDingetjes but having a Pluto-functionality extensions for packages that is loaded when PlutoRunner is loaded would be super nice!

PR #1882, opened 2022-02-02
Server start event

ctrekker:
Adds an event that responds with the Pluto server's web address after the server has been started.

### DIFF
--- a/src/notebook/Events.jl
+++ b/src/notebook/Events.jl
@@ -49,7 +49,7 @@ end
 # Triggered when the web server gets started
 struct ServerStartEvent <: PlutoEvent
     address::String
-    port::Int
+    port::UInt16
 end
 
 # Triggered when a notebook is saved
--- a/src/notebook/Events.jl
+++ b/src/notebook/Events.jl
@@ -45,6 +45,12 @@ function try_event_call(session, event::PlutoEvent)
     end
 end
 
+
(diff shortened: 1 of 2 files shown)


dralletje:
Yesss, this is very good

src/notebook/Events.jl
```diff
@@ -45,6 +45,12 @@ function try_event_call(session, event::PlutoEvent)
     end
 end
 
+
+# Triggered when the web server gets started
+struct ServerStartEvent <: PlutoEvent
+    address::String
```
ctrekker, line 51: I only include the unparsed address because
1. Everyone's needs are different and this includes the necessary information for connecting to the server
2. There are several different ways to specify the same network configuration

fonsp:
Maybe also add the port as integer?

src/notebook/Events.jl
fonsp: I vote for just giving the port as integer as a second field

pankgeorg: agree - the address that Pluto knows is not useful outside the machine either way (binding to `0.0.0.0:1234` means `<public_ip>:1234` for anyone outside)

ctrekker: Good point, I have added in the second field for port

PR #1883, opened 2022-02-02
Extract SessionActions.run method

rikhuijzer:


fonsp:
I think it's good to expose `update_save_run!` as public API! This PR is a bit odd, because it is not just `run`, it is more `add to session, set cell statuses to queued, update_save_run if the configuration wants to run, and call the file open event`. 

Maybe we should just expose `update_save_run!` without anything extra?

rikhuijzer:
Ah okay. So, if I understand correctly, in most cases `update_save_run!` will have the same effect and the other things in the method are more about managing things when someone is running Pluto in the browser

Okay, then this PR doesn't make sense no.

> Maybe we should just expose `update_save_run!` without anything extra?

At the PlutoStaticHTML side I'm already using many internals anyway, so it doesn't really matter whether it's explicitly exported or not. I'll close this PR and exposing can be picked up later or so

PR #1895, opened 2022-02-04
Cell metadata and persistent cell deactivation

lungben:
ToDo:

- [x] the indirectly disabled metadata is not updated directly when cells are activated/deactivated, but only after the next notebook update. This requires re-arranging dependency calculations, see #1209 
- [x] Comment out(indirectly and directly) deactivated cells
- [x] Load and parse metadata
- [x] Pass metadata to the frontend (not required for cell deactivation, but for other future use cases of metadata)


src/evaluation/Run.jl
```diff
@@ -458,7 +457,7 @@ function update_save_run!(session::ServerSession, notebook::Notebook, cells::Arr
 		sync_nbpkg(session, notebook; save=(save && !session.options.server.disable_writing_notebook_files))
 		if !(isempty(to_run_online) && session.options.evaluation.lazy_workspace_creation) && will_run_code(notebook)
 			# not async because that would be double async
-			run_reactive_async!(session, notebook, old, new, to_run_online; run_async=false, kwargs...)
+			run_reactive_async!(session, notebook, old, new, to_run_online; run_async=false, indirectly_deactivated=indirectly_deactivated, kwargs...)
```
rikhuijzer, line 460:
```suggestion
			run_reactive_async!(session, notebook, old, new, to_run_online; run_async=false, indirectly_deactivated, kwargs...)
```
Minimum Julia version ≥ 1.4 so we can use shorter kwarg syntax

src/notebook/Notebook.jl
```diff
@@ -204,13 +223,37 @@ function load_notebook_nobackup(io, path)::Notebook
...
             code_normalised = replace(code_raw, "\r\n" => "\n")
+
+            # remove the disabled on startup comments for further processing in Julia
+            code_normalised = replace(replace(code_normalised, _disabled_prefix => ""), _disabled_suffix => "")
+
             # remove the cell suffix
             code = code_normalised[1:prevind(code_normalised, end, length(_cell_suffix))]
 
-            read_cell = Cell(cell_id, code)
+            # parse metadata
```
rikhuijzer, line 250: Give answers to "why?" in comments and not answers to "what?". "what?" is already answered by the code. If a comment is used to denote a block of lines doing some bigger task, then consider moving it into a function. Julia inlines functions anyway so during runtime they are mostly free

sample/notebook_with_metadata.jl
```diff
@@ -0,0 +1,30 @@
...
+        el
+    end
+end
+
+# ╔═╡ 7d550d46-7dc4-11eb-3df0-db7ad0594043
+# ╠═╡ disabled = "directly"
+using PlutoUI
+
+# ╔═╡ 8399ed34-7dc4-11eb-16ec-e572834e149d
+# ╠═╡ disabled = "indirectly"
```
rikhuijzer, line 21: Probably there is a good reason for this, but can't the indirectly disabled be derived from the topology? Or is the topology disabled when a cell is disabled?

src/evaluation/Run.jl
lungben: Yes, I forgot about the increased Julia version requirement.

sample/notebook_with_metadata.jl
lungben: The indirectly status is derived from topology when the notebook is executed in Pluto. But the notebook.jl file should also run as a stand-alone Julia script. For this purpose, cells dependent on deactivated cells must be commented out, too.

The sample notebook, however, needs to be updated because it does not have the disabled cells commented out yet.

rikhuijzer: > For this purpose, cells dependent on deactivated cells must be commented out, too.

Nice!

lungben: Sample notebook is now updated.

The slider is now deactivated, not the `using PlutoUI` cell, because Pluto cannot figure out cells dependent on using calls (therefore using calls are always executed first).

src/notebook/Notebook.jl
lungben: While I agree that separating code into smaller functions is in general a good style, it would be different to the style of the existing code in this file. In addition, the code in this block has multiple output variables, making it harder to read than just having the code in a block. And for code blocks, I think a short high-level statement what the block does (like section headers in a book) is helpful for orientation.

src/notebook/Cell.jl
```diff
@@ -64,3 +66,16 @@ end
 function Base.convert(::Type{UUID}, string::String)
     UUID(string)
 end
+
+function get_cell_metadata(cell::Cell)::Dict{String,Any}
+    Dict(
+        if cell.running_disabled
```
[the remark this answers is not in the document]
lungben: The first thing I can definitively do.

Regarding removing `depends_on_disabled_cells `and `running_disabled` - this may cause a performance impact because this information is used on every cell evaluation, and the metadata dictionary is not concretely typed.
No idea if this makes a significant difference, however.

src/evaluation/Run.jl
```diff
@@ -54,14 +55,9 @@ function run_reactive!(session::ServerSession, notebook::Notebook, old_topology:
...
-    deactivated = filter(c -> c.running_disabled, notebook.cells)
-    indirectly_deactivated = collect(topological_order(notebook, new_topology, deactivated))
-    for cell in indirectly_deactivated
-        cell.running = false
-        cell.queued = false
-        cell.depends_on_disabled_cells = true
-    end
-
+	if indirectly_deactivated === nothing
+		indirectly_deactivated = disable_dependent_cells!(notebook, new_topology)
```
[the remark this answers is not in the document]
lungben: This may be a piece of premature optimization on my side ;-) 
I'll try your last suggestion to just re-call `disable_dependent_cells!`. If it turnes out to be a performance bottleneck we can still optimize it later.

src/notebook/Cell.jl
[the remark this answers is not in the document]
Pangoraw: My idea was that we could just give a single metadata API to Javascript/Julia, and cell disabling would be handling directly as a result but we can still change this later.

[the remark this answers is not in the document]
lungben: Sounds good.

The metadata is now published to the frontend in `cell_input.metadata`. Changing the cell disabling from using separate fields to metadata itself has quite a large impact, therefore I would suggest to do it later (if required at all).

fonsp:
frontend tests will give ✅ again if you update from `main`!

src/notebook/Notebook.jl
fonsp: I think you are both right :)

sample/notebook_with_metadata.jl
fonsp: > because Pluto cannot figure out cells dependent on using calls

That actually got fixed by Paul! https://github.com/fonsp/Pluto.jl/pull/1541 But it's a new type of reference (see the PR), which does not carry the "disabling" satus.

fonsp:
*EDIT:* I am wondering the same as Rik in this comment: https://github.com/fonsp/Pluto.jl/pull/1895#discussion_r802963760

We should store directly-disabled status in the file, but do we need to store the *indirectly*-disabled status in the file? We [already remove the Pluto-created comment when reading the notebook file, without checking whether it was (indirectly) disabled](https://github.com/fonsp/Pluto.jl/pull/1895/files#diff-72a3233e2574455ad35c1d723316dd0f0c7646e560f50f5f81934c30a0729565R245) (👍), so it's not necessary to read the file correctly, right?

Why I think this is important: if we *don't* store indirectly-disabled status in the file, then:
1. It doesn't need to be in `metadata`
1. `metadata` can just have a *boolean* field: `disabled` 🎉
1. This means that `metadata` does not contain "cache values", which I think is a problem (see below) 🎉
1. This means that we don't need `cell.running_disabled` anymore! No more double state! When clicking the "Disable" button, the frontend will set the `cell.metadata.disabled` value directly. (It's already synced, right?) 🎉

---

*Here is my original comment about reducing double state*

Right now, `metadata` is a pure function of `running_disabled` and `depends_on_disabled_cells`, which means that we don't need to store the value. Right now, all the links between `metadata` and the two boolean fields are manual: whenever anything changes, we manually sync between the dict and the booleans. That's not optimal! And it means that working with metadata or cell disabling will become more difficult (we would have to write comments like: *"if you change `cell.disabled`, remember to also change `cell.metadata["disabled"]` to a different string!"* 🤔)

I think it's a really good idea to have the metadata system be part of the implementation of #1205, but we should really go for it! Instead of using `metadata` as an "output", let's use `metadata` as a new, more flexible way to store state!

lungben:
I agree that storing the information about indirectly deactivated cells in the notebook file adds siginificant complexity and is not required for storing cell deactivation status (it is even not read when opening a notebook).

Currently, its only use is to comment out dependent cells in the notebook file so that the notebook running standalone has the same cells disabled as the notebook running in Pluto.

For commenting out disabled cells, there are the following options:

1. Comment out only directly deactivated cells - disadvantage: the notebook likely crashes when running standalone because undefined variables are used in the code.
2. Comment out directly and indirectly deactivated cells (current implementation) - this allows having the same cell disabling status in standalone scripts as in Pluto, but this adds complexity to metadata management.
3. Do not comment out any cells in notebook files, regardless of their disabling status - this would make metadata management easier, but notebooks would behave standalone different to running in Pluto (only for the latter cell disabling status would matter).

I think option 1. is not really an option, but option 3. would be an alternative.

What do you think?

fonsp:
Hey Benjamin!

I think that commenting out disabled *and* indirectly disabled cells in the file is great! So option 2!

But right now we also store `disabled = "indirectly"` in the file, and my suggestion is to remove this, but still comment out the cell.

Right now, on `main`, we have:
```julia
# ╔═╡ 8399ed34-7dc4-11eb-16ec-e572834e149d
@bind x Slider(1:10)

# ╔═╡ 88c83d42-7dc4-11eb-1b5e-a1ecad4b6ff1
x
```

Currently, on this PR, we have:
```julia
# ╔═╡ 8399ed34-7dc4-11eb-16ec-e572834e149d
# ╠═╡ disabled = "directly"
#=╠═╡
@bind x Slider(1:10)
  ╠═╡ =#

# ╔═╡ 88c83d42-7dc4-11eb-1b5e-a1ecad4b6ff1
# ╠═╡ disabled = "indirectly"
#=╠═╡
x
  ╠═╡ =#
```

But I suggest the following:
```julia
# ╔═╡ 8399ed34-7dc4-11eb-16ec-e572834e149d
# ╠═╡ disabled = true
#=╠═╡
@bind x Slider(1:10)
  ╠═╡ =#

# ╔═╡ 88c83d42-7dc4-11eb-1b5e-a1ecad4b6ff1\
#=╠═╡
x
  ╠═╡ =#
```

In the second cell, `# ╠═╡ disabled = false` is optional, because this is the default value.

I should have started my message with this example 🙃

lungben:
Yes I see - code is changed accordingly.

fonsp:
Thanks! I still think we could remove `cell.running_disabled` altogether, and store in it `metadata` instead, that's what i meant with the part below the line in https://github.com/fonsp/Pluto.jl/pull/1895#issuecomment-1040894204

ederag:
Minor comment: now the indirectly disabled cells comment blocks do not have any explicit explanation,
which might surprise some users, or make old commit changes a bit less readable
("where is this block comment coming from ?"),
but I agree, it is a good tradeoff with complexity :+1:.

I tried to play with this, with some file editing (removing comments, adding or removing disable lines),
and it looks great so far.
Thanks for listening, the diffs are as lean as possible.
It feels robust,
and I like that you only have to add a `# ╠═╡ disabled = true` line to start the notebook with the cell disabled
(no need to add block comments by hand).
Thanks !

fonsp:
Thanks for testing out the file editing @ederag !

j-fu:
In which respect is this better than #421 (I don't mind at all to have my idea dismissed, but modifiying the notebook format ist something which really needs to be considered carefully, so I just remind it).

lungben:
This PR essentially implements #421 on cell level, i.e. the metadata for each Pluto cell is its own TOML "file" (written as comment before the cell in the notebook.jl file). 

Suggested in #421, but not implemented here are:

1. Metadata on notebook level - I think this would be really helpful, but should be done in a separate PR.
2. Migration of existing cell metadata (UUID, hidden) to the new TOML metadata format. I agree that it would be nicer to have this also in TOML format, but this would be a breaking change in the notebook format - new notebooks would not work with old Pluto versions anymore. This PR is non-breaking, when using an old Pluto version the new metadata is just shown as a comment and ignored.

j-fu:
Ok I understand - probably this is the way to go in the moment.
Yeah agree for 1.

For 2: I can imagine that quite lot of stuff can become part of cell metadata, so a newer "rich metadata notebook" at certain point will not make much sense for an old pluto version.

I guess the reading code for the current notebook format will stay forever, and so can the writing code, which could be invoked when opening a notebook in "legacy format." if an author wants the ability to run a notebook under old pluto versions.

We had a bit of a similar situation for the Manifest.toml...

fonsp:
I think this is a really good middle ground! We get the future-proofness of using TOML for metadata (thanks for the suggestion!), which means that we won't have to struggle in the future on how to add new metadata items into the file format, but without a breaking change to the notebook format. 

The comparison with the Manifest.toml change is a good one! I think they handled it well by adding future-proof compatibility code in 1.6, and backwards-proof compat code in 1.7 and later.

But! I think we might be able to avoid a breaking file format change, because we have more flexibility than the original Manifest.toml did (which was the reason for the new format): for example, we can put anything next to the `macro bind` definition, everything above the first cell is ignored by old versions of Pluto. Another path is to "add cells at the bottom" with a special UUID, just like we did for #844 .

pankgeorg:
@fonsp @dralletje is this good to go?

dralletje:
Is it okay if I remove the frontend changes (and remove disabled_cell as it was before), feel like it would make the PR a lot cleaner 😁

lungben:
The frontend changes currently do 2 things:
1. make metadata available in the frontend and synchronize it with the backend
2. remove the explicit "cell_running_disabled" boolean variable and use the corresponding metadata instead.

While 1. is not required to store the cell disabling status, it is the basis for any future enhancements in the frontend using the new metadata system. Therefor I suggest to keep it in this PR.

dralletje:
> The frontend changes currently do 2 things:
> 
> 1. make metadata available in the frontend and synchronize it with the backend
> 2. remove the explicit "cell_running_disabled" boolean variable and use the corresponding metadata instead.
> 
> While 1. is not required to store the cell disabling status, it is the basis for any future enhancements in the frontend using the new metadata system. Therefor I suggest to keep it in this PR.

I want to keep 1. for sure! Just would like to also expose `cell_running_disabled` to the frontend as it was before (I'm afraid of frontend bugs :P)

Pangoraw:
Some motivations for 2. in https://github.com/fonsp/Pluto.jl/pull/1895#discussion_r806296243. Maybe we can namespace (under metadata.pluto.disabled) the disabled property in the metadata?

dralletje:
> Some motivations for 2. in [#1895 (comment)](https://github.com/fonsp/Pluto.jl/pull/1895#discussion_r806296243). Basically, if there is an API to get/set metadata from Julia/JS then there is going to be less logic if everything already refers to the metadata. Maybe we can namespace (under metadata.pluto.disabled) the disabled property in the metadata?
> 
> As a follow up it would be great to persists logs hiding for example.

I only wanted to keep the frontend changes to a minimum (because they span a lot of different files), but if you are confident it works now, then having it migrated now is very nice :)

ederag:
@dralletje Although the file format was the main focus of https://github.com/fonsp/Pluto.jl/pull/1895#issuecomment-1055285677, the user interface seemed fine as well
(as usual, quality work @lungben !)
Anyway, if you could precise what you mean by "frontend", then I'd be ready to perform more tests this week-end.

frontend/components/Cell.js
```diff
@@ -167,6 +168,7 @@ export const Cell = ({
                 selected: selected,
                 code_differs: class_code_differs,
                 code_folded: class_code_folded,
+                metadata: metadata,
```
pankgeorg, line 171: This is just for the css metadata class, if the cell has metadata, but it's always true, so I'm removing for now! If you had a use case in mind, shout!

pankgeorg:
Michiel's concern is real, pretty much this comment: https://github.com/fonsp/Pluto.jl/pull/1895/files#diff-6f965d00041ae73bc1bc31832951173264ef6f72375983989b13d99069f5aaf6R66
I'm for keeping the metadata as @lungben does, that along with the cell_memo will operate correctly for now. When we add more metadata, we'll need to add them to the `CellMemo` dependencies.

A quick word for my (very small!) updates:
-> on the backend, two things
1. metadata now follows the DEFAULT_METADATA "schema": cells will intialize to this dict and get updates accordingly
2. cells will deserialize _removing_ the DEFAULT_METADATA pairs from the TOML writer input dict. So `"disabled"`: `false` and every other metadatum we add will be removed there if it matched the process of [1]
-> On the front-end I fixed the possible invalidation issue and adjusted the code to _expect_ the metadata.default to exist and be false (and then true)
finally, we fix a small saving issue: I noticed that a cell needed to run after deactivating something, to get the right stuff saved. That was because we saved very early; we need to save _after_ the last field that gets serialized changes. Before that was `cell_order`, but now it is `depends_on_disabled_cells`. If more stuff get into the serialization of the notebook, that save (now on line 87) will need to be moved further down.

@lungben this is amazing work!
@lungben @fonsp @dralletje can you do a quick check so we get this into #main?

lungben:
Thanks!
Your changes look fine for me.

frontend/components/Cell.js
lungben: In principle, cell formatting could in the future be dependent on cell metadata, e.g. allowing different cell widths. But if something like this is introduced, passing metadata to the class could be re-introduced.

pankgeorg: Yes, and most likely it will be something inside the metadata, not the whole object (which now is always true-ish, so cells get a "metadata" class) let's add specific data from metadata like "cell-width" for example etc

PR #1896, opened 2022-02-04
With Distributed 2

pankgeorg:
Monkey patch distributed 😇


icweaver:
> Monkey patch distributed 

My brain keeps reading this sentence as you handing out patches with little monkeys on them to people, and I am ok with that

PR #1898, opened 2022-02-04
Send queued bond changes when last script finishes

fonsp:
Part of #1892 .

`is_notebook_idle()` checks (among other things) if no scripts are currently running. Queued bond updates are sent from `componentDidUpdate` in Editor.js, *if `is_notebook_idle()` returns `false`*. 

This creates an edge case where an async script takes a little while to run, and by the time it finishes, bond updates have queued up, but there is nothing triggering `componentDidUpdate` to send them. Bonds are then "stuck" until you interact with the notebook in another way (like typing somewhere).

# Solution

A callback that is fired when the last script is finished. On this callback, queued bond values are sent (if idle).


fonsp:
Pretty harmless PR because it's safe to call `this.send_queued_bond_changes` too often

PR #1911, opened 2022-02-11
Fix logging getting stuck when `maxlog` is specified and don't print `maxlog`, fixes #1902

### LINKED ISSUE #1902: Notebooks hangs when `maxlog` is used in logging and includes `maxlog` in output
https://user-images.githubusercontent.com/13287213/153057977-00c3b27e-5d57-45ea-a6ce-c0880ce5a233.mp4

Support for `maxlog` in logging was recently added in #1877. However I have issues with Pluto getting stuck on the second run of a cell with `maxlog` specified. In the video I stop after 10 seconds, but it doesn't finish after that either. Link to [notebook](https://gist.github.com/Joel-Dahne/202469a19f543575b8d634448f93d6fe).

One part of the code added in #1877 was
```julia
maybe_max_log = findfirst(((key, _),) -> key == "maxlog", next_log["kwargs"])
if maybe_max_log !== nothing
    n_logs = count(log -> log["id"] == next_log["id"], running_cell.logs)
    try
        max_log = parse(Int, next_log["kwargs"][maybe_max_log][2] |> first)

        # Don't show message with id more than max_log times
        if max_log isa Int && n_logs >= max_log
            return
        end
    catch
    end
end
```
I was able to get it to work for be by changing the `return` to `continue`. However I don't understand enough of Pluto's code to know if this is the proper fix. 

I would also propose to follow the default for logging in Julia and not include the `maxlog` argument in the log message
```julia
julia> for i in 1:10
           @info "Some info" i maxlog = 2
       end
┌ Info: Some info
└   i = 1
┌ Info: Some info
└   i = 2
```
I was able to get this behavior by adding the line
```julia
deleteat!(next_log["kwargs"], maybe_max_log)
``` 
inside the `try`-statement. But again I don't know enough about Pluto to know if editing the log-message is allowed.

And thank you for adding support for logging! I think it can be very helpful!

Joel-Dahne:
This is an attempt at fixing #1902.  It contains two commits. The first one fixes it so that it doesn't get stuck if multiple logs are encountered when `maxlog` is used on one of them. The second one removes the `maxlog` from the log-message, in line with how Julia handles it. It also adds tests for both of these behaviors.

This is the first time I'm contributing to Pluto so there are some things it would be good if they could be check.
1. At the moment it deletes the `maxlog` element from the log to not have it be printed in the log-message. I'm not sure if this is the right way to do it. I don't know if updating the log is okay, maybe it is used in other places as well? Maybe it is better to exclude `maxlog` from the message in the frontend instead of backend?
2. When writing the tests I mimicked the one existing test. However I don't understand the use of `poll` so I might have misused it.

### DIFF
--- a/src/evaluation/WorkspaceManager.jl
+++ b/src/evaluation/WorkspaceManager.jl
@@ -169,6 +169,10 @@ function start_relaying_logs((session, notebook)::SN, log_channel::Distributed.R
                 try
                     max_log = parse(Int, next_log["kwargs"][maybe_max_log][2] |> first)
 
+                    # Don't include maxlog in the log-message, in line
+                    # with how Julia handles it.
+                    deleteat!(next_log["kwargs"], maybe_max_log)
+
                     # Don't show message with id more than max_log times
                     if max_log isa Int && n_logs >= max_log
                         continue
--- a/src/evaluation/WorkspaceManager.jl
+++ b/src/evaluation/WorkspaceManager.jl
@@ -171,7 +171,7 @@ function start_relaying_logs((session, notebook)::SN, log_channel::Distributed.R
 
                     # Don't show message with id more than max_log times
                     if max_log isa Int && n_logs >= max_log
-                        return
+                        continue
                     end
                 catch
                 end
--- a/test/Logging.jl
+++ b/test/Logging.jl
@@ -26,6 +26,16 @@ using Pluto.WorkspaceManager: poll
             @test poll(5, 1/60) do
                 length(notebook.cells[begin].logs) == 2
             end
+
+            # Check that maxlog doesn't occur in the message
+            @test poll(5, 1/60) do
+                all(notebook.cells[begin].logs) do log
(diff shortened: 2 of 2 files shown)


fonsp:
I added documentation for `poll`, hope that helps! https://github.com/fonsp/Pluto.jl/commit/a14bcbabf3cca932a38d7d2d7d832240e3d12fb2

fonsp:
Thanks so much for adding tests, really helpful!!

Joel-Dahne:
> I added documentation for `poll`, hope that helps! [a14bcba](https://github.com/fonsp/Pluto.jl/commit/a14bcbabf3cca932a38d7d2d7d832240e3d12fb2)

Armed with this knowledge I slightly rewrote the tests!

Joel-Dahne:
I'm not sure if the failed tests are related?

fonsp:
Perfect, thanks so much!

PR #1920, opened 2022-02-14
Store cell order in NotebookTopology

fonsp:
I'm trying to decouple things... small steps towards an execution model where we keep two separate "states": the *actual evaluated* state and the *desired state*. `update_save_run!` should just update the actual state to become closer to the desired state.

This PR makes (the fields of) `NotebookTopology` immutable, and importantly, we now store cell order in the topology! This means that, in a lot of places, things become a pure function of `NotebookTopology`, and you no longer need to also pass in the `Notebook`. 

Currently in these functions, `NotebookTopology` is used for almost everything, and `Notebook` was only used to iterate "all cells in order". By moving a (shallow) copy of `notebook.cells` into the topology, we don't need the `notebook` anymore in a lot of places. This also removes one hidden link between a topology and its notebook.

This should also help with #1919 

## Future

Some ideas for future PRs:
- `NotebookTopology` is still mutable because it contains `Cell`s... I think this sort of makes sense: we should keep around a reference to the cell, so that we can change its output after running the cell.
- Just like in the frontend, cell *result* and cell *input* data could be split, and both should be immutable. Then a `Cell` is just a container with two fields: `Ref{CellInputData}`, `Ref{CellResultData}`. Or maybe we don't need the `Cell` anymore?
- I feel like we can take a good **halfway** step towards the first paragraph of this PR.


fonsp:
@Pangoraw what do you think?

Pangoraw:
Looks good! 👍

PR #1859, opened 2022-01-27
📎 Dynamic code hints: warn when creating a multiple definition

fonsp:
https://user-images.githubusercontent.com/6933510/151268574-6e102be9-b3f9-43ef-9d27-938725ccadcd.mov

The warning system is more general, and can be used for other syntax-matching hints.

PR #1925, opened 2022-02-16
local variable autocomplete

Pangoraw:
This collects all local variables in a new scopestate field along with their "validity" range (range in which they are valid). Can be used for autocompletion of local variable and enhanced linting in #1859 (also potentially renaming of variables). @dralletje, what do you think about this approach ? (lezer-matcher is insane btw :exploding_head:)

autocomplete_locals

### DIFF
--- a/frontend/components/CellInput/pluto_autocomplete.js
+++ b/frontend/components/CellInput/pluto_autocomplete.js
@@ -174,12 +174,15 @@ let override_text_to_apply_in_field_expression = (text) => {
  * @param {number} context_pos
  */
 const generate_scopestate_completions = function* (definitions, proposed, context_pos) {
+    let i = 0
     for (let [name, { valid_from }] of definitions.entries()) {
         if (!proposed.has(name) && valid_from < context_pos) {
             yield {
-                label: "Scopestate = " + name,
-                apply: name,
+                label: name,
+                type: "c_Any",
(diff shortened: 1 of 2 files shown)


Pangoraw:
i would like to fix this before merging:

https://github.com/fonsp/Pluto.jl/blob/2e02b821c5fefef158fae8180fe3fd1dfc139419/frontend/components/CellInput/scopestate_statefield.js#L379-L381

Pangoraw:
Should be good to go!

fonsp:
Awesome!

I noticed one issue: if I autocomplete `s` with a local var that starts with `s`, it selects that one immediately.

My guess is that the real Julia autocomplete is taking too long in this case (lots of options with `s`), and the autocomplete window pops up with just the `sqrt2`, which immediately triggers.

https://user-images.githubusercontent.com/6933510/159265990-a83c0964-cf63-4d6d-8479-8724e99fc71a.mov

fonsp:
@dralletje pointed out that this has a big performance impact because it makes `raise_scope` slower, and that function is already a bottleneck. Since I already merged this, we might need to uncomment some this feature on `main` while we work on performance :(

fonsp:
Enabled again in https://github.com/fonsp/Pluto.jl/pull/3151

Issue #1913, opened 2022-02-12
Editor froze: maximum call stack in scoped statefield


fonsp:
I was just copy pasting some code around, and suddenly the tab froze. Was not able to reproduce

<img width="585" alt="Schermafbeelding 2022-02-12 om 14 00 29" src="https://user-images.githubusercontent.com/6933510/153725772-e35a978d-b770-4b9a-af10-86e76fdd2113.png">

dralletje:
😱

fonsp:
I got it a second time when changing this:
```julia
reinterpret(Vector{SVector{Float64,2}}, aa)
# into
reinterpret(10, SVector{Float64,2}, aa)
```

I remembered the exact keystrokes that I did (I think), but after refreshing and trying many times to repeat it, it did not crash...

fonsp:
Happened again, on current `main`:
```
function (::Colon)(a::StaticArrays.SVector{S, <:Real}, step::Real, b::StaticArrays.SVector{S, <:Real}) where S
	diff = b - a
	diff_norm = norm(diff)
	(a + x*diff for x in 0:step/diff_norm:1)
end
```
<img width="588" alt="Schermafbeelding 2022-02-13 om 20 56 20" src="https://user-images.githubusercontent.com/6933510/153772405-5fc2a890-8745-467a-9f9b-d5b699156581.png">

fonsp:
Can we at least `try-catch` here? This seems like a pretty bad bug

fonsp:
https://user-images.githubusercontent.com/6933510/153776419-80e9e80b-f314-42f1-be1c-4e1295d17b45.png

dralletje:
I'd love to, but I don't know where it is that it breaks D:
Maybe disable the mixedparser once more? :/

pankgeorg:
> I'd love to, but I don't know where it is that it breaks D: Maybe disable the mixedparser once more? :/

We'll probably have to, but this code doesn't have something that's getting mixed

fonsp:
[...]

Issue #1893, opened 2022-02-03
CM6 incremental parsing for mixed modes


fonsp:
My computer was really slow today, and CM6 became glitchy inside JS. I think that CM6 is not using the incremental parse system inside the submodes, only in Julia.

These videos were taken in a really big notebook (https://github.com/ClimateMARGO/interactive) with 6x CPU slowdown (chrome devtools), which is similar to how it felt before rebooting :)

# JS

https://user-images.githubusercontent.com/6933510/152447878-6a60e1d3-b231-498f-9e9b-8f62fcc29cad.mov

# Julia

https://user-images.githubusercontent.com/6933510/152447922-100f2d08-6232-4cc6-a0f9-d6b48b4b7f81.mov

ilyagr:
I anyone is looking at this, there is a  "secret" JS function to toggle this. If you run `window.PLUTO_TOGGLE_CM_MIXED_PARSER()` in the JS console, mixed modes will turn on and this will be remembered for future Pluto invocations.

A minor bug report: the mixed parser does not work for `sql"""` strings. To be more precise, the color of `sql"""` strings is *different* than that of normal strings, but there is no SQL syntax highlighting. This is strange, since judging by the code I read it should work.

fonsp:
This is now a setting: #3556 !

fonsp:
The performance issue should have been fixed by https://github.com/JuliaPluto/Pluto.jl/pull/3552 but I'm not completely convinced...

PR #1940, opened 2022-02-22
Disable CM mixed parser

fonsp:
Because of the performance problems. @dralletje is working on this (#1893), but in the meantime, let's disable it (again 😥 ), to give us some time to get it right.

This *should* also be a workaround for #1913


fonsp:
Some promising news! https://github.com/JuliaPluto/lezer-julia/pull/7

PR #1941, opened 2022-02-22
Run file watching handler in async thread?

fonsp:
https://github.com/fonsp/Pluto.jl/pull/1941/files#diff-8712103475dc8f154593213d454f24645e17ec5019f0a0f0d792300380c98bc6R144-R145


Pangoraw:
May be of interest, there is an [asynchronous API](https://github.com/JuliaPluto/BetterFileWatching.jl/blob/6c2461ba048bd32098b0dfd8010bb82c002fd0b9/src/libwatcher.jl#L190) in the libwatcher wrapper:

```julia
watcher = BetterFileWatching.subscribe(f, dir; options) # not blocking

# and then later
BetterFileWatching.unsubscribe(watcher)
```

i still need to create libwatcher_jll and then I can start seriously testing on more platforms!

fonsp:
Thanks @Pangoraw !! 

You mentioned this problem once and I just remembered, I thought I would give a quick try of using what you said with the FileWatching stdlib, but it still doesn't fix our tests :( The failing test shows that an event was missed.

I'll close this PR and wait for your good news! 🤩

PR #1944, opened 2022-02-22
Fix #1781 - cell selection/move performance regression

### LINKED ISSUE #1781: Several Performance issues
In general larger notebooks are slow to interact with, at least on my not to new Laptop. This may originate from several smaller performance issues.

To track those down here is a list of surprisingly slow things that can be recreated easily without complicated notebooks. All of them should have nothing to do with compilation time.

- ~~Inside a new cell hold down CTRL+Return to created many new empty cells. At my setup, cell generation takes at least 0.2 seconds and is slow enough that it can't keep up with the speed of the automatic key press repetition. Was introduced before 0.16.4. Still slow but not as bad in 0.15.1.~~ This always has been this "slow" and the delay isn't actually noticable during everyday usage...what where you thinking past-me?
- Create a notebook with many empty cells (for example by holding down CTRL+Return). Try to select multiple cells with the ribbon band. With my setup, the more cells there are, the slower the ribbon reacts. Was introduced between https://github.com/fonsp/Pluto.jl/commit/e867a5ba13820afaf1cbb7c2a9e4817c2f84c831 and https://github.com/fonsp/Pluto.jl/commit/df7de2ae464ea2080f58146cef97d96ca5fa0f43
- Issue #1692. Was introduced between https://github.com/fonsp/Pluto.jl/commit/e867a5ba13820afaf1cbb7c2a9e4817c2f84c831 and https://github.com/fonsp/Pluto.jl/commit/df7de2ae464ea2080f58146cef97d96ca5fa0f43

I will try to narrow down the version when those where introduced later.

pankgeorg:
No features, just reacty optimization stuff and bug fixes

### DIFF
--- a/frontend/components/Cell.js
+++ b/frontend/components/Cell.js
@@ -1,5 +1,5 @@
 import _ from "../imports/lodash.js"
-import { html, useState, useEffect, useMemo, useRef, useContext, useLayoutEffect } from "../imports/Preact.js"
+import { html, useState, useEffect, useMemo, useRef, useContext, useLayoutEffect, useCallback } from "../imports/Preact.js"
 
 import { CellOutput } from "./CellOutput.js"
 import { CellInput } from "./CellInput.js"
@@ -36,7 +36,6 @@ const useCellApi = (node_ref, published_object_keys, pluto_actions) => {
  *  cell_dependencies: import("./Editor.js").CellDependencyData
  *  nbpkg: import("./Editor.js").NotebookPkgData?,
  *  selected: boolean,
- *  selected_cells: Array<string>,
  *  force_hide_input: boolean,
  *  focus_after_creation: boolean,
  *  [key: string]: any,
@@ -48,23 +47,20 @@ export const Cell = ({
     cell_dependencies,
     cell_input_local,
     notebook_id,
-    on_update_doc_query,
-    on_change,
-    on_focus_neighbor,
     selected,
-    selected_cells,
     force_hide_input,
     focus_after_creation,
     is_process_ready,
     disable_input,
     nbpkg,
+    variables_in_all_notebook,
 }) => {
     let pluto_actions = useContext(PlutoContext)
+    const on_update_doc_query = pluto_actions.set_doc_query
+    const on_focus_neighbor = pluto_actions.focus_on_neighbor
+    const on_change = useCallback((val) => pluto_actions.set_local_cell(cell_id, val), [cell_id, pluto_actions])
     const notebook = pluto_actions.get_notebook()
-    let variables_in_all_notebook = Object.fromEntries(
-        Object.values(notebook?.cell_dependencies ?? {}).flatMap((x) => Object.keys(x.downstream_cells_map).map((variable) => [variable, x.cell_id]))
-    )
-    const variables = Object.keys(notebook?.cell_dependencies?.[cell_id]?.downstream_cells_map || {})
+    const variables = useMemo(() => Object.keys(notebook?.cell_dependencies?.[cell_id]?.downstream_cells_map || {}), [cell_id, notebook])
     // cm_forced_focus is null, except when a line needs to be highlighted because it is part of a stack trace
     const [cm_forced_focus, set_cm_forced_focus] = useState(null)
     const [cm_highlighted_line, set_cm_highlighted_line] = useState(null)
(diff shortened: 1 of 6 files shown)


frontend/components/Cell.js
```diff
@@ -48,23 +47,20 @@ export const Cell = ({
...
     disable_input,
     nbpkg,
+    variables_in_all_notebook,
 }) => {
     let pluto_actions = useContext(PlutoContext)
+    const on_update_doc_query = pluto_actions.set_doc_query
+    const on_focus_neighbor = pluto_actions.focus_on_neighbor
+    const on_change = useCallback((val) => pluto_actions.set_local_cell(cell_id, val), [cell_id, pluto_actions])
     const notebook = pluto_actions.get_notebook()
-    let variables_in_all_notebook = Object.fromEntries(
```
pankgeorg, line 64: This creates a new object in each render! We pushed that into the notebook component as it only needs to be calculated once!

frontend/components/Cell.js
```diff
@@ -48,23 +47,20 @@ export const Cell = ({
...
-        Object.values(notebook?.cell_dependencies ?? {}).flatMap((x) => Object.keys(x.downstream_cells_map).map((variable) => [variable, x.cell_id]))
-    )
-    const variables = Object.keys(notebook?.cell_dependencies?.[cell_id]?.downstream_cells_map || {})
+    const variables = useMemo(() => Object.keys(notebook?.cell_dependencies?.[cell_id]?.downstream_cells_map || {}), [cell_id, notebook])
```
pankgeorg, line 63: Classic memoization candidate

frontend/components/Cell.js
```diff
@@ -131,7 +127,36 @@ export const Cell = ({
     const set_waiting_to_run_smart = (x) => set_waiting_to_run(x && should_set_waiting_to_run_ref.current)
 
     const cell_api_ready = useCellApi(node_ref, published_object_keys, pluto_actions)
-
+    const on_delete = useCallback(() => {
```
pankgeorg, line 130: Memoize the callbacks, instead of creating new ones.

frontend/components/Cell.js
```diff
@@ -131,7 +127,36 @@ export const Cell = ({
...
+                }
+                on_change(new_code)
+            }
+        },
+        [code_folded, cm_forced_focus, pluto_actions, on_change]
+    )
+    const on_add_after = useCallback(() => {
+        pluto_actions.add_remote_cell(cell_id, "after")
+    }, [pluto_actions, cell_id, selected])
+    const on_code_fold = useCallback(() => {
```
pankgeorg, line 153: `on_code_fold` and `on_run` are refactored to use new actions that pick the state immediately, instead of passing that down to props

frontend/components/Editor.js
```diff
@@ -463,6 +463,24 @@ export class Editor extends Component {
                 const delta = before_or_after == "before" ? 0 : 1
                 return await this.actions.add_remote_cell_at(index + delta, code)
             },
+            // These actions are basically helpers because pushing selected cells all
+            // the way down is too heavy to handle!
+            fold_code: async (cell_id, selected) => {
```
pankgeorg, line 468: All these new functions are optimizations. Instead of doing the math in each cell, we use this functions instead. That means we can only have one function (this one) that needs to refer to `this`.`state`.`selected_cells` and we don't need to re-memo each time! (we did that once per cell🙈)

frontend/components/Editor.js
```diff
@@ -1344,7 +1359,7 @@ patch: ${JSON.stringify(
                                 on_selection=${(selected_cell_ids) => {
                                     // @ts-ignore
                                     if (
-                                        selected_cell_ids.length !== this.state.selected_cells ||
+                                        selected_cell_ids.length !== this.state.selected_cells.length ||
```
pankgeorg, line 1362: This small bug was devastating! length is always different from array, which means that `onmousemove` changed the `selected_cells` array which means that we re-diffed everything! On the bright side of things, this small bug helped us refactor all the selected stuff and now it's fast even with this bug in!

frontend/components/Notebook.js
```diff
@@ -9,21 +9,17 @@ let CellMemo = ({
     cell_input,
     cell_input_local,
     notebook_id,
-    on_update_doc_query,
```
pankgeorg, line 12: we pick these from `pluto_actions` now!

frontend/components/Notebook.js
```diff
@@ -9,21 +9,17 @@ let CellMemo = ({
...
     focus_after_creation,
     force_hide_input,
     is_process_ready,
     disable_input,
     show_logs,
     set_show_logs,
     nbpkg,
+    variables_in_all_notebook,
 }) => {
-    const selected_cells_diffable_primitive = (selected_cells || []).join("")
```
pankgeorg, line 26: this is one of the places we diffed selected_cells. we don't need to thanks to the new actions!

frontend/components/Notebook.js
```diff
@@ -143,7 +130,13 @@ export const Notebook = ({
             }, render_cell_outputs_delay(notebook.cell_order.length))
         }
     }, [cell_outputs_delayed, notebook.cell_order.length])
-
+    let variables_in_all_notebook = useMemo(
```
pankgeorg, line 133: @dralletje I respected the `let` here, just for you <3

frontend/components/Editor.js
```diff
@@ -1315,9 +1333,6 @@ patch: ${JSON.stringify(
                         <${Notebook}
                             notebook=${this.state.notebook}
                             cell_inputs_local=${this.state.cell_inputs_local}
-                            on_update_doc_query=${this.actions.set_doc_query}
-                            on_cell_input=${this.actions.set_local_cell}
-                            on_focus_neighbor=${this.actions.focus_on_neighbor}
```
fonsp, line 1320: Nice find! We are we passing this down 🤦?

pankgeorg: We were yes - no longer!

dralletje:
Nice

PR #1945, opened 2022-02-22
🏃 Don't use body.classList for ctrl_down to avoid style recalc

fonsp:
We had this feedback on slack:

image

This PR addresses this problem!

When running this giant notebook: https://fonsp-disorganised-mess.netlify.app/big.html with 6x CPU throttle in devtools, pressing and releasing Cmd caused a **500ms synchronous style recalculation**, really bad!

Here is the profile of quickly pressing Cmd once:

image

This was because we were adding/removing the `ctrl_down` class on `<body>`. As it turns out: adding/removing *any* class on body causes an expensive style calculation. Oops!

## First attempt: 
I used a `data` attribute on body instead. This does not trigger a recalculation on its own (👍), but using that data attribute as a selector in CSS `body[data-ctrl-down="true"]` re-introduces the performance problem. 👎

## Second attempt:
Instead of attaching the attribute to body, we do `document.querySelectorAll("pluto-variable-link")` and set the attribute on all links directly! 

This is very fast (because querySelectorAll with a simple selector is fast because of 🌟browser magic🌟), and solves the problem! Note that CM6 only renders code that is inside the viewport, so even in a big notebook, there will never be that many `pluto-variable-link`. 

It has some issues like "holding control and then scrolling a lot reveals links that are not highlighted" but it felt fine.

## Result

Same situation (pressing Cmd once, massive notebook, 6x slowdown):

image

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -974,18 +974,28 @@ patch: ${JSON.stringify(
             this.patch_listeners.forEach((f) => f(patches))
         }
 
+        let ctrl_down_last_val = { current: false }
+        const set_ctrl_down = (value) => {
+            if (value !== ctrl_down_last_val.current) {
+                ctrl_down_last_val.current = value
+                document.body.querySelectorAll("pluto-variable-link").forEach((el) => {
+                    el.setAttribute("data-ctrl-down", value ? "true" : "false")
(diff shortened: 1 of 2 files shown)


dralletje:
Nice

PR #1948, opened 2022-02-23
🧹 Remove service worker

fonsp:
We don't need it anymore for offline support (#1561). In fact, we are not doing any CDN requests anymore, so the SW is not doing anything for released Pluto.

For unreleased Pluto, your browser is already caching CDN requests. Our SW makes caching more aggressive, but what's the point?

Then one last reason to keep the SW: it means more aggressive caching for CDN assets used inside cell outputs. This *might* help when you are on a big international train trip, but not sure how much it actually helps.

@pankgeorg or @dralletje what do you think?


pankgeorg:
Looks good, let's do it!

Issue #890, opened 2021-02-03
Give better error message when two packages export the same function


dpsanders:
Loading two packages that export a function with the same name gives a "not defined" error message in Pluto.
Only in the terminal is the warning shown that they export the same function name and it should be qualified.

fonsp:
Do you have an example for us to test?

dpsanders:
The actual example is IntervalRootFinding.jl and IntervalConstraintProgramming.jl, which both (unfortunately) export `Contractor`.  But ICP takes a while to load, so maybe not the best example.

fonsp:
Fixed by #1957 !

![Schermafbeelding 2022-04-22 om 14 05 56](https://user-images.githubusercontent.com/6933510/164711041-166c8d85-9927-4bbb-873a-6ed2e22d86c2.png)

PR #1957, opened 2022-02-26
📺 Show stdout and stderr in notebook

fonsp:
For backwards-compatibility with ~software written in the 70s~ 😵‍💫 code that still needs to make the switch to Julia's awesome [Logging system](https://docs.julialang.org/en/v1/stdlib/Logging/)! 

I tried to use visual design to address the reasons that we originally had for not supporting stdout in Pluto:
- In the Julia REPL and Jupyter, It can be hard/impossible to understand the difference between something being *returned* and something being *displayed as text*. In a descriptive dataflow system, this difference is important!
- Motivate users to *return values* (much nicer display system), and to split large code into smaller cells.
- Motivate users to use [Logging](https://docs.julialang.org/en/v1/stdlib/Logging/) for logging, not 🖨 printing unstructured text to magic buffer.
- Stdin/stdout is often presented as a "basic feature" of a language, but I disagree! I see it as a niche feature for inter-process communication and super-optimized IO, not something we should teach to beginners. 🤕 

Anyway, it turned out to be *more* confusing/frustrating to silently hide stdout, so here we go! A boring old *terminal* 

Schermafbeelding 2022-02-26 om 00 33 50

*Some common functions that didn't work before!*

---

Schermafbeelding 2022-02-25 om 23 58 24

*You like ASCII art? Then you will love **real** art!*

---

Schermafbeelding 2022-02-25 om 23 57 09

*Oooh is this hacking!?*

---

https://user-images.githubusercontent.com/6933510/155816712-c824ba8b-a13f-473f-b7f1-a4d36c0752df.mov

*Help what is this green stuff!*

xoxo
~ @fonsp


fonsp:
Tests were failing because of race conditions in `main`, I addressed these in main: https://github.com/fonsp/Pluto.jl/compare/8f088f7a9a84d2b98eb2e456bb5b5a2c726dd315...a13644e583382ea147a17aaace4ededb215a9235

diegozea:
Hi! Small question: Will it be a way to hide stdout/stderr without hiding the returned value? In Jupyter is sometimes annoying (and slow) to run processes that fill stdout with a massive amount of text; would that be also a problem with Pluto in the future? Cheers,

fonsp:
Yes, in the ... button next to cell, you have the option to hide logs (this PR turns stdout into a special log). But please try it out and let us know if this helps! Maybe we should also set a maximum size on captured stdout

gdalle:
This is brilliant! Thanks a lot

fonsp:
Fix #890

PR #1968, opened 2022-03-06
fix a bug where the stdout popup throws an error

Pangoraw:
pkg_popup

the pkg popup is instantiated for every event so it handles event from the stdout popup.

https://github.com/fonsp/Pluto.jl/blob/d90e71334bd7f6ec832843881ebd35322a2b0cf4/frontend/components/Popup.js#L57-L61

this is a minimal fix, but maybe it can be safer to instantiate two different popups?


fonsp:
Thanks! Yeah this should be better..

PR #1974, opened 2022-03-11
🧹 refactor: Move initialisation code out of Editor component

fonsp:
Things:
- Move "state initialisation" code from the `Editor.js` component into `editor.js` glue code.
- The `Editor` component now has two props: `launch_params` and `initial_notebook_state`. 

(The extra hope is that this will let me do SSR on the Editor component with https://preactjs.com/guide/v10/server-side-rendering/ !)

PR #1976, opened 2022-03-12
🧩 `<pluto-editor>` component that can be placed inside layout

fonsp:
After this PR, the `editor.html` will look like:

```html
<body class="loading no-MαθJax">

<pluto-editor class="fullscreen"></pluto-editor>

</body>
```

(and the result is exactly as it was before this PR)

If you change it into this:

```html
<body class="loading no-MαθJax">
    
    <div style="
        display: flex; 
        flex-direction: row;
        min-height: 100vh;
        align-items: stretch;
        ">
        
        <div style="
        font-family: system-ui;
        flex: 0 1 16rem;
        background: #282936;
        color: white;
        font-weight: 600;
        z-index: 23400;
        ">
            <ul>
                <li>Home</li>
                <li>About</li>
                <li>Contact</li>
                <li>Chapter 1</li>
                <li>Chapter 1</li>
                <li>Chapter 1</li>
                <li>Chapter 1</li>
                <li>Final stuff</li>
            </ul>
        </div>
        
        <pluto-editor></pluto-editor>
        
    </div>
</body>
```

Then the editor will render inside the `pluto-editor` element! This PR builds on top of #1974 and #1975  and adds the CSS tweaks necessary to make our editor composable like that!

image

The goal is to make the sidebar-left-notebook-right layout possible without putting the notebook in an iframe, like we currently do in https://computationalthinking.mit.edu/Spring21/inverse_climate_model/

It also adds makes a difference between `<pluto-editor>` and `<pluto-editor class=fullscreen>`. fullscreen is what you are used to. In not-fullscreen-mode, the cell drag shoulders are only 20px wide instead of stretching out of the screen on the left, and the editor is always centered horizontally.

---

Now here's something funny! If you then change
```html
<pluto-editor></pluto-editor>
```
into
```html
<pluto-editor></pluto-editor>
<pluto-editor></pluto-editor>
```
then you get this:

https://user-images.githubusercontent.com/6933510/157995819-1a7ef26b-eea8-4b7f-b6b2-07c74cee29d6.mov

# Future

Hey..... what if `<pluto-editor>` was an actual custom web component? Right now it takes in launch parameters from the URL (like the notebook ID to connect to) or from the `window.pluto_...` variables, but those can be props! Then you can embed Pluto notebooks on random pages!

# todo

- [x] cell rectangle select broken
- [x] new notebook does not fill screen vertically
- [x] "Save changes" button disappeared

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -1217,7 +1217,7 @@ patch: ${JSON.stringify(
                     <${PlutoJSInitializingContext.Provider} value=${this.js_init_set}>
                     <${Scroller} active=${this.state.scroller} />
                     <${ProgressBar} notebook=${this.state.notebook} binder_phase=${this.state.binder_phase} status=${status}/>
-                    <header className=${export_menu_open ? "show_export" : ""}>
+                    <header id="pluto-nav" className=${export_menu_open ? "show_export" : ""}>
                         <${ExportBanner}
                             notebookfile_url=${this.export_url("notebookfile")}
                             notebookexport_url=${this.export_url("notebookexport")}
--- a/frontend/components/SelectionArea.js
+++ b/frontend/components/SelectionArea.js
@@ -38,16 +38,34 @@ const in_request_animation_frame = (fn) => {
 export const SelectionArea = ({ on_selection, set_scroller, cell_order }) => {
     const mouse_position_ref = useRef()
     const is_selecting_ref = useRef(false)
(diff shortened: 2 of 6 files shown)


fonsp:
@greimel take a look! The goal is that this will allow us to remove the need for an iframe, which means that you will only need one HTML file per notebook.

dralletje:
AWESOME

fonsp:
I thought it would be cool to make this an actual web component, github copilot thought so too!

image

PR #1959, opened 2022-02-28
Add a benchmark for compilation times

rikhuijzer:
Want meten is weten (because measuring is knowning). I've tried to do something fancy like https://github.com/rikhuijzer/PrecompileMacro.jl/issues/1, but that will take some (a lot of) time to make. So, as a first step, this PR just adds a separate benchmark for compilation time. I think that I've missed a lot of entry points, but I hope that once we are able to pin down what spots are compiled exactly, then we can start making progress on improving the performance.

PR #1977, opened 2022-03-12
add precompile statements to sessionactions

fonsp:
Co-Authored-By: Rik Huijzer <20724914+rikhuijzer@users.noreply.github.com>

Split out from #1959

### DIFF
--- a/src/webserver/SessionActions.jl
+++ b/src/webserver/SessionActions.jl
@@ -65,6 +65,7 @@ function open(session::ServerSession, path::AbstractString; run_async=true, comp
     try_event_call(session, OpenNotebookEvent(nb))
     nb
 end
+precompile(open, (ServerSession, String))
 
 function add(session::ServerSession, nb::Notebook; run_async::Bool=true)
     session.notebooks[nb.notebook_id] = nb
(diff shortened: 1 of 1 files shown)


fonsp:
Before: https://github.com/fonsp/Pluto.jl/runs/5521590989?check_suite_focus=true#step:4:1122
After: https://github.com/fonsp/Pluto.jl/runs/5521595334?check_suite_focus=true#step:4:1122

That's surprising, almost no difference in allocations? So I guess that https://github.com/fonsp/Pluto.jl/pull/1934 already did most of the work.

But nice! It looks like we can really easily compare PRs now, thanks Rik!

rikhuijzer:
> Before: https://github.com/fonsp/Pluto.jl/runs/5521590989?check_suite_focus=true#step:4:1122 After: https://github.com/fonsp/Pluto.jl/runs/5521595334?check_suite_focus=true#step:4:1122
> 
> That's surprising, almost no difference in allocations? So I guess that #1934 already did most of the work.
> 
> But nice! It looks like we can really easily compare PRs now, thanks Rik!

Yeah sorry. I commented in that different PR that the difference for me locally was 0.4 seconds. Not really much, but on the other hand they are basically for free. More gains for `SessionActions.open` require diving into what `Core.Compiler.Timings.ROOT in typeinfer.jl:73` is doing (see the last image in https://github.com/fonsp/Pluto.jl/pull/1934). I'm planning on diving into that

PR #1978, opened 2022-03-12
🧂 more precompile for dynamic functions!

fonsp:


fonsp:
@rikhuijzer i realised i don't really know what i'm doing 😅 can you take a look at adding more precompiles for our dynamic things?

rikhuijzer:
> @rikhuijzer i realised i don't really know what i'm doing sweat_smile can you take a look at adding more precompiles for our dynamic things?

Yeah sure. Gladly. Can you tell give in code or pseudocode are list of methods are called? Or, I can just see what the firebasey sends to the backend and trigger those?

fonsp:
`Firebasey.diff` gets called on two outputs of `notebook_to_js`, and it recursively calls itself on their children to compare. Run the Firebasey notebook to see what I mean

src/webserver/WebServer.jl
```diff
@@ -323,6 +324,7 @@ function run(session::ServerSession, pluto_router)
         end
     end
 end
+precompile(run, (ServerSession, HTTP.Router))
```
rikhuijzer, line 327: Also doesn't have a positive effect I'm afraid.

rikhuijzer:
I've been going deeper on `Pluto.run` since we should be able to shave some time off there. So, similarly to https://github.com/fonsp/Pluto.jl/pull/1934, I've used SnoopCompile to figure out where the time is spent. That gives this flamegraph:

image

I've obtained this by running the following notebook:

```julia
begin
    using Pkg
    
    Pkg.activate(; temp=true)
    Pkg.add([
        "SnoopCompile",
        "ProfileSVG"
    ])
    Pkg.develop(; path="/home/rik/git/Pluto.jl")
end
```

```julia
begin
    using SnoopCompile
    using ProfileSVG
    using Pluto
end
```

```julia
tinf = @snoopi_deep begin
	port = 13435
    options = Pluto.Configuration.from_flat_kwargs(; port, launch_browser=false, workspace_use_distributed=true, require_secret_for_access=false, require_secret_for_open_links=false)
    🍭 = Pluto.ServerSession(; options)
    server_task = @async Pluto.run(🍭)

    retry(; delays=ExponentialBackOff(n=5, first_delay=0.5)) do
        HTTP.get("http://localhost:$port/edit").status == 200
    end
    server_task
end
```

```julia
let
     g = flamegraph(tinf)
     ProfileSVG.view(g)
end
```

This flamegraph is good news. All inference is rooted on the same block, so if we can get the right `precompile` statement somewhere on `from_flat_kwargs`, then we can possibly get some proper improvements. I'll fiddle with it and report back here once I got it.

rikhuijzer:
Great. Adding `precompile(from_flat_kwargs, ())` actually makes it take 0.8 seconds **longer**. The graph shows this:

image

The rest of the time is spent in `Configurations.from_field_kwargs`, so to improve things we have to look there.

rikhuijzer:
As another possibility, maybe we should just build `PackageCompiler.jl` into Pluto? Like,

```julia
julia> using Pluto

julia> Pluto.compile()

julia> Pluto.run()
```
to have a super speedy Pluto. Only need to call `Pluto.compile` once after updating Pluto. And to have the normal Pluto, use
```julia
julia> using Pluto

julia> Pluto.run()
```

src/webserver/WebServer.jl
```diff
@@ -67,6 +67,7 @@ function run(; kwargs...)
     options = Configuration.from_flat_kwargs(; kwargs...)
     run(options)
 end
+precompile(run, ())
```
rikhuijzer, line 70: This one works. Although I cannot measure `Pluto.run()` easily, I can measure `Configuration.from_flat_kwargs` which is the biggest source of latency according to SnoopCompile

## Without `precompile(run, ())`

```julia
$ julia --project -ie 'using Pluto'
...

  2.158083 seconds (3.60 M allocations: 194.337 MiB, 7.19% gc time, 99.96% compilation time)
```

## With `precompile(run, ())`

```julia
  1.431652 seconds (2.64 M allocations: 143.595 MiB, 11.35% gc time, 99.92% compilation time)
```

src/webserver/WebServer.jl
rikhuijzer, line 327:
```suggestion
```
You can test the precompile directive in the REPL to be sure that it works. This one doesn't:

```julia
julia> precompile(Pluto.run, (Pluto.ServerSession, HTTP.Router))
false
```
because not all types are concrete
```julia
julia> isconcretetype(HTTP.Router)
false
```

rikhuijzer:
With JET.jl, I've been digging down in why `SessionActions.open` keeps taking 8 seconds even though `precompile` is called on it. I've found one example of a problem. The compiler has a really bad time on `maybe_macroexpand`:

```julia
julia> using Pluto

julia> julia> precompile(Pluto.ExpressionExplorer.maybe_macroexpand, (Expr,))
true

julia> @time @eval Pluto.ExpressionExplorer.maybe_macroexpand(:(x = 1))
  0.213755 seconds (397.54 k allocations: 20.237 MiB, 31.72% gc time, 100.06% compilation time)

julia> @time @eval Pluto.ExpressionExplorer.maybe_macroexpand(:(x = 1))
  0.000607 seconds (75 allocations: 3.484 KiB)
```

So that's more than 100% compilation time :rofl: :grimacing:  JET shows 120 lines of runtime dispatches and optimization failures.

fonsp:
AMAZING 🌟🌟

Issue #1984, opened 2022-03-15
Feature request: choose page title


gdalle:
Hi there!
When I open a notebook in my browser, the tab name only says "⚡ Pluto.jl ⚡". Same thing with link sharing on platforms like [Discourse](https://discourse.julialang.org/t/getting-started-with-julia/77735/6). Would it be possible to choose the name of the web page, or at least add the file name automatically?

rikhuijzer:
You mean for notebooks hosted via PlutoSliderServer.jl? When running notebooks locally, I do see the title

fonsp:
https://github.com/fonsp/Pluto.jl/commit/4c31a6b51a4f40b77b729b44fec90ea86da7a839 is 10% of a fix: it removes the default title (`⚡ Pluto.jl ⚡`) from the HTML document, which means that social media previews like discourse will use the filename as the page title.

fonsp:
While working on the new plutojl.org website, I also found this problem! I am thinking about a standardized "frontmatter" system, allowing you to define the title, description, author, date, and more of the HTML export.

Right now, I (quickly) implemented by scanning for a global variable called `frontmatter`, like in this notebook: https://plutojl-preview.netlify.app/docs/widgets/javascript/ (link will probably break in the future). This is pretty cool, but maybe a GUI would be nicer... not sure if that's worth the time though.

gdalle:
Thanks for working on this! Not sure a GUI is worth it, putting `frontmatter` in a cell at the top (as long as this functionality is documented) seems very feasible for most Pluto users

fonsp:
#2008 was merged, but it was not yet used to solve this issue. Still to do: the `generate_html(notebook::Notebook)` function should call `frontend(notebook)` and use that to generate the header fields. They can be inserted with `replace(conents, "<head>" => "<head>$(stuff)")`

PR #1993, opened 2022-03-18
Fix #1987 and #1990

### LINKED ISSUE #1987: MathJax not rendered in exported static HTML
simpleMJX jl-staticHTMLexport

MathJax is not rendered in exported Static HTML.

This gist contains the Pluto notebook and its static html export.
https://gist.github.com/mdAshford/2a7debbe3b022b88146cdc7d150cfa8f

I'm running Pluto.jl v0.18.2

The offending code in the exported html appears to be in the script tag where MathJax is loaded. (line 56 in the gist). 
Changing `not-the-src-yet="https://cdn...` to `src="https://cdn...` solves the problem.

```html
<!-- This script will be enabled by JS after the notebook has initialized to prevent taking up bandwidth during the initial load. -->
<script type="text/javascript" id="MathJax-script" not-the-src-yet="https://cdn.jsdelivr.net/npm/mathjax@3.1.2/es5/tex-svg-full.js" async></script>
```

Thanks for a great platform!

### LINKED ISSUE #1990: Math display in HTML export suddenly won't work
Hey Pluto people!
I just realized that the HTML export no longer renders math properly: see https://gdalle.github.io/IntroJulia/efficiency.html for an example. Note that the math is displayed when the Pluto server runs, so it is really a matter of export.
Is it just me?

ctrekker:
These changes should fix the problems with lazy MathJax loading. Additionally this PR fixes a problem with the pluto-editor web component argument parser. See code comments for more info.

Closes #1987 
Closes #1990

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -761,17 +761,6 @@ patch: ${JSON.stringify(
         }
         this.on_disable_ui()
 
-        if (this.state.static_preview) {
-            this.setState({
-                initializing: false,
-                binder_phase: this.state.offer_binder ? BinderPhase.wait_for_user : null,
-            })
-            // view stats on https://stats.plutojl.org/
-            count_stat(`article-view`)
-        } else {
-            this.connect()
-        }
-
         setInterval(() => {
             if (!this.state.static_preview && document.visibilityState === "visible") {
                 // view stats on https://stats.plutojl.org/
@@ -1115,6 +1104,20 @@ patch: ${JSON.stringify(
                 // and don't prevent the unload
             }
         })
+    }
+
+    componentDidMount() {
+        if (this.state.static_preview) {
+            console.log("setting state to initializing: false")
+            this.setState({
+                initializing: false,
+                binder_phase: this.state.offer_binder ? BinderPhase.wait_for_user : null,
+            })
+            // view stats on https://stats.plutojl.org/
+            count_stat(`article-view`)
+        } else {
+            this.connect()
+        }
     }
 
     componentDidUpdate(old_props, old_state) {
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -1108,7 +1108,6 @@ patch: ${JSON.stringify(
 
     componentDidMount() {
         if (this.state.static_preview) {
-            console.log("setting state to initializing: false")
             this.setState({
                 initializing: false,
                 binder_phase: this.state.offer_binder ? BinderPhase.wait_for_user : null,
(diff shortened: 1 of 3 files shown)


frontend/components/Editor.js
```diff
@@ -1117,6 +1106,20 @@ patch: ${JSON.stringify(
         })
     }
 
+    componentDidMount() {
+        if (this.state.static_preview) {
+            console.log("setting state to initializing: false")
+            this.setState({
+                initializing: false,
+                binder_phase: this.state.offer_binder ? BinderPhase.wait_for_user : null,
+            })
```
ctrekker, line 1115: Setting the state in the constructor was never actually going through, so instead we wait for the component to mount, _then_ set initializing to false to trigger the static render logic.

frontend/editor.js
```diff
@@ -153,7 +153,7 @@ class PlutoEditorComponent extends HTMLElement {
     }
 
     connectedCallback() {
-        const new_launch_params = Object.fromEntries(Object.entries(launch_params).map(([k, v]) => [k, from_attribute(this, k) ?? v]))
+        const new_launch_params = Object.fromEntries(Object.entries(launch_params).map(([k, v]) => [k, from_attribute(this, k) || v]))
```
ctrekker, line 156: Classic example of JS weirdness. Consider the `truthy()` value (true or false from def of `truthy` fn) that `disable_ui` returns when `from_attribute` is called. The `??` operator will return the first non-nullish value, which is guarenteed to be whatever truthy returns in `from_attribute`, be it true or false (since true and false are both non-nullish). This means that if `disable_ui` is set in `launch_params` but not in attributes, the `launch_params` value of `disable_ui` will still be overridden. This meant in static exports the UI was still being shown.

Switching to the `||` operator will yield the first truthy value (or the rightmost value if there are no truthy values) instead of the first nullish value, which I believe is sufficient to maintain the same behavior for other parameters while also properly handling boolean parameters. Let me know if I'm missing a case here though.

fonsp:
Oops, we were working on this at exactly the same time!!

fonsp:
I like your PR better :)

frontend/editor.js
fonsp: Also found this while fixing this PR! Fixed in https://github.com/fonsp/Pluto.jl/commit/d2ea880ec38141f9064f09b446600dce2c4cbbbf

Pangoraw:
I think the same logic of `!this.state.initializing` should probably be used for this code block:

https://github.com/fonsp/Pluto.jl/blob/05a0e0533c89c486dd8b8304d94852cc0a781039/frontend/components/Editor.js#L1141-L1143

because the document title is never set in static exports (see #1984 and `document.title === ""`).

PR #1995, opened 2022-03-18
Allow JS state (`this`) inside DivElement

fonsp:
cc @ghaetinger


dralletje:
Smart!

fonsp:
Maybe we should solve script ID before merging this..

ghaetinger:
I've just tested this out using D3, which requires `this` to be persisted on updates. Here's how it works:

Using this PR:

https://user-images.githubusercontent.com/23220288/160865679-0ef9b119-3396-4b5d-b52b-2011718894cc.mp4

Without this PR:

https://user-images.githubusercontent.com/23220288/160866035-da0bdab7-dadd-4966-94cb-cab3793b22af.mp4

The animation looks really cool and it's completely removed without this 😢

fonsp:
Thanks for testing!!

PR #2008, opened 2022-03-28
💌 `Pluto.frontmatter`

fonsp:
```julia
frontmatter(nb::Notebook; raise::Bool=false)::Dict{String,Any}
```

Extract frontmatter from a notebook, which is extra meta-information that the author attaches to the notebook, often including *title*, *description*, *tags*, *author*, and more. Search for *frontmatter* online to learn more.

Currently, you can give frontmatter to a notebook by defining a global variable `frontmatter` anywhere in the notebook, which should be a named tuple. This might change in the future, because we might make this a GUI thing!

Related to #1984


src/notebook/frontmatter.jl
```diff
@@ -0,0 +1,64 @@
+
+const FrontMatter = Dict{String,Any}
+
+
+"""
+```julia
+frontmatter(nb::Notebook; raise::Bool=false)::Dict{String,Any}
+```
```
rikhuijzer, line 8:
```suggestion
    frontmatter(nb::Notebook; raise::Bool=false)::Dict{String,Any}
```

fonsp:
Yeah I think that that's nice for dates, not sure what else... Any cool ideas where this is useful?

rikhuijzer:
> Yeah I think that that's nice for dates, not sure what else... Any cool ideas where this is useful?

Images maybe? That it handles storing images and getting permalinks for the open graph for Twitter thumbnails and stuff

PR #2013, opened 2022-03-29
Disable local_autocomplete 😢

fonsp:
Reverts #1925 because of performance issues

I added [completeAnyWord](https://codemirror.net/6/docs/ref/#autocomplete.completeAnyWord) so that we still have local autocomplete, although less accurate

This also means that using globals inside a cell will cause duplicate completions

but this will be fixed by https://github.com/codemirror/autocomplete/pull/10 EDIT merged and used in pluto by #2021 

cc @Pangoraw


fonsp:
Enabled again in https://github.com/fonsp/Pluto.jl/pull/3151

Issue #1695, opened 2021-11-27
Support one bond defining another when setting both simultaneously


fonsp:
See https://github.com/JuliaPluto/PlutoSliderServer.jl/issues/38

lungben:
Copied from https://github.com/JuliaPluto/PlutoSliderServer.jl/issues/38 :

This currently works in Pluto, but not in PlutoSliderServer:

```
@bind x Slider(1:10)

@bind y Slider(1:x)

x, y
```
Why? When moving the y slider in normal Pluto, we just send the y value to the Pluto server, it sets it in the notebook process and runs the dependent cells (just cell 3). In PlutoSliderServer, moving y will send x and y together, this is our stateless idea.

Pluto does not handle this correctly (this is a bug in Pluto, not PSS): it first sets both x and y in the notebook process, and then runs the cells that depend on them: cells 2 and 3. The problem is: running cell 2 will set y to its initial value: 1, and you will always see x, 1 in the third cell.

What should be done?

If the SliderServer submits updated values for `(x, y)` then:

1. Update the Slider in `@bind y Slider(1:x)`
2. Do not set `y` as the value from the Slider, but keep `y` as the value in the request from SliderServer.

As far as I understand, a situation where a variable value (here `y`) is requested both from the SliderServer and result of a dependent cell in the notebook itself can only happen for `@bind` values (otherwise it would be an invalid multiple definition of a variable).
Thus, a fix could be to reset a variable back to its request value if the variable is set in a `@bind`.

@fonsp could you please give me a quick pointer which part of the Pluto code receives the update requests of the SliderServer?

lungben:
As far as I can see from the code, the updates arrive in Pluto here: https://github.com/fonsp/Pluto.jl/blob/main/src/evaluation/RunBonds.jl#L1

The information which variables are in the request is in `bound_sym_names`.

This information needs to be populated downstream to `run_reactive!()`, i.e. https://github.com/fonsp/Pluto.jl/blob/main/src/evaluation/Run.jl#L129, where the value in the request from PlutoSliderServer should overwrite the new (default) value obtained from recalculation of the cell (if they are different).

Is this the right approach?

lungben:
If the first bound value (here `max_x`) is changed so that the choice of the 2nd bound value (here `x`) is no longer valid, an ArgumentError is currently raised. This can be fixed by updating the 2nd bound value in the GUI.

[...]

PR #2014, opened 2022-03-30
Support dependent bound variables for SliderServer

lungben:
https://github.com/fonsp/Pluto.jl/issues/1695

### DIFF
--- a/src/evaluation/Run.jl
+++ b/src/evaluation/Run.jl
@@ -2,6 +2,7 @@ import REPL:ends_with_semicolon
 import .Configuration
 import .ExpressionExplorer: FunctionNameSignaturePair, is_joined_funcname, UsingsImports, external_package_names
 import .WorkspaceManager: macroexpand_in_workspace
+import .MoreAnalysis: find_bound_variables
 
 Base.push!(x::Set{Cell}) = x
 
@@ -14,7 +15,8 @@ function run_reactive!(
 	deletion_hook::Function = WorkspaceManager.move_vars, 
 	user_requested_run::Bool = true, 
 	already_in_run::Bool = false, 
-	already_run::Vector{Cell} = Cell[]
+	already_run::Vector{Cell} = Cell[],
+	externally_updated_variables::Dict{Symbol, Any} = Dict{Symbol, Any}()
 )::TopologicalOrder
     if !already_in_run
         # make sure that we're the only `run_reactive!` being executed - like a semaphor
@@ -132,6 +134,18 @@ function run_reactive!(
                 user_requested_run = (user_requested_run && cell ∈ roots)
             )
             any_interrupted |= run.interrupted
+
+			# Support one bond defining another when setting both simultaneously in PlutoSliderServer
+			# https://github.com/fonsp/Pluto.jl/issues/1695
+			bound_variables_defined_by_this_cell = find_bound_variables(notebook.topology.codes[cell].parsedcode)
+			redefined_external_bound_variables = bound_variables_defined_by_this_cell ∩ keys(externally_updated_variables)
+			for bond_var ∈ redefined_external_bound_variables
+				# set the redefined bound variables to their original value from the request
+				bond_val = externally_updated_variables[bond_var]
+				expr = :($bond_var = $bond_val)
+				WorkspaceManager.eval_in_workspace((session, notebook), expr)
+			end
+
         end
 
         cell.running = false
--- a/src/evaluation/RunBonds.jl
+++ b/src/evaluation/RunBonds.jl
@@ -36,6 +36,7 @@ function set_bond_values_reactive(;
     end
 
     new_values = [notebook.bonds[bound_sym].value for bound_sym in to_set]
+    externally_updated_variables = Dict{Symbol, Any}(zip(bound_sym_names, new_values))
     
     function custom_deletion_hook((session, notebook)::Tuple{ServerSession,Notebook}, old_workspace_name, new_workspace_name, to_delete_vars::Set{Symbol}, methods_to_delete::Set{Tuple{UUID,FunctionName}}, to_reimport::Set{Expr}; to_run::AbstractVector{Cell})
         to_delete_vars = Set([to_delete_vars..., to_set...]) # also delete the bound symbols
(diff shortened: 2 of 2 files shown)


lungben:
Thanks for pointing out the value transformations!

It is a general issue for dependent sliders that `y` may contain an invalid value after changing `x`, either by being not allowed anymore (out of range for sliders, invalid selections for `Select`, etc.) or by the value transform issue you described.
However, after the user updates the element bound to `y` it should have a valid value again. Thus, maybe this is an acceptable behavior?

As far as I know, PlutoSliderServer does not distinguish in its request which variable has been explicitly changed by the user and which is only co-dependent. Thus, we cannot check if the latter still has a valid value with the updated UI element (and reset it to the default if not). Or do I miss something?

Pangoraw:
For the valid value problem, there is plan to integrate [AbstractPlutoDingetjes.validate_value](https://docs.juliahub.com/AbstractPlutoDingetjes/UHbnu/1.1.4/#AbstractPlutoDingetjes.Bonds.validate_value-Tuple{Any,%20Any}) inside PlutoSliderServer and prevent invalid value altogether (from calling the API directly).

You are right that the request always contains both bonds values, so actually the transformation should be done with respect to the new widget value (not previous as i said). Therefore, you should be okay by just transforming after the cell was run:

```diff
		        for bond_var ∈ redefined_external_bound_variables
				# set the redefined bound variables to their original value from the request
				bond_val = externally_updated_variables[bond_var]
-				expr = :($bond_var = $bond_val)
+				expr = :($bond_var = Main.PlutoRunner.transform_value($(QuoteNode(bond_var)), $bond_val))
				WorkspaceManager.eval_in_workspace((session, notebook), expr)
			end
```

lungben:
If the first bound value (here `max_x`) is changed so that the choice of the 2nd bound value (here `x`) is no longer valid, an ArgumentError is currently raised. This can be fixed by updating the 2nd bound value in the GUI.

image

Maybe we should catch this error and show a more descriptive error to the user?

lungben:
Wrapping `Main.PlutoRunner.transform_bond_value()` in a try-catch block did not work for me, neither locally in Run.jl nor if it is part of the `expr` to be executed in the notebook process.

fonsp:
Awesome! Very cool that this could be done with such a small PR!

We should add tests for this into Pluto.jl itself, right now there are only on PSS. It would be nice if the tests do not depend on PlutoUI, so you can do:


```julia
begin
    struct TransformSlider
        range::AbstractRange
    end
    
    Base.show(io::IO, m::MIME"text/html", os::TransformSlider) = write(io, "<input type=range value=$(minimum(os.range)) min=$(minimum(os.range)) max=$(maximum(os.range))>")
    
    Bonds.initial_value(os::TransformSlider) = Bonds.transform_value(os, minimum(os.range))
    Bonds.possible_values(os::TransformSlider) = os.range
    Bonds.transform_value(os::TransformSlider, from_js) = from_js * 2
end
```

which you can use like so:

```julia
@bind x TransformSlider(1:10)

@bind y TransformSlider(1:x)
```

lungben:
From my side, this PR is now ready for merging.

- The failure of the tests on MacOS is not related to my changes, as far as I can see (3 tests from the set "Reload from file" are failing, and only on MacOS)
- Unit Tests for the change in this PR have been added
- The behavior of dependent bound variables in Pluto itself is the same as without this PR as far as I can see
- With PlutoSliderServer, dependent bound variables now work correctly

fonsp:
Thanks! Can you add a test with a transformed value as in https://github.com/fonsp/Pluto.jl/pull/2014#issuecomment-1090220317 ? Since this PR had some discussion and changes related to it, it makes sense to also add a test, right?

It should be easy to add, by adding three new cells to the test notebook, (defining `a` and `b` instead of `x` and `y`).

lungben:
Done

Interestingly, the TransformSlider requires setting of `🍭.options.evaluation.workspace_use_distributed = true` in the tests, otherwise the bound value is not transformed. But I think this is unrelated to my PR because the same setting is also in the testset "AbstractPlutoDingetjes.jl".

fonsp:
Weird! But that's a good find! I made https://github.com/fonsp/Pluto.jl/issues/2056 to track that separately from this PR

fonsp:
Hey! I noticed that we might be able to clean this up a bit: we now have the old mechanism for setting bond values: `custom_deletion_hook`, and the new mechanism, `externally_updated_variables`. 

In theory, `externally_updated_variables` should be a generalization of the trick in `custom_deletion_hook`, so we should be able to just delete that hook in `set_bond_values_reactive`. I didn't think this through 100% yet, but @lungben or @Pangoraw could you look into this? This new mechanism should also be very similar to the way #1052 injects values. If it's easy to implement, let's do it in this PR.

lungben:
As far as I can see, both `custom_deletion_hook` and the new functionality loop over symbol/value pairs and eval them in the workspace with `WorkspaceManager.eval_in_workspace((session, notebook), :($(bound_sym) = Main.PlutoRunner.transform_bond_value($(QuoteNode(bound_sym)), $(new_value))))`. 
We could put this functionality into a utility function.

`custom_deletion_hook`  in addition has `WorkspaceManager.move_vars()`, which is not used in the new functionality of this PR.

lungben:
@fonsp can this PR be merged?

fonsp (approved):
Thanks @lungben , and apologies for the delay! 

I initially thought that this PR would be difficult to review, because I was worried about how it interplays with funky combinations of bonds. But in fact, this PR does not change behaviour when only a single bond is being set, because there is only one cell defining that bond, and that one is excluded from the reactive run (because [we use the cells that reference a set bond as the root of the reactive run](https://github.com/fonsp/Pluto.jl/blob/76e4dd60aa15f2259eaa537981745741edbdb4c8/src/evaluation/RunBonds.jl#L49)).

There is one exception: the following notebook behaves differently:

```julia
begin
    if @isdefined(x)
        # cool
    end
    
    bond = @bind x html"<input type=range>"
    
    x = 999
    
    bond
end
```
```julia
x
```

because the first cell both *defines* **and** *references* `x`, so [it is a cell that references the bound symbol](https://github.com/fonsp/Pluto.jl/blob/76e4dd60aa15f2259eaa537981745741edbdb4c8/src/evaluation/RunBonds.jl#L49), and it will be part of the reactive run. This mechanism can be used for single-cell interactions like [this one](https://computationalthinking.mit.edu/Spring21/inverse_climate_model/).

As it happens, the example above gets stuck in an infinite loop [on the current Pluto](https://user-images.githubusercontent.com/6933510/169170647-59a3e628-9bab-4614-8cfa-08638eb09f91.mov) (long story), and [after this PR](https://user-images.githubusercontent.com/6933510/169170631-7a96a709-0537-4065-bfa7-f68feac4ef3e.mov), it doesn't get stuck, and the `x = 999` line is "ignored".

But! The [single-cell interaction example](https://computationalthinking.mit.edu/Spring21/inverse_climate_model/) that I mentioned above **still works**! So I take that as a sign that the behaviour is generally as expected, and can even contribute to preventing loops.

TLDR: 👍👍

PR #2018, opened 2022-03-31
Implement @skip_as_script functionality inside cell metadata

disberd:
First attempt to implement the functionality of the `@skip_as_script` macro directly within the cell metadata in the notebook file.

Being able to comment out cells in the file without the need of te `@skip_as_script` macro is quite useful especially when developing packages with Pluto notebooks as basis.

I currently have one problem with the context menu button to toggle the skip_as_script functionality.
I tried copying the functionality from the disabled cell toggle from @lungben, but it seems the change in the metadata for the cell is only correctly registered by the front end after re-running the cell.
This is not ideal as this flag would not need re-running cell for its functionality, but I left the re-run in the added code till I get some help or understand how to toggle this without rerunning.
Maybe some feedback on this from @fonsp or @pankgeorg?


pankgeorg:
The implementation looks pretty neat, the metadata system seems quite powerful! Any idea why the tests are failing?

disberd:
I think there was one test where a cell was specifically created programmatically without the new `skip_as_script` entry, but in the code to save the notebook that metadata field is explicitly accessed.
If that was the only problem, I hope my second commit fixes that

lungben:
> I tried copying the functionality from the disabled cell toggle from @lungben, but it seems the change in the metadata for the cell is only correctly registered by the front end after re-running the cell.

For cell disabling / enabling, the cell is currently always re-executed. If the cell is disabled, this does nothing, except disabling all dependent cells. If a cell is enabled, running the cell brings the notebook back into a valid state.

fonsp:
Do you need help resolving the conflicts?

disberd:
@fonsp I think this is now ready for review.

I did not change the visual aspect of it from the beginning as I have no nice idea on how to improve it nor much bandwidth do it in the next days.
If you prefer to avoid having the button in the context menu and you are fine with the rest of the changes, I can maybe just remove the button on the context menu and just leave this as a backend implementation that can be only accessed via the internal `pluto_actions` so that it's accessible to who might need it and polish the visuals in a separate PR?

greimel:
I've just tried this and this is what happened.

https://user-images.githubusercontent.com/6280307/173086486-f5d209c8-2991-4667-8d36-5a0f599f38d4.mov

disberd:
@greimel Thanks for trying this out.

I'll try out to see what is tue problem causing issue 1. (Though the idea with @fonsp was to push the basic functionality and probably avoid having a context menu item just for this, moving the front-end api for skippi cells part of a subsequent PR)

Regarding the additional functionality it does seem interesting but I think it would need some more discussion to check whether we are not risk hitting some corner cases where you wouldn't want to skip cells automatically.
In any case, I think it's best to try and push the basic functionality as soon as possible here and then add that dependent cells skip functionality as a separate PR.

frontend/components/CellInput.js
```diff
@@ -877,6 +884,15 @@ const InputContextMenu = ({ on_delete, cell_id, run_cell, running_disabled, any_
                       ${running_disabled ? html`<span class="enable_cell ctx_icon" />` : html`<span class="disable_cell ctx_icon" />`}
                       ${running_disabled ? html`<b>Enable cell</b>` : html`Disable cell`}
                   </li>
+                  <li
+                      onClick=${toggle_skip_as_script}
+                      title=${skip_as_script
+                          ? "Skip this cell when running the notebook as script, making it exclusive to the notebook"
+                          : "Run this cell also when calling the notebook as a script"}
```
Pangoraw, line 891: I think the descriptions do to match the label and should be switched:

```suggestion
                      title=${skip_as_script
                          : "Run this cell also when calling the notebook as a script"
                          ? "Skip this cell when running the notebook as script, making it exclusive to the notebook"}
                       
```

Pangoraw:
The tests are sometimes failing because they are timing sensitive, nothing related with the PR! 🙌 

It look great! What has been decided about styling/hiding the button for now ([mentioned here](https://juliapluto.github.io/weekly-call-notes/2022/06-06/notes.html)) ?

disberd:
I think @fonsp still didn't have the chance to think what to do about it. You have seen the piece of parchment TODO list during yesterday's call :D

fonsp:
Awesome! My main changes were:

[renaming to "Disable in file" and changing the wording of the hover title help](https://github.com/fonsp/Pluto.jl/pull/2018/commits/d46136fdd3a13c8d9f5368d200fe8bc538d9bc40)

[adding a popup when you click the mysterious new line next to a cell to explain what's up](https://github.com/fonsp/Pluto.jl/pull/2018/commits/11ff7745a96e7469191e153e31e36bb92efac57a)

Schermafbeelding 2022-06-16 om 00 33 20

With these changes, I feel like the feature (while still niche) is not too intimidating anymore! So let's just release it publicly for now, instead of adding a secret switch somewhere to enable it.

disberd:
Thanks a lot @fonsp! Definitely look nicer now!

Issue #1013, opened 2021-03-21
Disable "Open" button on main menu when nothing is typed


digital-carver:
Is the Open button below "Open from file:" in the Pluto welcome page supposed to open the OS File manager window to allow us to choose? 

* If so, it doesn't do anything on my system - Firefox 86 on Linux KDE. Clicking the button does nothing, no errors on the console either. 
* If it's not intended to open a browsable file manager window, then the button should probably be shown disabled until there is some text in the "Enter path or URL" textbox. Though, actually opening a file manager window would be much more convenient, so adding that functionality would be the preferred solution in case it isn't already there. 

(Also, this part should probably be a separate issue, but perhaps the "notebook = " kwarg should be mentioned in the `Pluto.run` help doc, as that feels like an option that would also count as common (I found it thanks to coming across [this issue](https://github.com/fonsp/Pluto.jl/issues/936) that mentions it). And despite the text saying "For the full list, see the Pluto.Configuration module.", there isn't an easily accessible list of options in the module, since it's split in multiple places and not very obvious. Perhaps a docstring in that module listing all the options would be helpful. Let me know if this is something worth opening a separate issue for.)

fonsp:
Can you remove the second issue and consider posting it separately?

fonsp:
Thanks for the report, the button is not supposed to work when nothing is typed. Also see #833.

HirumalPriyashan:
@fonsp I think we can close this issue since https://github.com/fonsp/Pluto.jl/pull/2020 is merged.

fonsp:
Thanks!

PR #2020, opened 2022-04-01
Disable "Open" button on main menu when nothing is typed

HirumalPriyashan:
## Resolves
- https://github.com/fonsp/Pluto.jl/issues/1013

## Approach taken

- Set disabled property for `Open` button at start
- Add an id to `Open` button
- Set an event handler for input event to editor state and enable-disable the button


fonsp:
Awesome! Can you use react to set the disabled state (`useState`) instead of modifying DOM directly? Take a look at https://reactjs.org/docs/hooks-intro.html

I think that you also don't need the `setTimeout`, but if you do, be sure to add a comment in code explaining why

HirumalPriyashan:
Thanks for the feedback, made the suggested changes with `51cc7c3`.

> I think that you also don't need the setTimeout, but if you do, be sure to add a comment in code explaining why

Added `setTimeout` here as a workaround for unexpected behavior of `setState`, added a comment in code as well.

HirumalPriyashan:
> What happens if you remove setTimeout? You probably don't need it anymore now that you rewrote it to use react state

There is a delay in update of state, sometimes it updates the state at the second or third input change

fonsp:
Can you make a video?

HirumalPriyashan:
sure,

without setTimeout

https://user-images.githubusercontent.com/57571152/161557807-127cb297-76b4-4c07-9b76-9cbc7d1ba510.mp4

with setTimeout

https://user-images.githubusercontent.com/57571152/161557936-1e9dbc00-a72d-476f-be68-6e64350b1c7d.mp4

HirumalPriyashan:
Thanks, I will look into that and update the PR.

HirumalPriyashan:
@fonsp updated the PR with a _codemirror compartment_

fonsp:
Thanks! Can you run the VS Code formatter? This should happen automatically if you save the file in VS Code: https://github.com/fonsp/Pluto.jl/tree/main/.vscode

HirumalPriyashan:
@fonsp I am using VS Code and formatting happens automatically when I save files, But  it doesn't seem to do any formatting to the current code.

fonsp:
Be sure to install the "workspace recommended extensions", type `@recommended ` in the extensions search bar, and then click install:

Schermafbeelding 2022-04-20 om 14 16 11

HirumalPriyashan:
Yes, I got them installed and been using them for a while now.
Still I have no idea why they didn't work here.

image

Issue #2019, opened 2022-03-31
Pluto hangs when using MATLAB.jl 


fcdimitr:
# Pluto hangs when using MATLAB.jl 

Pluto notebook that calls MATLAB commands runs indefinitely.
Perhaps it is an issue with parsing the terminal output?
 
Minimum example: https://gist.github.com/fcdimitr/c48cc6f6c4764a61b555f0c985bf76c0

I noticed the issue with version v0.18.4, it was working with v0.18.0

disberd:
I just tried running your notebook on latest Pluto and I don't notice this:

![image](https://user-images.githubusercontent.com/12846528/161040436-8b973734-c34f-452c-a8be-4d8807b4683e.png)

fcdimitr:
Thank you for testing it on Windows!

The issue appeared on my side on both Linux (Ubuntu) and macOS 11 (Monterey).

I do not have access to a Windows machine to try it.

fcdimitr:
I am retesting with a simpler version; only trying to start a `MATLAB` session with a 0-length buffer to store MATLAB output text:

```julia
using MATLAB
MATLAB.default_msession_ref[] = MSession(0; flags="-nodisplay -nosplash -nodesktop");
```

This command alone freezes `Pluto` indefinitely on `macOS` with `Julia v1.7.2`

![Screen Shot 2022-04-02 at 18 35 50](https://user-images.githubusercontent.com/4839092/161390378-9560bd02-eacc-47c7-9ed7-f4b2735f9b9e.png)

The same command works as expected through `REPL`.

fcdimitr:
Thanks to @lungben for the following workaround:

> You could try to deactivate multiprocessing in Pluto:
> https://github.com/fonsp/Pluto.jl/blob/15a94f415dba4fbad753ddac5153fa9286ce0eb2/src/Configuration.jl#L106
> `Pluto.run(workspace_use_distributed=false)`

fonsp:
Might be linked to #2001

fonsp:
I don't have a matlab license so it won't be easy for me to work on this. @fcdimitr be sure to subscribe to #2001, if we make a candidate PR to fix that issue, try it out on the matlab notebook.

fcdimitr:
Disabling output forwarding, via #2022, resolved the issue on my side.

Issue #2001, opened 2022-03-24
Can't use PlutoSliderServer *inside* a Pluto notebook


fonsp:
Using PlutoSliderServer inside Pluto get stuck at this point:

<img width="942" alt="image" src="https://user-images.githubusercontent.com/6933510/160024140-37729dd8-5a13-47f3-a38a-fb3835ab6498.png">

The same in the REPL:

https://user-images.githubusercontent.com/6933510/160024786-82b123b5-7acd-4d2f-8a78-2b9abcf5646d.mov

MWE:

https://gist.github.com/fonsp/8863107980a4210cf939330b7854c030

# Cause

It looks like this is caused by #1957. If you disable this feature (by changing `capture_stdout` to `false` in [this line](https://github.com/fonsp/Pluto.jl/blob/0c5ce9deadbccb35c591bb2166a993dec3a98f2a/src/runner/PlutoRunner.jl#L533)), everything works as expected.

fonsp:
@lungben Your comment in #2019 suggests that this might be a unix-only issue. Can you run the notebook on windows to confirm?

disberd:
This is what I get with the MWE notebook on windows:
![image](https://user-images.githubusercontent.com/12846528/161538335-de2022fe-9548-4c5f-b118-bf5345fd8b31.png)

fonsp:
😅 of course, thanks

fonsp:
@fcdimitr Good idea! https://github.com/fonsp/Pluto.jl/pull/2022#issuecomment-1087659014

fcdimitr:
Pull request #2022 did not resolve this issue.

My guess is that `PlutoSliderServer` internally starts `Pluto 0.18.4` without disabling stdout forwarding. I still see the output under the cell when starting `PlutoSliderServer`.

If the pull request is merged and the option is exposed from `PlutoSliderServer` then I believe we could bypass this error by disabling terminal output when calling `PlutSliderServer`.

fonsp:
This is what the notebook looks like when running with `Pluto.run(capture_stdout=false)` (#2022): 

<img width="783" alt="image" src="https://user-images.githubusercontent.com/6933510/161979249-df01342e-f97a-4a4c-99a3-7517dbbc5c73.png">

The last cell is stuck. Interrupting gives this error message:

```
TaskFailedException

nested task error: InterruptException:

Stacktrace:

[...]

PR #2022, opened 2022-04-04
Option `capture_stdout` to enable/disable stdout capture

fonsp:


fcdimitr:
Disabling output capturing resolves issue #2019 with `MATLAB`. Thanks @fonsp! 

However, issue #2001 is not resolved. My guess is that `PlutoSliderServer` internally starts `Pluto` 0.18.4 **without** disabling stdout forwarding. I still see the output under the cell when starting `PlutoSliderServer`. 

I believe this feature should be merged into the next release. Then `PlutoSliderServer` may expose `capture_stdout` as well, in order to bypass this error.

fcdimitr:
Disabling output capturing resolves issue #2019 with `MATLAB`. Thanks @fonsp! 

However, issue #2001 is not resolved. My guess is that `PlutoSliderServer` internally starts `Pluto` 0.18.4 **without** disabling stdout forwarding. I still see the output under the cell when starting `PlutoSliderServer`. 

I believe this feature should be merged into the next release. Then `PlutoSliderServer` may expose `capture_stdout` as well, in order to bypass this error.

fonsp:
Thanks for the info @fcdimitr ! Very strange that #2001 still happens...

True that releasing this would help, but the main problem of course is that there is an underlying blocking issue with our system which we should fix. 👀

PR #2016, opened 2022-03-30
Notebook metadata

ctrekker:
This PR adds a field `metadata` to Pluto's `notebook` structure. Nothing too special, but a few details are worth mentioning:

- Metadata is saved directly below file version number and above first cell
- Updates to metadata from client is synced over WS properly

### TODO
- [x] Save / load mechanism
- [x] Client synchronization
- [x] Tests
- [x] ~Read-only testing to pin format~ not necessary

### Prefix Ideas
This is arguably the least important part of this PR, but currently the metadata is saved in the following format:

```
### A Pluto.jl notebook ###
# v0.18.4

#> [frontmatter]
#> title = "Fancy title!"
#> description = "......"
#> other_metadata = 42

...rest of notebook here...
```

This format works just fine, but Pluto's cell prefix format sets a precedent for wild and unique prefixes. If anyone has ideas that would make it implicitly clear that the lines marked with this prefix are _notebook-level_ metadata statements, let me know.

A box around properties could be neat (but maybe overkill or distracting and maybe make editing properties from the file mildly annoying)

```
### A Pluto.jl notebook ###
# v0.18.4
# ┌────────────────────────┐
# │ title = "Fancy title!" │
# │ description = "......" │
# │ other_metadata = 42    │
# └────────────────────────┘

...rest of notebook here...
```

PR #2023, opened 2022-04-04
Option to hide logs per cell persistently

Pangoraw:
This also removes a global notebook log toggle that was already there but not used?
We can put it back once #2016 is merged.


frontend/components/Notebook.js
```diff
@@ -66,6 +64,7 @@ let CellMemo = ({
         // Object references may invalidate this faster than the optimal. To avoid this, spread out objects to primitives!
         cell_id,
         metadata.disabled,
+        metadata.show_logs,
```
Pangoraw, line 67: @disberd not sure if you already know, but i think this where the memo is and the `skip_as_script` property should be added to re-render a cell.

disberd: Thanks!
Indeed yesterday @pankgeorg showed me this part :)

frontend/components/CellInput.js
```diff
@@ -833,13 +833,11 @@ const InputContextMenu = ({ on_delete, cell_id, run_cell, running_disabled, any_
                       ${running_disabled ? html`<span class="enable_cell ctx_icon" />` : html`<span class="disable_cell ctx_icon" />`}
                       ${running_disabled ? html`<b>Enable cell</b>` : html`Disable cell`}
                   </li>
-                  ${any_logs
```
fonsp, line 836: I'd like to keep the `any_logs` check so that we only show the option to hide/show if there are logs. Because if you have never seen logs, it's hard to understand what the button does.

fonsp:
Awesome!!

frontend/components/CellInput.js
Pangoraw: Makes sense! that makes the change mostly frontend since logs have to still be appended!

rikhuijzer:
This is great. Thanks @Pangoraw

PR #2026, opened 2022-04-06
Add yield()s around logger redirect

### LINKED ISSUE #2001: Can't use PlutoSliderServer *inside* a Pluto notebook

fonsp:
While debugging this issue I added random println statements everywhere, and suddenly the issue was gone 😯 So I guessed that `println` triggers a `yield`, and this turned out to work!

Both `yield`s are needed: the first one fixes the stalling issue, the second one makes sure that all logs are redirected. In the #2001 example, you only see this third log with both yields:

image

Fix #2001

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -664,7 +664,7 @@ function move_vars(old_workspace_name::Symbol, new_workspace_name::Symbol, vars_
                 catch ex
                     if !(ex isa UndefVarError)
                         @warn "Failed to move variable $(symbol) to new workspace:"
-                        showerror(stderr, ex, stacktrace(catch_backtrace()))
+                        showerror(original_stderr, ex, stacktrace(catch_backtrace()))
                     end
                 end
             end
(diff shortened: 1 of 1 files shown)


fonsp:
@fcdimitr Can you check if this fixes #2019? You shouldn't need the option from #2022, just use the default configuration

fonsp:
Actually, this is still not redirecting all the logs... it's missing these two:

Schermafbeelding 2022-04-06 om 15 50 40

fcdimitr:
@fonsp I just tested on the main branch (ac6d5a10860be0d0d01afb7114fc36d01290ff73), after you merged; issue #2019 remains unresolved. Pluto gets stuck at the first `MATLAB` command.

PR #2027, opened 2022-04-06
Scope log_channel to a single notebook_id

fonsp:
This should improve situations where you are running Pluto inside Pluto.


dralletje:
Yes!!!

PR #2028, opened 2022-04-06
Improve readability and inferability inside `Pluto.run(session, router)`

rikhuijzer:
This PR improves readability and inferability of `Pluto.run(session, router)`. Inside that method, the type of `port` wasn't inferred correctly according to `Cthulhu.@descend/@code_warntype`. With the change of this PR, the number of non concrete variables (red variables right at the top) inside `@code_warntype` is reduced from 8 to 4 on Julia 1.8-beta3. I think that the reason that it wasn't inferred is that it's a huge method and the compiler will at some point just give up when things are too difficult.

This improved inferability potentially allowed a larger part of the method to be compiled during

```julia
precompile(run, (ServerSession, HTTP.Handlers.Router{Symbol("##001")}))
```

which seems to be the case because `Pluto.run` went from 623MiB to 558MiB when running `test/compiletimes.jl` in Julia 1.8-beta3. For more info about finding these types of issues and why they affect precompilation, see https://huijzer.xyz/posts/inference/.

### DIFF
--- a/src/webserver/WebServer.jl
+++ b/src/webserver/WebServer.jl
@@ -119,7 +119,7 @@ end
 
 const is_first_run = Ref(true)
 
-"Return a port and serversocket to use while taking into acoount the `favourite_port`."
+"Return a port and serversocket to use while taking into account the `favourite_port`."
 function port_serversocket(hostIP::Sockets.IPAddr, favourite_port)
     local port, serversocket
     if favourite_port === nothing
--- a/src/webserver/WebServer.jl
+++ b/src/webserver/WebServer.jl
@@ -119,6 +119,22 @@ end
 
 const is_first_run = Ref(true)
 
+"Return a port and serversocket to use while taking into acoount the `favourite_port`."
+function port_serversocket(hostIP::Sockets.IPAddr, favourite_port)
+    local port, serversocket
+    if favourite_port === nothing
+        port, serversocket = Sockets.listenany(hostIP, UInt16(1234))
+    else
+        port = UInt16(favourite_port)
+        try
+            serversocket = Sockets.listen(hostIP, port)
+        catch e
+            error("Port with number $port is already in use. Use Pluto.run() to automatically select an available port.")
+        end
+    end
+    return port, serversocket
+end
+
 function run(session::ServerSession, pluto_router)
     if is_first_run[]
         is_first_run[] = false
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -4,6 +4,10 @@ include("helpers.jl")
 
 @timeit_include("compiletimes.jl")
 verify_no_running_processes()
+if get(ENV, "PLUTO_TEST_ONLY_COMPILETIMES", nothing) == "true"
+    show(TOUT; compact=true, sortby=:firstexec)
+    exit(0)
+end
```
rikhuijzer, line 10: I've added this because it's very useful for local development. Just set the ENV var and run `pkg> test` to get the compilation times.

Pangoraw: Nice! Related: I use a script that only loads the relevant Pluto test files based on a given regex ([here](https://gist.github.com/Pangoraw/e1fb2db3c2d756f8c12ea70af2e3892e#file-testpluto-jl)). So that i can only `tpluto Macro|Notebook` This does not work with compiletimes since the print is at the bottom of runtests.jl tough.

rikhuijzer: Great. I like it. Maybe I'll use your script in the future

rikhuijzer: @Pangoraw Does it also print the total times in the end? I guess so right? Maybe it's actually a great idea to print it once after compiletimes and another once all the tests are done. That way people could also do

```julia
$ julia --project

julia> using TestEnv

julia> TestEnv.activate()

julia> include("test/helpers.jl");

julia> include("test/compiletimes.jl")

```

src/webserver/WebServer.jl
```diff
@@ -119,6 +119,22 @@ end
 
 const is_first_run = Ref(true)
 
+"Return a port and serversocket to use while taking into acoount the `favourite_port`."
```
Pangoraw, line 122:
```suggestion
"Return a port and serversocket to use while taking into account the `favourite_port`."
```

Pangoraw (approved):
Interesting that moving into smaller functions makes such changes to compilation

test/runtests.jl
Pangoraw: the total time is 37s on my screenshot, not sure to which total it is refering tough because only this test file is run.

PR #2029, opened 2022-04-06
Generalize binder to other backends

fonsp:
Some JS changes to allow us to add another dynamic backend :o (it will be localhost:1234 for the secret new sample system)


fonsp:
@pankgeorg One could add a JH button?

pankgeorg:
> @pankgeorg One could add a JH button?

yes, totally!

PR #2039, opened 2022-04-19
Add a precompile statement to `PlutoRunner` and some minor refactorings

rikhuijzer:
Some small changes to make the PlutoRunner slightly more snappy.

Changes:

- Drop `Base.@kwdef` to make the code simpler and avoid having to compile a kwarg method.
- Fix some type inferability problems by adding type assertions. This is tested against Julia 1.8-beta3.
- Narrow down the type of `old_logger` a bit. This doesn't help in performance, but probably helps in readability.
- Add a `precompile` statement for `run_expression`.

## Time to first X

On Linux with Julia 1.8-beta3.

### Before (`master`)

```julia
$ julia --project -ie 'using Pluto'
[...]

julia> module Foo end
Main.Foo

julia> @time @eval Pluto.PlutoRunner.run_expression(Foo, :(1 + 1), Pluto.uuid1(), nothing);
[...]
  1.653029 seconds (2.94 M allocations: 152.834 MiB, 13.22% gc time, 99.80% compilation time)
```

### After (this PR)

```julia
$ julia --project -ie 'using Pluto'
[...]

julia> module Foo end
Main.Foo

julia> @time @eval Pluto.PlutoRunner.run_expression(Foo, :(1 + 1), Pluto.uuid1(), nothing);
[...]
  1.287273 seconds (1.60 M allocations: 82.489 MiB, 2.78% gc time, 99.70% compilation time)
```

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -28,12 +28,12 @@ MimedOutput = Tuple{Union{String,Vector{UInt8},Dict{Symbol,Any}},MIME}
 const ObjectID = typeof(objectid("hello computer"))
 const ObjectDimPair = Tuple{ObjectID,Int64}
 
-Base.@kwdef struct CachedMacroExpansion
+struct CachedMacroExpansion
     original_expr_hash::UInt64
     expanded_expr::Expr
     expansion_duration::UInt64
     has_pluto_hook_features::Bool
-    did_mention_expansion_time::Bool=false
+    did_mention_expansion_time::Bool
 end
 const cell_expanded_exprs = Dict{UUID,CachedMacroExpansion}()
 
@@ -176,7 +176,7 @@ replace_pluto_properties_in_expr(other; kwargs...) = other
 "Similar to [`replace_pluto_properties_in_expr`](@ref), but just checks for existance and doesn't check for [`GiveMeCellID`](@ref)"
 has_hook_style_pluto_properties_in_expr(::GiveMeRerunCellFunction) = true
 has_hook_style_pluto_properties_in_expr(::GiveMeRegisterCleanupFunction) = true
-has_hook_style_pluto_properties_in_expr(expr::Expr) = any(has_hook_style_pluto_properties_in_expr, expr.args)
+has_hook_style_pluto_properties_in_expr(expr::Expr)::Bool = any(has_hook_style_pluto_properties_in_expr, expr.args)
 has_hook_style_pluto_properties_in_expr(other) = false
 
 
@@ -216,7 +216,7 @@ module CantReturnInPluto
     We do macro expansion now, so we can also check for `return` statements "statically".
     This method goes through an expression and replaces all `return` statements with `throw(CantReturnInPlutoException())`
     """
-    function replace_returns_with_error(expr::Expr)
+    function replace_returns_with_error(expr::Expr)::Expr
         if expr.head == :return
             :(throw($(CantReturnInPlutoException())))
         elseif expr.head == :quote
@@ -259,12 +259,12 @@ function try_macroexpand(mod, cell_uuid, expr)
     end
 
     elapsed_ns = time_ns()
-    expanded_expr = macroexpand(mod, expr_not_toplevel)
+    expanded_expr = macroexpand(mod, expr_not_toplevel)::Expr
     elapsed_ns = time_ns() - elapsed_ns
 
     # Removes baked in references to the module this was macroexpanded in.
     # Fix for https://github.com/fonsp/Pluto.jl/issues/1112
-    expr_without_return = CantReturnInPluto.replace_returns_with_error(expanded_expr)
+    expr_without_return = CantReturnInPluto.replace_returns_with_error(expanded_expr)::Expr
     expr_without_globalrefs = globalref_to_workspaceref(expr_without_return)
 
     has_pluto_hook_features = has_hook_style_pluto_properties_in_expr(expr_without_globalrefs)
@@ -274,11 +274,13 @@ function try_macroexpand(mod, cell_uuid, expr)
         register_cleanup_function=(fn) -> UseEffectCleanups.register_cleanup(fn, cell_uuid),
     )
 
+    did_mention_expansion_time = false
     cell_expanded_exprs[cell_uuid] = CachedMacroExpansion(
-        original_expr_hash=expr_hash(expr),
-        expanded_expr=expr_to_save,
-        expansion_duration=elapsed_ns,
-        has_pluto_hook_features=has_pluto_hook_features,
+        expr_hash(expr),
+        expr_to_save,
+        elapsed_ns,
+        has_pluto_hook_features,
+        did_mention_expansion_time
     )
 
     return (sanitize_expr(expr_to_save), expr_hash(expr_to_save))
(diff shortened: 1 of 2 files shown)


src/runner/PlutoRunner.jl
```diff
@@ -512,13 +514,13 @@ function run_expression(
     # We add the time it took to macroexpand to the time for the first call,
     # but we make sure we don't mention it on subsequent calls
     expansion_runtime = if expanded_cache.did_mention_expansion_time === false
-        # Is this really the easiest way to clone a struct with some changes? Pfffft
+        did_mention_expansion_time = false
```
Pangoraw, line 517:
```suggestion
        did_mention_expansion_time = true
```

src/runner/PlutoRunner.jl
```diff
@@ -176,7 +176,7 @@ replace_pluto_properties_in_expr(other; kwargs...) = other
 "Similar to [`replace_pluto_properties_in_expr`](@ref), but just checks for existance and doesn't check for [`GiveMeCellID`](@ref)"
 has_hook_style_pluto_properties_in_expr(::GiveMeRerunCellFunction) = true
 has_hook_style_pluto_properties_in_expr(::GiveMeRegisterCleanupFunction) = true
-has_hook_style_pluto_properties_in_expr(expr::Expr) = any(has_hook_style_pluto_properties_in_expr, expr.args)
+has_hook_style_pluto_properties_in_expr(expr::Expr)::Bool = any(has_hook_style_pluto_properties_in_expr, expr.args)
```
Pangoraw, line 179: In which way does adding type assertion help? `any` should be easy to infer.

src/runner/PlutoRunner.jl
```diff
@@ -259,12 +259,12 @@ function try_macroexpand(mod, cell_uuid, expr)
     end
 
     elapsed_ns = time_ns()
-    expanded_expr = macroexpand(mod, expr_not_toplevel)
+    expanded_expr = macroexpand(mod, expr_not_toplevel)::Expr
```
Pangoraw, line 262: same question about type assertions

Pangoraw (commented):
@rikhuijzer you should be interested by tooking a look at #1881 where this improvements could be more beneficial. PlutoRunner is currently included() in each notebook process in Loader.jl and I don't think precompile happens in this case?

src/runner/PlutoRunner.jl
rikhuijzer: Normally, yes. But `expr.args` is a `Vector{Any}`:

```julia
julia> ex = :(1 + 1)
:(1 + 1)

julia> f(ex) = ex.args;

julia> @code_warntype f(ex)
MethodInstance for f(::Expr)
  from f(ex) in Main at REPL[15]:1
Arguments
  #self#::Core.Const(f)
  ex::Expr
Body::Vector{Any}
1 ─ %1 = Base.getproperty(ex, :args)::Vector{Any}
└──      return %1
```
which results in a `Union{Missing, Bool}` instead of `Bool`:
```julia
julia> g() = any(==("foo"), ex.args)

julia> @code_warntype g()
MethodInstance for g()
  from g() in Main at REPL[17]:1
Arguments
  #self#::Core.Const(g)
Body::Union{Missing, Bool}
1 ─ %1 = (==)("foo")::Base.Fix2{typeof(==), String}
│   %2 = Base.getproperty(Main.ex, :args)::Any
│   %3 = Main.any(%1, %2)::Union{Missing, Bool}
└──      return %3
```
In general, you're right indeed.
```julia
julia h() = any(==("foo"), ["bar"]);

```
has return type `Bool`.

src/runner/PlutoRunner.jl
rikhuijzer: Awesome!!! :heart:  I just wanted to go on a long debugging quest to find the cause for the failure!

src/runner/PlutoRunner.jl
rikhuijzer: Good point. For some reason, it couldn't be inferred here. I don't know why exactly. I'm happy to hear better solutions.

For the `CantReturnInPluto.replace_returns_with_error` below, I suspect that the compiler doesn't step into the module, but I'm not sure.

rikhuijzer:
Thanks a lot for the review Paul. Much appreciated :heart: 

> @rikhuijzer you should be interested by tooking a look at #1881 where this improvements could be more beneficial.

Yeah, I've waited now for a few weeks for that PR to make progress, but I figured that it's not going forward. Performance gains to `PlutoRunner` are very important for the user, I think. So, I figured that it's time to make progress.

src/runner/PlutoRunner.jl
Pangoraw: Yeah macroexpand can return anything if the expr is a macrocall (not happening in our case because the expr is wrapped in a `:toplevel`):

```julia
julia> macro x()
           nothing
       end
@x (macro with 1 method)

julia> @macroexpand(@x()) === nothing
true
```

Pangoraw: Yeah macroexpand can return anything if the expr is a macrocall (not happening in our case because the expr is wrapped in a `:toplevel`):

```julia
julia> macro x()
           nothing
       end
@x (macro with 1 method)

julia> @macroexpand(@x()) === nothing
true
```

rikhuijzer:
I've added `PlutoRunner.run_expression` to the benchmark now so that we can track it. Output on Julia 1.8-beta3 on my system:

```
               Total measured:                   29.9s            2.48GiB

 import Pluto                            1    561ms    1.9%   77.6MiB    3.1%
 compiletimes.jl                         1    28.3s   98.1%   2.36GiB   96.9%
   Pluto.Cell                            1   39.7ms    0.1%   62.3KiB    0.0%
   Pluto.Notebook                        1    540ms    1.9%   87.0MiB    3.5%
   PlutoRunner.run_expression            1    1.51s    5.2%   86.8MiB    3.5%
   SessionActions.open                   1    10.4s   36.1%    563MiB   22.6%
   SessionActions.shutdown               1    415ms    1.4%   43.0MiB    1.7%
   Configuration.from_flat_kwargs        1   13.1ms    0.0%   13.5KiB    0.0%
   Pluto.run                             1    4.41s   15.3%    558MiB   22.4%
```

src/runner/PlutoRunner.jl
fonsp: But does a return type help with inference? I thought that `f(x) = z(x)::Bool` helps, but `f(x)::Bool = z(x)` does not? Not sure why i think that

fonsp:
Could you also take a look at `PlutoRunner.format_output((a=[1,2],c=Dict("d"=>(5,6,true))))`?

fonsp (approved):

src/runner/PlutoRunner.jl
fonsp: @rikhuijzer

rikhuijzer: Sorry for late reply. I didn't see this one yet.

> But does a return type help with inference? I thought that `f(x) = z(x)::Bool` helps, but `f(x)::Bool = z(x)` does not? Not sure why i think that

Depends on where you care about the inference. In this case, the assertion on the method for `has_hook_style_pluto_properties_in_expr` doesn't help because the `any` returns an `Union{Missing,Bool}` and the type annotation on the function sets that to `Bool`.

However, this may help a lot for places where the method is **used**, because thanks to the annotation, the type can be inferred which, in turn, may allow more code to be hit during precompilation.

In this case, I just added it because `has_hook_style_pluto_properties_in_expr` **must** be a boolean at line 281. So, adding the annotation improves:
1. code readability
2. code inferability
3. debugability because it fails at an earlier point. Namely, when an incorrect type is returned

fonsp:
🎉 Awesome!! Did this really reduce `SessionActions.open` from 13s to 10s? That's a lot!

rikhuijzer:
Nope. Maar 20 MiB reducties helaas. Een gedeelte van de compile time is simpelweg verplaatst naar `PlutoRunner.run_expression` omdat die nu ook in de test zit maar ik snap de verwarring.

PR #2049, opened 2022-04-22
(technically api breaking) Remove the `notebook` field of `NewNotebookEvent`

fonsp:
I found it confusing to pass in a notebook, even though this event is fired _before/during_ its creation, and the event result will set the notebook id.

### DIFF
--- a/src/notebook/Events.jl
+++ b/src/notebook/Events.jl
@@ -85,7 +85,6 @@ end
 
 # Triggered when we open a new notebook
 struct NewNotebookEvent <: PlutoEvent
-    notebook::Notebook
 end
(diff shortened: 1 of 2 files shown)


pankgeorg (approved):
Looks good. While technically breaking, It shouldn't break any code as the notebook passed is "guaranteed" to be not useful. In any case, if someone is using this to log a UUID, please use `OpenNotebookEvent` instead.

fonsp:
Haha you weren't supposed to merge this oops

fonsp:
Oh well 💁 pretty unlikely that this is a problem

Issue #1179, opened 2021-05-19
Hide UndefVarError if the upstream cell errored


fonsp:
In this notebook, there is an error in the 2nd cell, causing **many dependent cells to error**, because they (indirectly) depend on a variable that should have been defined in the erred cell. 

<img width="863" alt="Schermafbeelding 2021-05-19 om 12 09 28" src="https://user-images.githubusercontent.com/6933510/118795628-12377c80-b89b-11eb-8d14-44910df5222a.png">

We should hide/de-emphasize these errors, and make them point to the _source error_.

---

For new contributors: This can be done entirely in the JS frontend, because the cell dependency information is available to frontend code. Computing this on the Julia side and writing this info in the error data is also okay. Also see https://github.com/fonsp/pluto-developer-instructions

lungben:
How to detect affected cells in the frontend:

- `editor_state.notebook.cell_results[cell_id].errored == true` and

- `editor_state.notebook.cell_results[cell_id].output.body.msg` starts with `UndefVarError:` and 

- at least one upstream dependency cell, as given in the values of the dict `editor_state.notebook.cell_dependencies[cell_id].upstream_cells_map`, errored (i.e. `editor_state.notebook.cell_results[cell_id].errored == true`). Which error does not matter here.

HirumalPriyashan:
![image](https://user-images.githubusercontent.com/57571152/164886203-9bf0fc60-dc5d-4125-92af-87ffc4a36f71.png)

Is this what we are expecting here?

fonsp:
Fixed by #2057

PR #2057, opened 2022-04-26
Hide UndefVarError if the upstream cell errored

HirumalPriyashan:
- Addresses https://github.com/fonsp/Pluto.jl/issues/1179

**Implementation details**
- check whether the current is erred due to a `UndefVarError`.
- find that undefined variable and check whether its due to an error in an upstream cell
- if the err is due to an err in upstream cell, replace the current cell output with upstream cell's erred output

**New behavior**
image


fonsp:
Do you think that it is possible to implement this in https://github.com/fonsp/Pluto.jl/blob/main/frontend/components/ErrorMessage.js instead? And always be sure to give a (short) description to you PR: how is it implemented, maybe explain why you chose this implementation, and show screenshots/videos of the new behaviour.

HirumalPriyashan:
@fonsp 

> Do you think that it is possible to implement this in https://github.com/fonsp/Pluto.jl/blob/main/frontend/components/ErrorMessage.js instead? 

I have taken this approach as we don't have access to cell result of errored upstream cell (`notebook.cell_results[upstream_cell_id]`) in `ErrorMessage.js`

But seems like we can get those values through `pluto_actions.get_notebook()` in `ErrorMessage.js`. Will check the possibility of that approach and update the PR.

> And always be sure to give a (short) description to you PR: how is it implemented, maybe explain why you chose this implementation, and show screenshots/videos of the new behaviour.

Updated the PR description.

HirumalPriyashan:
@fonsp 

Updated the PR using https://github.com/fonsp/Pluto.jl/blob/main/frontend/components/ErrorMessage.js .
Since there is no point of showing upstream error body in the current cell, updated the msg with a new error msg.

**New behavior**
image

fonsp:
Some ideas:
- Can you make it recursive? i.e. if you have three cells, `x = sqrt(-1)`, `y = x`, `z = y`, then the last two cells should show the redirect message instead of an undefvarerror.
- What happens when there are multiple upstream variables? E.g. `a = b = sqrt(-1)`, `a + b`.
- Can you insert a clickable link to the upstream variable(s)?
- Let's remove the stack trace

HirumalPriyashan:
> * Can you make it recursive? i.e. if you have three cells, `x = sqrt(-1)`, `y = x`, `z = y`, then the last two cells should show the redirect message instead of an undefvarerror.
> * Let's remove the stack trace

At the moment, these two are already available with the changes. Please let me know if I did not understand it correctly. But I think we can point error in the third cell to `a` since the error is due to failure of computing of `a` not `b`.
image

> * What happens when there are multiple upstream variables? E.g. a = b = sqrt(-1), a + b.

Right now, it just displays the first upstream cell which identified as errored. But sure, we can update it for multiple variables.

> * Can you insert a clickable link to the upstream variable(s)?

Sounds cool, yes we can do that.

HirumalPriyashan:
**New behavior**

*Added a clickable link for upstream variables*
*Notice for the fifth cell errors are produced for the root cause, since `c` is errored due to `a` and `b`.*

image

fonsp:
This is great, thanks @HirumalPriyashan !!

I made some small changes, including:
- [`dd0a631` (#2057)](https://github.com/fonsp/Pluto.jl/pull/2057/commits/dd0a6314423b94021945d36bae4f52c442e7de7d): this is not really internal *state* (you don't want this value to persist across renders)
- [`1bb4aae` (#2057)](https://github.com/fonsp/Pluto.jl/pull/2057/commits/1bb4aae9be0f5b1c8c8f10951cfda2154cc52d30) this fixes [this issue](https://user-images.githubusercontent.com/6933510/169162054-d80fa7be-1ccf-4a4e-8f0b-97cc45e4fb0c.png), which could be confusing/frustrating in a large cell with multiple errors.

fonsp (approved):

KronosTheLate:
I am late to the party, but want to give a huge appreciation for this fix <3

PR #2054, opened 2022-04-22
Generate precompile directives via PrecompileSignatures.jl

rikhuijzer:
After manually adding a lot of `precompile` directives in earlier PRs such as the following:

- https://github.com/fonsp/Pluto.jl/pull/1934
- https://github.com/fonsp/Pluto.jl/pull/1978

it seemed that there should be a better way. What if we could automatically generate precompile directives? Of course, guessing what types are going to be put into functions is impossible to predict without actually running code and keeping track of calls via [`SnoopCompile.jl`](https://github.com/timholy/SnoopCompile.jl), but what we **can** do is read the method signatures from the package and generate precompile directives for any concrete signatures. Yes, it's a very brute force way, but if it works it works.

This PR demonstrates the new and yet unregistered [`PrecompileSignatures.jl`](https://github.com/rikhuijzer/PrecompileSignatures.jl) package. This package reads the method signatures inside `Pluto.jl` and generates about 415 precompile directives from that.

## Benchmarks

With Julia 1.8-beta3:

### main

Precompilation time: 8 seconds
```
 Section                        ncalls     time    %tot     alloc    %tot
 ────────────────────────────────────────────────────────────────────────
 import Pluto                        1    562ms    1.9%   74.6MiB    3.0%
 compiletimes.jl                     1    29.3s   98.1%   2.38GiB   97.0%
   PlutoRunner.run_expression        1    1.41s    4.7%   79.3MiB    3.2%
   SessionActions.open               1    17.9s   59.9%   1.31GiB   53.4%
   Pluto.run                         1    4.80s   16.1%    563MiB   22.4%
 ──────────────────────────────────────────────────────────────────────── 
```

### This PR

Precompilation time: 12 seconds
```
 Section                        ncalls     time    %tot     alloc    %tot
 ────────────────────────────────────────────────────────────────────────
 import Pluto                        1    1.24s    4.4%    151MiB    7.5%
 compiletimes.jl                     1    27.1s   95.6%   1.81GiB   92.5%
   PlutoRunner.run_expression        1    1.52s    5.4%   78.0MiB    3.9%
   SessionActions.open               1    16.2s   57.1%    889MiB   44.3%
   Pluto.run                         1    4.31s   15.2%    441MiB   22.0%
 ────────────────────────────────────────────────────────────────────────
```

EDIT: Earlier versions of this comment showed less promising results. This was due to me enthusiastically removing all manually added `precompile` directives; even the ones which couldn't be automatically added by PrecompileSignatures.jl.

PR #2062, opened 2022-04-26
Reduce overspecializations in PlutoRunner.format_output

rikhuijzer:
- Continuation from https://github.com/fonsp/Pluto.jl/pull/2054. I'm creating a new PR to avoid being blocked.

In response to https://github.com/fonsp/Pluto.jl/pull/2039#issuecomment-1103845524. This PR reduces the time to first X for
```julia
Pluto.PlutoRunner.format_output((a=[1,2],c=Dict("d"=>(5,6,true))))
```
from
```
  2.126683 seconds (8.14 M allocations: 427.361 MiB, 13.08% gc time, 99.59% compilation time)
```
to
```
  1.376244 seconds (3.77 M allocations: 194.306 MiB, 17.16% gc time, 99.42% compilation time)
```
on Julia 1.8-beta3. 

The trick is very simple. I'm avoiding overspecialization, see https://timholy.github.io/SnoopCompile.jl/stable/pgdsgui/ for more information. This means that the code is less optimized, but as can be seen below that shouldn't matter in these cases. The TTFX is reduced but the `@btime` benchmark remains roughly the same.

## Details

I found the problems by running the following script many times:
```julia
using CodeTracking
using SnoopCompile
using Pluto

ex = quote
    Pluto.PlutoRunner.format_output((a=[1,2],c=Dict("d"=>(5,6,true))))
end

tinf = @snoopi_deep eval(ex)
sorted_children = sort(tinf.children; by=inclusive, rev=false)
display(sorted_children)
println()
display(tinf)
@btime eval(ex)
```
Notes:
```
## Original

InferenceTimingNode: 0.761881/2.040207 on Core.Compiler.Timings.ROOT() with 69 direct children
  1.122 ms (1763 allocations: 110.10 KiB)

## After @nospecialize on tree_data_nt_row:

InferenceTimingNode: 0.764838/1.701227 on Core.Compiler.Timings.ROOT() with 79 direct children
  1.123 ms (1761 allocations: 110.05 KiB)

## After @nospecialize on tree_data for AbstractVector

InferenceTimingNode: 0.767447/1.660461 on Core.Compiler.Timings.ROOT() with 79 direct children
  1.114 ms (1766 allocations: 110.20 KiB)

## Replace the broadcast in tree_data(::Tuple, ...)

InferenceTimingNode: 0.761647/1.537517 on Core.Compiler.Timings.ROOT() with 76 direct children
  1.115 ms (1756 allocations: 109.73 KiB)

## Replace the broadcast in tree_data(::NamedTuple, ...)

[tree_data_nt_row(z, recur_io) for z in zip(eachindex(x), x)]

InferenceTimingNode: 0.750875/1.838307 on Core.Compiler.Timings.ROOT() with 60 direct children
  1.112 ms (1749 allocations: 109.48 KiB)

## For loop for tree_data(::NamedTuple, ...)

InferenceTimingNode: 0.731788/1.396680 on Core.Compiler.Timings.ROOT() with 68 direct children
  1.118 ms (1747 allocations: 109.24 KiB)

## For loop for tree_data(::Tuple, ...)

InferenceTimingNode: 0.665682/1.304220 on Core.Compiler.Timings.ROOT() with 65 direct children
  1.110 ms (1748 allocations: 109.20 KiB)
```


src/runner/PlutoRunner.jl
```diff
@@ -1240,21 +1244,28 @@ function tree_data(@nospecialize(x::AbstractDict{<:Any,<:Any}), context::IOConte
...
-function tree_data_nt_row(pair::Tuple, context::IOContext)
+function tree_data_nt_row(@nospecialize(pair::Tuple), context::IOContext)
     # this is an entry of a NamedTuple, the first element of the Tuple is a Symbol, which we want to print as `x` instead of `:x`
     k, element = pair
     string(k), format_output_default(element, context)
 end
 
 
-function tree_data(@nospecialize(x::NamedTuple), context::IOContext)
+function tree_data(@nospecialize(x::NamedTuple), @nospecialize(context::IOContext))
```
rikhuijzer, line 1254: @fonsp Are multiple subtypes of `IOContexts` passed to Pluto during running? If yes, then this `@nospecialize` will probably help. Otherwise, it doesn't add much

fonsp: ah true, it's the same type every time. there was a point last year where we were just adding nospecialize randomly on everything. 

Do you have benchmarks for this function?

fonsp:
We could also do
```julia
Tuple[
...
for x in y
]
```
right? That's what i did to solve this problem in some other places in plutorunner. I prefer this style over `push!`, but if `push!` is better then you could also change the rest of the file

src/runner/PlutoRunner.jl
rikhuijzer: > Do you have benchmarks for this function?

Benchmark:
```julia
julia --project -ie 'using Pluto; @time @eval Pluto.PlutoRunner.tree_data((; a=[1]), IOContext(IOBuffer()))'
```
tested on Julia 1.7.2.

## Before

```
  1.355317 seconds (5.00 M allocations: 269.918 MiB, 18.98% gc time, 99.59% compilation time)
```

## After

```
  0.924939 seconds (2.42 M allocations: 132.716 MiB, 21.76% gc time, 99.49% compilation time)
```

fonsp (approved):

rikhuijzer:
> We could also do
> 
> ```julia
> Tuple[
> ...
> for x in y
> ]
> ```
> 
> right? That's what i did to solve this problem in some other places in plutorunner. I prefer this style over `push!`, but if `push!` is better then you could also change the rest of the file

I've checked that one too. When I used the list comprehension, it showed a lot of new method instances for `generator` and those kinds of functions. With the loop, it only specializes on `eachindex`.

In most cases, the difference between loops and list comprehensions isn't very big, but in this case it was on `Tuple`s. Julia loves to specialize on those.

Issue #2076, opened 2022-04-29
`update_dependency_cache!` performance bottleneck


fonsp:
Here is a profile of changing the bond value of `x` (i.e. moving the Slider) in the PlutoUI sample notebook, which takes a surprisingly long time (90ms for me):

<img width="550" alt="image" src="https://user-images.githubusercontent.com/6933510/165957177-279a9e55-c68d-483b-a2a9-f377f95c7a1a.png">

It shows that `update_dependency_cache!` is the bottleneck, taking about 2/3 of the total time. Since this is a non-essential feature, and because the topology does not change when running bonds, it should be possible to eliminate it.

[bond value benchmark.jl.txt](https://github.com/fonsp/Pluto.jl/files/8591263/bond.value.benchmark.jl.txt)

fonsp:
Maybe related, there are still lots of unresolved cells after running the notebook:

<img width="717" alt="image" src="https://user-images.githubusercontent.com/6933510/165959815-090c7232-cede-4521-94f2-c55470d730ee.png">

@Pangoraw

fonsp:
#2077 fixed the `unresolved_cells` issue, which fixes the performance issue in `sample/PlutoUI.jl.jl`, because `update_dependency_cache!`. 

However, if any cells contains a macrocall that can not be resolved, then this issue becomes noticeable again. Notice that by removing `@asdf` (the only unresolved cell), performance increases again:

https://user-images.githubusercontent.com/6933510/165965321-19da2c98-da95-4f93-bcd1-7335984e6e5b.mov

fonsp:
@Pangoraw I also traced back the problem to this point! I made https://github.com/fonsp/Pluto.jl/pull/2079, which would be equivalent to your suggestion, right?

fonsp:
#2079 This PR took out the first ⛰ in the profile:

<img width="617" alt="Schermafbeelding 2022-04-29 om 17 18 53" src="https://user-images.githubusercontent.com/6933510/165974192-dc43c430-17a8-440b-9e00-6d4674e11f19.png">

# New profile

<img width="681" alt="Schermafbeelding 2022-04-29 om 17 20 47" src="https://user-images.githubusercontent.com/6933510/165974647-ca228621-29e4-43e2-a22a-e65b4a3805be.png">

fonsp:
Fixed! Working on a PR

PR #2079, opened 2022-04-29
`resolve_topology`: maintain identity when nothing changed

fonsp:
(Part of #2076)

This PR ensures that `resolve_topology` returns exactly the same `NotebookTopology` as its input, when no cells were be resolved. Returning the same topology means that our caching system can do its magic: https://github.com/fonsp/Pluto.jl/blob/caf7326b58a1aef5b8f3ecc4b1cf6bfd4f0dfce7/src/analysis/topological_order.jl#L98-L105

### DIFF
--- a/src/analysis/data structures.jl
+++ b/src/analysis/data structures.jl
@@ -77,10 +77,10 @@ begin
         get!(aid.default, aid.container, key)
     end
     function Base.merge(a1::ImmutableDefaultDict{K,V}, a2::ImmutableDefaultDict{K,V}) where {K,V}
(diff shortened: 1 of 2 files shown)


fonsp:
This PR took out the first ⛰ in the profile:

Schermafbeelding 2022-04-29 om 17 18 53

# New profile

Schermafbeelding 2022-04-29 om 17 20 47

Pangoraw:
Nice one! Thanks to #2077, there will also be less unresolved cells all together and `resolve_topology` should be called less!

PR #2080, opened 2022-04-29
Caching for `update_dependency_cache!`

### LINKED ISSUE #2076: `update_dependency_cache!` performance bottleneck

fonsp:
Depends on #2079, fix #2076

### DIFF
--- a/src/analysis/DependencyCache.jl
+++ b/src/analysis/DependencyCache.jl
@@ -40,9 +40,13 @@ function update_dependency_cache!(cell::Cell, topology::NotebookTopology)
 end
 
 "Fills dependency information on notebook and cell level."
-function update_dependency_cache!(notebook::Notebook)
+function update_dependency_cache!(notebook::Notebook, topology::NotebookTopology=notebook.topology)
     notebook._cached_topological_order = topological_order(notebook)
-    for cell in values(notebook.cells_dict)
-        update_dependency_cache!(cell, notebook.topology)
+    
+    if notebook._cached_cell_dependencies_source !== topology
+        notebook._cached_cell_dependencies_source = topology
+        for cell in all_cells(topology)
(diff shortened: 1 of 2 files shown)


src/analysis/DependencyCache.jl
```diff
@@ -40,9 +40,13 @@ function update_dependency_cache!(cell::Cell, topology::NotebookTopology)
 end
 
 "Fills dependency information on notebook and cell level."
-function update_dependency_cache!(notebook::Notebook)
+function update_dependency_cache!(notebook::Notebook, topology::NotebookTopology=notebook.topology)
     notebook._cached_topological_order = topological_order(notebook)
-    for cell in values(notebook.cells_dict)
-        update_dependency_cache!(cell, notebook.topology)
+    
+    if notebook._cached_cell_dependencies_source !== topology
```
Pangoraw, line 46: We can use `notebook._cached_topological_order.input_topology` instead and move `     notebook._cached_topological_order = topological_order(notebook)` after the dependency cache update ?

fonsp: I had that before, but felt like it was cleaner to store it a second time, because calling `topological_order(notebook)` on its own would give a false cache for the dependency cache. We never call it on its own right now, but we might want to do that later.

We could also look into a caching package

PR #2081, opened 2022-04-30
Continue CI on error

rikhuijzer:
Currently, all CI test jobs get cancelled as soon as one job fails. This is very useful to save GitHub Actions minutes. However, Pluto's tests are a bit flakely which often causes all jobs to be cancelled for no good reason. Even worse, in many cases, this happens after 20 minutes. So, when the jobs are almost done anyway! Instead of getting a lot of valuable information, we now get just a bunch of "cancelled" reports.

EDIT: Got it. Other CI configs set `fail-fast: false` explicitly. For example, https://github.com/JuliaData/DataFrames.jl/blob/main/.github/workflows/ci.yml.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -48,4 +48,3 @@ jobs:
             
             # 🚗
             - uses: julia-actions/julia-runtest@v1
-              continue-on-error: ${{ matrix.julia-version == 'nightly' }}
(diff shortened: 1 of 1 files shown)


rikhuijzer:
Hmm. It's probably something else because it doesn't match the docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

I'll check on Slack

fonsp (approved):

Issue #763, opened 2020-12-11
When running in Linux, but a Windows path given, try to map to Linux path and strip quote marks


BioTurboNick:
I'm running Pluto in WSL on Windows, and encountering an issue with interoperability.

In File Explorer, Shift+Right Click => "Copy as path" produces the path in quotation marks, e.g. `"C:\Users\nicho\Documents\Pluto\notebook.jl"`.

To use that, I have to manually convert it to `/mnt/c/Users/nicho/Documents/Pluto/notebook.jl`.

It would be a nice quality-of-life feature if it would perform this transformation for me.

fonsp:
Thanks! 

This would be easyish to implement, WSL comes with a tool for that: https://devblogs.microsoft.com/commandline/windows10v1803/#interoperability

See my comments here:
https://github.com/tpapp/DefaultApplication.jl/issues/13

fonsp:
Yes, yes, and you can start in `src/webserver/Static.jl`, search for `open`

PR #2085, opened 2022-05-01
First commit to change to wsl path

SamarthMayya:
@fonsp, I have followed your comments and added some changes in the `open` function.

### DIFF
--- a/src/webserver/SessionActions.jl
+++ b/src/webserver/SessionActions.jl
@@ -32,6 +32,12 @@ end
 
 "Open the notebook at `path` into `session::ServerSession` and run it. Returns the `Notebook`."
 function open(session::ServerSession, path::AbstractString; run_async=true, compiler_options=nothing, as_sample=false, notebook_id::UUID=uuid1())
+    "Code to check whether the system is wsl, in which case, path is changed to corresponding wsl path"
+    if Sys.islinux() && isfile("/proc/sys/kernel/osrelease") &&
+        contains(read("/proc/sys/kernel/osrelease", String), r"Microsoft|WSL"i)
+        path = read(`wslpath -u $(path)`, String)
+    end
+    
     if as_sample
         new_filename = "sample " * without_pluto_file_extension(basename(path))
         new_path = numbered_until_new(joinpath(new_notebooks_directory(), new_filename); suffix=".jl")


SamarthMayya:
resolves #763

fonsp (changes requested):
Looks good! Can you:
- put this in a separate function
- use `try catch` to make sure that it never causes a crash
- share a video recording of the new behaviour

Thanks!!

SamarthMayya:
> Looks good! Can you:
> 
> * put this in a separate function
> * use `try catch` to make sure that it never causes a crash
> * share a video recording of the new behaviour
> 
> Thanks!!

I don't use wsl, so is there a way I can do a video recording on linux?

src/webserver/SessionActions.jl
```diff
@@ -30,8 +30,24 @@ function open_url(session::ServerSession, url::AbstractString; kwargs...)
...
+    temp_path = path
+    try
+        if Sys.islinux() && isfile("/proc/sys/kernel/osrelease") &&
+            contains(read("/proc/sys/kernel/osrelease", String), r"Microsoft|WSL"i)
+            temp_path = read(`wslpath -u $(path)`, String)
+        end
+    catch
+        temp_path = init_path
+    end
+    return temp_path
```
rikhuijzer, line 45:
```suggestion
    try
        if Sys.islinux() && isfile("/proc/sys/kernel/osrelease") &&
            contains(read("/proc/sys/kernel/osrelease", String), r"Microsoft|WSL")
            return read(`wslpath -u $(path)`, String)
        end
    catch
        return path
    end
```

fonsp:
@pankgeorg can you test this on WSL?

src/webserver/SessionActions.jl
```diff
@@ -30,8 +30,23 @@ function open_url(session::ServerSession, url::AbstractString; kwargs...)
     return nb
 end
 
+function convert_path_to_wsl(path)
+    "Code to check whether the system is wsl, in which case, path is changed to corresponding wsl path"
+    try
+        if Sys.islinux() && isfile("/proc/sys/kernel/osrelease") &&
```
pankgeorg, line 36: I've at least one fallen into an issue where this blocks if the file is open and pluto doesn't open - Investigating!

src/notebook/PathHelpers.jl
```diff
@@ -1,5 +1,37 @@
 import Base64: base64decode
 
+# from https://github.com/JuliaLang/julia/pull/36425
+function detectwsl()
+    Sys.islinux() &&
+    isfile("/proc/sys/kernel/osrelease") &&
+    occursin(r"Microsoft|WSL"i, read("/proc/sys/kernel/osrelease", String))
+end
+
+"""
```
rikhuijzer, line 10:
```suggestion
"""
    maybe_convert_path_to_wsl(path)
    
Return the WSL path if the system is using the Windows Subsystem for Linux (WSL) and return `path` otherwise.
```

src/webserver/WebServer.jl
```diff
@@ -6,19 +6,12 @@ import .PkgCompat
...
-    occursin(r"Microsoft|WSL"i, read("/proc/sys/kernel/osrelease", String))
-end
-
 function open_in_default_browser(url::AbstractString)::Bool
     try
         if Sys.isapple()
             Base.run(`open $url`)
             true
-        elseif Sys.iswindows() || detectwsl()
+        elseif Sys.iswindows() || detectwsl()  #= see PathHelpers.jl =#
```
rikhuijzer, line 14:
```suggestion
        elseif Sys.iswindows() || detectwsl()
```
Non-idiomatic Julia making it hard to read/understand

src/notebook/PathHelpers.jl
rikhuijzer: You can decide to drop the `maybe_convert_path_to_wsl(path)` part, but I think starting with "Return" is a good idea for readability. Otherwise, the docstring is more a comment and therefore confusing for readers

src/webserver/SessionActions.jl
pankgeorg: I'm using the `detectwsl()` function instead of inlining it, even though I'm scared it blocks, because that's what microsoft suggests. I would do `uname -r` if it wasn't 'suggested'.

src/notebook/PathHelpers.jl
pankgeorg: Yeap, thanks for the review!

src/webserver/WebServer.jl
pankgeorg: Agreed!

rikhuijzer (approved):
Looks good to me

pankgeorg:
> @pankgeorg could you explain how these changes solve the problem of Pluto not opening when the file is open somewhere else😅

They don't; I replied to myself saying I'm not changing this part of the code! The code didn't work because the function wasn't invoked in the right place and didn't remove the `\n` in the end. This, plus a small reorganization is what you see in my changes!

SamarthMayya:
> Looks good! Can you:
> 
> * put this in a separate function
> * use `try catch` to make sure that it never causes a crash
> * share a video recording of the new behaviour
> 
> Thanks!!

@pankgeorg could you share a video recording of the functionality?

pankgeorg:
> > Looks good! Can you:
> > 
> > * put this in a separate function
> > * use `try catch` to make sure that it never causes a crash
> > * share a video recording of the new behaviour
> > 
> > Thanks!!
> 
> @pankgeorg could you share a video recording of the functionality 😅? Thanks in advance!

rec

rikhuijzer:
@pankgeorg let's merge?

fonsp:
Awesome @SamarthMayya!

PR #2096, opened 2022-05-05
Delete method docs with method

### LINKED ISSUE #2071: markdown/docstring issue with multiple dispatch, docstrings multiply
Thank you for reporting an issue about Pluto! Let's get it fixed!

# To report an issue, you need two things:
1. 📹

https://user-images.githubusercontent.com/24848149/165875676-90485eb0-612e-4bec-8727-affd445d1da1.mov

2. 📝
https://gist.github.com/wisehe/58ffd0d6d5b5bd2f92af9d0891c26029
3. 🤕 
When I have multiple functions with the same name, any markdown above those functions seems to "duplicate" itself each time the cell is run. If I remove the other functions, the behavior will go back to normal once the cell with markdown is run.

Pangoraw:
Fixes #2071.

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -688,6 +688,28 @@ end
 isfromcell(method::Method, cell_id::UUID) = endswith(String(method.file), string(cell_id))
 
 """
+    delete_method_doc(m::Method)
+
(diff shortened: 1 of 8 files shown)


fonsp:
and thanks for the refactor!

fonsp (approved):

PR #2098, opened 2022-05-09
Slightly reduce TTFX for `SessionActions.open`

rikhuijzer:
I've been running 

```julia
ex = quote
    session = Pluto.ServerSession()
    session.options.server.disable_writing_notebook_files = true
    session.options.evaluation.workspace_use_distributed = false
    path = joinpath(pkgdir(Pluto), "sample", "Basic.jl")
    Pluto.SessionActions.open(session, path; run_async=false)
end
tinf = @snoopi_deep eval(ex)

sorted_children = if true
      sorted_children = sort(tinf.children; by=inclusive, rev=false)
      display(sorted_children)
      println()
      sorted_children
      # To figure out where those things are defined, use
      # ascend(Core.MethodInstance(sorted_children[end]))
  end
display(tinf)
@btime eval(ex)
```

in Julia 1.8-beta3 to find methods which take long to compile. Unfortunately, there are 40 or so methods here which all take 0.01-0.09 seconds. This PR gets rid of some of them.

## Benchmarks

Running multiple since the time varies and SnoopCompile.jl doesn't show allocations:

`main`: 16.3 seconds
this branch: 15.9 seconds
`main`: 16.2 seconds
this branch: 15.9 seconds
`main`: 16.3 seconds

Okay, so 0.3 seconds gain. It's something

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -320,8 +320,8 @@ end
 # Possible leaf: value
 # Like: a = 1
 # 1 is a value (Int64)
-function explore!(value, scopestate::ScopeState)::SymbolsState
-    # includes: LineNumberNode, Int64, String, 
+function explore!(@nospecialize(value), scopestate::ScopeState)::SymbolsState
+    # includes: LineNumberNode, Int64, String, Markdown.LaTeX, DataType and more.
     return SymbolsState()
 end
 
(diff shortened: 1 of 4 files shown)


rikhuijzer:
In 82a2c13, the explicit `precompile` call for the method with signature is removed again
```julia
function load_notebook_nobackup(@nospecialize(io::IO), @nospecialize(path::AbstractString))::Notebook
```
This will be automated by https://github.com/rikhuijzer/PrecompileSignatures.jl/pull/9.

PR #2103, opened 2022-05-11
Fix error message when opening a port fails

yha:
This error message is too specific, and is misleading when listening on the port fails due to missing permissions.


Pangoraw:
Thanks!

PR #2113, opened 2022-05-18
Use URI(uri, ...) instead of merge(uri, ...)

### LINKED ISSUE #2112: Warning: `merge(uri::URI; kw...)` is deprecated, use `URI(uri; kw...)` instead.
Every time I use `Pluto`, I get this warning:
```julia
julia> using Pluto

julia> Pluto.run()
[ Info: Loading...
┌ Warning: `merge(uri::URI; kw...)` is deprecated, use `URI(uri; kw...)` instead.
│   caller = ip:0x0
└ @ Core :-1
```

rikhuijzer:
Closes #2112.

This newer syntax is available since at least Nov 28, 2020 as far as I can tell (https://github.com/JuliaWeb/HTTP.jl/pull/627). Around that time, HTTP was at version 0.8, so this change should work for the HTTP range of 0.9.1-0.9.17 which is specified in `Project.toml`.

### DIFF
--- a/src/webserver/WebServer.jl
+++ b/src/webserver/WebServer.jl
@@ -358,7 +358,7 @@ function pretty_address(session::ServerSession, hostIP, port)
     else
         root
     end
-    merge(HTTP.URIs.URI(new_root), query=url_params) |> string
+    string(HTTP.URI(HTTP.URI(new_root); query=url_params))
 end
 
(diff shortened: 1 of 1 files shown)


rikhuijzer:
I'm gonna merge this since it can be easily reverted if needed and because the tests pass.

Issue #1952, opened 2022-02-24
Use bundled editor.js in static HTML export


fonsp:
I noticed that this can take down time-to-first-paint by 50% on a first visit!

@dralletje Why are we not doing this yet? You left this mysterious comment:

https://github.com/fonsp/Pluto.jl/blob/404bfcea6e8e66a89b8c43ca0d780b972cf5acc2/src/notebook/Export.jl#L32-L33

Why would we need a new Parcel workflow? Can't we use https://cdn.jsdelivr.net/gh/fonsp/Pluto.jl@0.18.1/frontend-dist/ ? We can check whether the export is being generated for exactly a Pluto release, and we can check the existence of `frontend-dist` locally to see if it will exist on the CDN. Should work?

PR #2132, opened 2022-05-25
Automatically generate SRI attributes in bundle

fonsp:
I forgot why I did this........ 😵‍💫  but might be useful for #1952 depending on how we solve that

### DIFF
--- a/frontend-bundler/add_sri.js
+++ b/frontend-bundler/add_sri.js
@@ -1,3 +1,8 @@
+// Go through all the CSS/JS imports in an HTML file, and add SRI attributes. More info here:
+// https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#examples
+
+// I really really tried to do this using a parcel plugin but it's "not possible". So right now this is just a separate script that you run with the html filenames as arguments.
+
 let path = require("path")
 let fs = require("fs/promises")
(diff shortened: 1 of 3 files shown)


fonsp:
I really really tried to do this using a parcel plugin but it's "not possible"

fonsp:
This script adds the `integrity` attribute, but not the `crossorigin` attribute. You can only use integrity on remote (jsdelivr) scripts if `crossorigin=anonymous`, so this caused the static html exports in 0.19.7 to break :(

I created a PR to fix this but it's "out of scope" https://gitlab.com/Abogical/posthtml-sri/-/merge_requests/5

Abogical:
Hello @fonsp !

I might be a year late, but I just wanted to apologize if my tone was rude or arrogant in the discussion we had on that PR.

I like Julia's ecosystem, especially Pluto. so I was pleasantly surprised to see my package being used there. It might not mean much, but I appreciate it.  And thanks for your work on Pluto ❤️

fonsp:
Hey @Abogical, thanks for your message!

It's very important to stick to your scope and be strict about it, sometimes that means disappointing users :) I think I was just surprised that it was out of scope, but no worries.

Have a good time with Pluto, and we are always looking for contributors who know JavaScript!

-fons

PR #2133, opened 2022-05-25
Use bundled editor.html for static export if possible

fonsp:
Because it loads much faster! See #1952 

TODO:

- [x] actually run this code
- [x] see if it works
- [x] Maybe also use `replace_with_cdn` for the non-bundled case? 
- [x] And use URIs.jl to resolve the relative URLs?


Project.toml
```diff
@@ -25,6 +25,7 @@ RelocatableFolders = "05181044-ff0b-4ac5-8273-598c1e38db00"
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
 TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
+URIs = "5c2747f8-b7ea-4ff2-ba2e-563bfd36b1d4"
```
fonsp, line 28: (this was already an indirect dependency through HTTP)

Issue #2140, opened 2022-05-30
a=myfunction(a,...)


filchristou:
The line ``a = rand(a, 4)`` freezes Pluto, while normally it should give ``UndefVarError: a not defined``  as in REPL.
The same happens for different function like `f = findall(f, 5)`

tested on Pluto v0.19.5

Pangoraw:
Thanks for the report! Could you share a notebook and/or a video showcasing the problem with steps to reproduce ?

rikhuijzer:
I can reproduce on Linux 64 bit with Julia 1.7.2 and Pluto 0.19.5. Running a cell with `a = rand(a, 4)` in an empty notebook will run for a few seconds and then the bar at the top stops at about 1/3rd of the width showing that things are still running. Even after a few minutes, doing things such as adding a new cell are still blocked

filchristou:
I now experience that it is more general. 
inserting a cell like ``a = a + 1`` everywhere in the notebook (even empty) will have the same effect.

rikhuijzer:
I created a notebook which should freeze and opened it directly via `Pluto.SessionActions.open(Pluto.ServerSession(), path)`. This didn't freeze, so it looks like a front end problem.

According to `git bisect`, the problem was introduced in #2057 (@fonsp).

HirumalPriyashan:
Sent a fix in https://github.com/fonsp/Pluto.jl/pull/2142

PR #2142, opened 2022-06-01
Fix freezing for `a = a`

HirumalPriyashan:
*Purpose*
- To fix https://github.com/fonsp/Pluto.jl/issues/2140
- The error was introduced by revisiting same cell back and forth, which makes the function to recursive infinitely.

*Approach*
- Introduced a list to keep track on visited cells to avoid revisiting.
- This should also make the process of identifying of errored upstream cells much faster.

image

### DIFF
--- a/frontend/components/ErrorMessage.js
+++ b/frontend/components/ErrorMessage.js
@@ -141,7 +141,7 @@ export const ErrorMessage = ({ msg, stacktrace, cell_id }) => {
             pattern: /^UndefVarError: (.*) not defined\.?$/,
             display: (/** @type{string} */ x) => {
                 const notebook = /** @type{import("./Editor.js").NotebookData?} */ (pluto_actions.get_notebook())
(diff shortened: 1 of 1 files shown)


fonsp:
@HirumalPriyashan Can you try to get our VS Code formatter set up? Be sure to clone the whole repository and open that folder with VS Code as its own workspace. Maybe remove other formatting extensions, and install all our recommended extensions. https://github.com/JuliaPluto/pluto-developer-instructions

fonsp:
> @HirumalPriyashan Can you try to get our VS Code formatter set up? Be sure to clone the whole repository and open that folder with VS Code as its own workspace. Maybe remove other formatting extensions, and install all our recommended extensions. [JuliaPluto/pluto-developer-instructions](https://github.com/JuliaPluto/pluto-developer-instructions)

And once you found the solution, consider making a PR to https://github.com/JuliaPluto/pluto-developer-instructions to add anything we missed!

PR #2139, opened 2022-05-29
Return a `Rich` object instead of `Dict` for `show_richest`

rikhuijzer:
I was trying to reduce the TTFX for `SessionActions.open` and ended up in `PlutoRunner` again.

With this PR, `show_richest` will now always return a `Rich` object instead of a `Dict`. This improves readability and also makes the compiler a bit happier. It reduces allocations from ~90 MB to ~80 MB on the TTFX benchmark of #2117 on `@btime` it reduces time by 17 μs (wow wow). Anyway, the main argument for this PR is code readability.

PR #2145, opened 2022-06-02
Document serialization and reduce duplication in `PlutoRunner`

rikhuijzer:
- Documents that `MimedOutput` should not be a struct (#2139)
- Adds the `objectid2str` and `circular` methods to `PlutoRunner` to reduce duplicaiton

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -825,7 +825,7 @@ const table_column_display_limit_increase = 30
 
 const tree_display_extra_items = Dict{UUID,Dict{ObjectDimPair,Int64}}()
 
-# This is not a struct to make it easier to pass these objects between distributed threads.
+# This is not a struct to make it easier to pass these objects between distributed processes.
 const FormattedCellResult = NamedTuple{(:output_formatted, :errored, :interrupted, :process_exited, :runtime, :published_objects, :has_pluto_hook_features),Tuple{PlutoRunner.MimedOutput,Bool,Bool,Bool,Union{UInt64,Nothing},Dict{String,Any},Bool}}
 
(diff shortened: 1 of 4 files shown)


fonsp (approved):
Looks nice, thanks again!!

src/runner/PlutoRunner.jl
```diff
@@ -24,7 +24,9 @@ import Logging
 
 export @bind
 
+# This is not a struct to make it easier to pass these objects between distributed threads.
 MimedOutput = Tuple{Union{String,Vector{UInt8},Dict{Symbol,Any}},MIME}
```
fonsp, line 28: should this also be a `const`?

fonsp: if not, we should add a comment why

src/runner/PlutoRunner.jl
```diff
@@ -823,6 +825,7 @@ const table_column_display_limit_increase = 30
 
 const tree_display_extra_items = Dict{UUID,Dict{ObjectDimPair,Int64}}()
 
+# This is not a struct to make it easier to pass these objects between distributed threads.
```
Pangoraw, line 828: Nice! Small nitpick :see_no_evil: 
```suggestion
# This is not a struct to make it easier to pass these objects between distributed processes.
```

src/runner/PlutoRunner.jl
rikhuijzer: Yep. Tests pass.

src/runner/PlutoRunner.jl
rikhuijzer: Thanks! Good catch

Issue #2144, opened 2022-06-01
ANSI colored output in captured stdout


fonsp:
_Originally posted by @diegozea in https://github.com/fonsp/Pluto.jl/issues/894#issuecomment-1143816541_

I found this while using programs designed for the terminal, distributed in the Julia ecosystem as jll packages. For example:

```julia
# ╔═╡ 4c4d8eaa-87ae-4f2b-8f33-4cced2cc824c
using FastME_jll

# ╔═╡ 0505615f-bf25-4316-923e-eb4dcef34129
fastme() do _fastme
	run(`$_fastme -h`)
end
```
Here the problem is not only color code but also bold character and other things I have no idea about:
Pluto:
![image](https://user-images.githubusercontent.com/2822757/171451756-836d7b37-6312-4435-8afd-aeeed0d25e16.png)

Terminal:

![image](https://user-images.githubusercontent.com/2822757/171451866-11f6b262-d9b9-4027-a909-3016afa79106.png)

fonsp:
This is different from #894 because stdout/stderr is captured and displayed as a log. This should be relatively easy to implement, take a look at how we use AnsiUp in https://github.com/fonsp/Pluto.jl/blob/v0.19.5/frontend/components/PkgTerminalView.js .

fonsp:
@HirumalPriyashan maybe? ☺️

diegozea:
Thanks! It works perfectly now :smiley: 

![image](https://user-images.githubusercontent.com/2822757/173821204-4b0fcc59-a287-465b-9984-ad40c68e7539.png)

Issue #2088, opened 2022-05-02
Feature request: double click the cell output to show the code


barucden:
It is currently possible to show/hide code using the quite small "eye" button next to the cell. I frequently use it to hide code containing markdown to display only the nicely formatted output. Sometimes, after hiding the code, I would like to display the code again and edit the wording etc., which is currently only possible by clicking the "eye" button.

I was thinking that it would be more convenient to show the code by double-clicking the cell output. The cursor might even jump inside of the cell code. I believe this would be also useful for cells producing figures or even plain julia code.

Would it be desired by other users too? And is it possible to implement technically?

Thanks!

gustaphe:
Another advantage of this is it could be temporary, like is the case with erroring cells, so that shift+enter hides the source again.

fonsp:
We could add a button to copy output to the cell context menu (which only shows when the output mime is `text/plain`)!

![Schermafbeelding 2022-06-03 om 00 24 20](https://user-images.githubusercontent.com/6933510/171748349-3304ca72-5b03-4504-b8ee-2eb624c9ade7.png)

gustaphe:
I think there is some confusion here. The suggestion and the new title do not reflect the original feature request.

fonsp:
Right! Normally, when an issue is a *direct feature request* instead of *problem statement*, we either convert it to a discussion or submit it to our feedback system, and close it: https://github.com/fonsp/Pluto.jl/issues/182

This time, I just changed the implementation to something that works with in our context, and I decided to keep the same issue instead of closing it and creating a new one.

barucden:
Oh, I apologize for that. I now see that I should have used a different communication channel.

However, I would still like to hear what other people think about the original proposal. I, for instance, liked the @gustaphe's idea of showing the code only temporarily. Should I just open a discussion for that (it seems that sending a message through the feedback system does not allow for a discussion)?

I don't see how the new proposal is related to my original one so I cannot comment on that.

fonsp:
Oh oops, I misread the original post 😅

fonsp:
My concern with double clicking is that it already has a function in selection (select the whole word). 

I think that the feature suggestion makes sense from a purely IDE viewpoint, but I am worried that the behaviour is unexpected for notebooks that contain more prose, like https://computationalthinking.mit.edu/Spring21/hw3/ We want Pluto notebooks to be both: an IDE, but it should also look and feel like a web article.

gustaphe:
[...]

PR #2148, opened 2022-06-02
ANSI colored output in captured stdout

HirumalPriyashan:
- addresses https://github.com/fonsp/Pluto.jl/issues/2144

image

### DIFF
--- a/frontend/components/Logs.js
+++ b/frontend/components/Logs.js
@@ -3,6 +3,7 @@ import { cl } from "../common/ClassTable.js"
 import { html, useState, useEffect, useLayoutEffect, useRef, useMemo } from "../imports/Preact.js"
 import { SimpleOutputBody } from "./TreeView.js"
 import { help_circle_icon, open_pluto_popup } from "./Popup.js"
+import AnsiUp from "../imports/AnsiUp.js"
 
 // Defined in editor.css
 const GRID_WIDTH = 10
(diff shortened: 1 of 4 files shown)


fonsp:
Nice! You can set `:color => true` here https://github.com/fonsp/Pluto.jl/blob/d05bccc6dd60434dfae15756a3a5fd25bd800444/src/runner/PlutoRunner.jl#L923 to tell Julia packages that they can write to `stdout` in color. 

Can you also make a screenshot of Crayons.jl?

HirumalPriyashan:
image

fonsp:
Nice! Looks like we were using AnsiUp incorrectly, which is why the second cell in your screenshot was shown in red (this state was remaining after rendering the first cell). I added a commit to fix that

fonsp:
image

Maybe #2088 next? :)

HirumalPriyashan:
@fonsp 
Let's update ansi_up version as well.
https://github.com/fonsp/Pluto.jl/blob/d1ebadef5689dbacf158e347d342def237e39c8c/frontend/editor.html#L29
@v5.1.0 supports underline, italic as well.

icweaver:
oh heck this is cool!

fonsp:
Oooh awesome, lets do that!

rafaqz:
Is this meant to work with regular `show` text/plain output now? I notice say DimensionalData.jl output is coloured in the REPL but single color in Pluto.jl, losing a lot of useful information.

PR #2154, opened 2022-06-04
Make `compute_symbolreferences` signature concrete

rikhuijzer:


src/analysis/ExpressionExplorer.jl
```diff
@@ -1132,7 +1132,7 @@ function compute_symbolreferences(ex::Any)::SymbolsState
     symstate
 end
 
-function try_compute_symbolreferences(ex::Any)::SymbolsState
```
fonsp, line 1135: This was `Any` because it can take the output of `Meta.parse(::String)` as an input, which confusingly is not always an `Expr`:

```julia
Meta.parse("123") isa Int64
```

In internal Pluto use, we always have expressions wrapped in a `:toplevel` Expr and more, which means that it is always an `Expr`. So `Any` is there to support use cases outside of Pluto's use, like people using Pluto.jl to do syntax analysis

rikhuijzer: Hmm okay I should have let you review this one. Apologies!

PR #2155, opened 2022-06-04
Refactor `ExpressionExplorer.explore!`

rikhuijzer:
This PR applies an extract method refactoring to `ExpressionExplorer.explore!`.

This PR mostly avoids compilation for a few methods, but it doesn't hit the true culprit for the TTFX which is visible via StatProfilerHTML:

image

About 90% of the time is spent inside `compute_symbolreferences` itself (not inside children). It's unclear to me what Julia is doing there exactly. (EDIT: Julia is probably recursively compiling things there since it needs to compile all the children before it can compile the parent; otherwise the LLVM cannot hardcode a link to the right child function.)

Details:

image

Note that 936 (88%) samples are spent inside `compute_symbolreferences` of which only 7 samples are spent in `typeinf_ext_toplevel` and 159 samples are spent in `explore!`. The rest is spent somewhere else.

## Benchmark

```julia
julia --project -ie 'using Pluto; @time @eval Pluto.ExpressionExplorer.compute_symbolreferences(:(1 + 1))'
```
and
```julia
julia> ENV["PLUTO_TEST_ONLY_COMPILETIMES"] = "true"

pkg> test
```

### Julia 1.7.3

Before (`main`):

```
  2.879041 seconds (2.13 M allocations: 114.277 MiB, 7.42% gc time, 99.97% compilation time)
```
```
 ────────────────────────────────────────────────────────────────────────
                                              Time          Allocations
             Total measured:                 61.3s            3.04GiB

 Section                        ncalls     time    %tot     alloc    %tot
 ────────────────────────────────────────────────────────────────────────
 import Pluto                        1    1.70s    3.4%    148MiB    5.1%
 compiletimes.jl                     1    49.0s   96.6%   2.71GiB   94.9%
   PlutoRunner.run_expression        1    2.48s    4.9%    134MiB    4.6%
   SessionActions.open               1    32.7s   64.6%   1.69GiB   59.1%
   Pluto.run                         1    6.11s   12.1%    433MiB   14.8%
 ────────────────────────────────────────────────────────────────────────
```

After (this PR):

```
  2.449334 seconds (773.73 k allocations: 42.148 MiB, 4.68% gc time, 99.96% compilation time)
```

```
 ────────────────────────────────────────────────────────────────────────
                                              Time          Allocations
             Total measured:                 60.6s            3.02GiB

 Section                        ncalls     time    %tot     alloc    %tot
 ────────────────────────────────────────────────────────────────────────
 import Pluto                        1    1.76s    3.5%    198MiB    6.8%
 compiletimes.jl                     1    48.3s   96.5%   2.64GiB   93.2%
   PlutoRunner.run_expression        1    2.46s    4.9%    134MiB    4.6%
   SessionActions.open               1    31.9s   63.7%   1.61GiB   57.0%
   Pluto.run                         1    6.15s   12.3%    433MiB   14.9%
 ────────────────────────────────────────────────────────────────────────
```

So, 0.4-0.9 second reduction or so on my system. Note that the second benchmark hits some more cases so will see slightly better improvements.

### Julia 1.8-rc1:

Before (`main`):

```
  1.506095 seconds (925.74 k allocations: 48.377 MiB, 8.76% gc time, 99.96% compilation time)
```

After (this PR):

```
  1.259567 seconds (137.68 k allocations: 6.947 MiB, 99.94% compilation time)
```

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -1197,10 +1197,7 @@ hide_argument_name(x::Any) = x
 # UTILITY FUNCTIONS
 ###
 
-"Get the global references, assignment, function calls and function defintions inside an arbitrary expression."
-function compute_symbolreferences(ex::Expr)::SymbolsState
-    symstate = explore!(ex, ScopeState())
-
+function handle_recursive_functions!(symstate::SymbolsState)
     # We do something special to account for recursive functions:
     # If a function `f` calls a function `g`, and both are defined inside this cell, the reference to `g` inside the symstate of `f` will be deleted.
     # The motivitation is that normally, an assignment (or function definition) will add that symbol to a list of 'hidden globals' - any future references to that symbol will be ignored. i.e. the _local definition hides a global_.
@@ -1211,7 +1208,14 @@ function compute_symbolreferences(ex::Expr)::SymbolsState
         inner_symstate.references = setdiff(inner_symstate.references, K)
         inner_symstate.funccalls = setdiff(inner_symstate.funccalls, K)
     end
-    symstate
+    return nothing
+end
+
+"Get the global references, assignment, function calls and function defintions inside an arbitrary expression."
+function compute_symbolreferences(ex::Expr)::SymbolsState
+    symstate = explore!(ex, ScopeState())
+    handle_recursive_functions!(symstate)
+    return symstate
 end
 
 function try_compute_symbolreferences(ex::Expr)::SymbolsState
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -453,24 +453,28 @@ function explore_macrocall!(ex::Expr, scopestate::ScopeState)
     return symstate
 end
 
+function funcname_symstate!(funcname::FunctionName, scopestate::ScopeState)::SymbolsState
+    if length(funcname) == 0
+        explore!(ex.args[1], scopestate)
+    elseif length(funcname) == 1
+        if funcname[1] ∈ scopestate.hiddenglobals
+            SymbolsState()
+        else
+            SymbolsState(funccalls = Set{FunctionName}([funcname]))
+        end
+    elseif funcname[1] ∈ scopestate.hiddenglobals
+        SymbolsState()
+    else
+        SymbolsState(references = Set{Symbol}([funcname[1]]), funccalls = Set{FunctionName}([funcname]))
+    end
+end
+
 function explore_call!(ex::Expr, scopestate::ScopeState)::SymbolsState
     # Does not create scope
 
     if is_just_dots(ex.args[1])
-        funcname = ex.args[1] |> split_funcname
-        symstate = if length(funcname) == 0
-            explore!(ex.args[1], scopestate)
-        elseif length(funcname) == 1
-            if funcname[1] ∈ scopestate.hiddenglobals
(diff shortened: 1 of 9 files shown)


src/analysis/ExpressionExplorer.jl
```diff
@@ -97,14 +97,6 @@ Base.push!(x::Set) = x
...
 ###
 
-function explore_inner_scoped(ex::Expr, scopestate::ScopeState)::SymbolsState
-    # Because we are entering a new scope, we create a copy of the current scope state, and run it through the expressions.
-    innerscopestate = deepcopy(scopestate)
-    innerscopestate.inglobalscope = false
-
-    return mapfoldl(a -> explore!(a, innerscopestate), union!, ex.args, init = SymbolsState())
-end
-
```
rikhuijzer, line 107: The git diff is a mess here. This method isn't removed but moved down.

rikhuijzer:
With the unspecialized `mapfoldl` in f0724b4, the Julia 1.7.3 TTFX goes from
```
  2.449334 seconds (773.73 k allocations: 42.148 MiB, 4.68% gc time, 99.96% compilation time)
```
to
```
  2.152037 seconds (741.94 k allocations: 40.188 MiB, 99.97% compilation time)
```

fonsp:
First glance looks good! This code is heavily tested so I'm sure that this won't break anything, I leave it up to @Pangoraw to do final review.

src/analysis/ExpressionExplorer.jl
```diff
@@ -344,419 +337,528 @@ Returns whether or not an assignment Expr(:(=),...) is assigning to a new functi
...
+        # map
+        for e in itr
+            push!(applied, f(e))
         end
+        # reduce
+        out = init
+        for e in applied
+            union!(out, e)
+        end
+        return out
```
Pangoraw, line 416: We can prevent allocating and growing an entire vector here:
```suggestion
        out = init
        for e in itr
            union!(out, f(e))
        end
        return out
```

src/analysis/ExpressionExplorer.jl
```diff
@@ -806,10 +908,10 @@ explore_module_definition!(expr, scopestate; module_depth::Number = 1) = Symbols
 "Go through a quoted expression and use explore! for :\$ expressions"
 function explore_interpolations!(ex::Expr, scopestate)
     if ex.head == :$
-        explore!(ex.args[1], scopestate)
+        return explore!(ex.args[1], scopestate)::SymbolsState
```
Pangoraw, line 911: curious to why type assertions are sometimes needed :thinking:

Pangoraw (commented):
Another ttfp decrease :raised_hands: ! I am interested in knowing what are the biggest offending factors in the previous code. the lesson is that: 
 - we shouldn't write 300 lines functions?
 - we shouldn't pass many closures to a stdlib function which can results in many variants being compiled?
 - a bit of both?

src/analysis/ExpressionExplorer.jl
rikhuijzer: Not really for TTFX. Here, the compiler just couldn't figure out the return type via type inference which is not good for runtime performance. I don't think that it will affect TTFX (except that a subsequent function call in that function can now be precompiled in the precompilation phase because the compiler knows what for what type to compile...)

Pangoraw (approved):
Thanks for digging into this and for the clear explanation!

PR #2161, opened 2022-06-07
Also trigger compilation for Unicode

### LINKED ISSUE #2153: Unicode autocomplete takes long to respond
After loading a notebook and running stuff, Pluto responds immediately for normal autocomplete (for example, `contain<TAB>` completes to `contains`). For Unicode, however, the first time it takes 2 seconds here (for example, `\lambda<TAB>`).

Anyone here who can tell me which endpoint this Unicode autocomplete is hitting? I can't find it. We should be able to get the time to zero by adding a small async workload somewhere during notebook startup.

rikhuijzer:
- Closes #2153


rikhuijzer:
I find it very difficult to test whether this actually changes things. It went from 1 second to maybe 0.2? I have no idea.

@fonsp can you review this?

PR #2163, opened 2022-06-08
Set `AUTO_RELOAD_FROM_FILE_DEFAULT` to `true`

rikhuijzer:
- Let's enable https://github.com/fonsp/Pluto.jl/pull/1029?

I'm been using it as default for a few months now and it worked flawlessly.


fonsp:
Thanks for bringing up this subject again! I also have good experience with it on MacOS, which OS are you using Rik?

I still see that our file watching tests fail about 10% of the time on `main` (and I manually rerun), so I still don't feel fully confident about it. I was kind of waiting for https://github.com/JuliaPluto/BetterFileWatching.jl/pull/2 , and then switch to BetterFileWatching, which would be a more reliable and production-tested version of file watching. 

cc @Pangoraw who better understands the current instabilities and how we can improve

rikhuijzer:
> Thanks for bringing up this subject again! I also have good experience with it on MacOS, which OS are you using Rik?

I'm on three different Linux systems.

> I still see that our file watching tests fail about 10% of the time on main (and I manually rerun), so I still don't feel fully confident about it. I was kind of waiting for https://github.com/JuliaPluto/BetterFileWatching.jl/pull/2 , and then switch to BetterFileWatching, which would be a more reliable and production-tested version of file watching.

Sounds good :+1:

Pangoraw:
It would be great to have feedback on windows as it seems more brittle there (the same is currently true for https://github.com/JuliaPluto/BetterFileWatching.jl/pull/2)

lungben:
I played around with the auto reload option on Windows (Win 11, Julia 1.8 RC1) and have not observed any issue.

disberd:
I just noticed yesterday that the auto-reload from file doesn't seem to track file changes in metadata.
Do we want also to support that? I guess it would be consistent but is probably best as separate PR

fonsp:
> I just noticed yesterday that the auto-reload from file doesn't seem to track file changes in metadata. Do we want also to support that? I guess it would be consistent but is probably best as separate PR

Did you make an issue or PR @disberd  ?

disberd:
Hi @fonsp, I think I did not in the end

fonsp:
Can you make one?

fonsp:
Maybe we can enable it by default on UNIX only?

rikhuijzer:
I am not sure. It worked for me on Linux, but I don't know what the risks are. If enabled on Linux, then also Mac by default?

PR #2168, opened 2022-06-09
TopologyUpdate: don't preserve identity when cells are removed 

### LINKED ISSUE #2167: Deleted variables still seem to exist?
Sorry if this has been noted somewhere already, but I just ran into this on `main` (5614714) and v0.19.8:

https://user-images.githubusercontent.com/25312320/172898388-8bf7607a-c361-464f-9f3a-83e824d346f4.mp4

It seems like variables are still hanging around after they are "deleted"?

Pangoraw:
This lead to a subtle bug where the identity branch in
`Base.merge(::ImmutableDefaultDict, ::ImmutableDefaultDict)` would cause the
the `delete_unsafe!` to be applied to both new and old topologies :worried:.

> **Note**
> identity was already not preserved with the cell_order being different!

Fixes #2167

### DIFF
--- a/src/analysis/TopologyUpdate.jl
+++ b/src/analysis/TopologyUpdate.jl
@@ -43,7 +43,11 @@ function updated_topology(old_topology::NotebookTopology, notebook::Notebook, ce
 		new_nodes = merge(setdiffkeys(old_topology.nodes, removed_cells), updated_nodes)
 	end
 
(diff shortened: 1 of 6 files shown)


rikhuijzer (approved):

Pangoraw:
frontend tests keep failing, I will investigate later :thinking:

PR #2173, opened 2022-06-15
Allow customizing the :1234 "port hint"

fonsp:
Before this PR, you can either set a custom port (e.g. 53455), or leave it empty, and Pluto will find the first available port after 1234.

This PR makes it possible to still get an automatically assigned port, while customizing where the search starts.

This is mostly useful for custom applications that embed Pluto, that want to avoid the popular 1234 port to avoid people connecting directly to the backend.


fonsp:
cc @Illusion47586

fonsp:
merge when tests pass

PR #2177, opened 2022-06-16
Preparations for native Windows app

Illusion47586:
Added support for desktop FileSystem APIs via electron. It checks for existence of `plutoDesktop` property on the windows object, and works accordingly. Please refer to [PlutoDesktop](https://github.com/JuliaPluto/PlutoDesktop) for more information.


rikhuijzer:
@Illusion47586 Great stuff with the front end. I’ve renamed the PR title to be more in line with https://cbea.ms/git-commit/

Illusion47586:
Thanks @rikhuijzer ! I'll take care of the title guidelines when I make any new changes.

frontend/hooks/usePlutoDesktop.js
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * A hook to check whether pluto is running inside Desktop,
+ * and execute a callback if that is the case.
+ * @param {DesktopHookCallback | undefined} callback A function that takes in a window (having `plutoDesktop` as a property)
+ * as its argument, if `plutoDesktop` is not found, it is not called. **DO NOT PASS ANY FUNCTION THAT CHANGES STATE OF THE COMPONENT.**
+ * @returns A boolean denoting whether pluto is running inside Desktop or not and a condtional callback utility function.
+ */
+const usePlutoDesktop = (/** @type {DesktopHookCallback | undefined} */ callback) => {
+    const [isDesktop, setIsDesktop] = useState(false)
```
fonsp, line 15: You can use `const isDesktop = window.plutoDesktop != null`, right?

And then

```js
useEffect(() => {
        if (isDesktop) {
            // console.log(
            //   'Running in Desktop Environment! Found following properties/methods:',
            //   window.plutoDesktop
            // );
            if (callback) callback(window)
        }
    }, [callback])
```

fonsp: Actually, maybe the `usePlutoDesktop` is not really necessary? You could replace this entire file with:

```js
export const isDesktop = window.plutoDesktop != null
```

And just use a `useEffect` in ExportBanner.js

Illusion47586: Ah yes, this work too! Committed this!

Illusion47586: > Actually, maybe the `usePlutoDesktop` is not really necessary? You could replace this entire file with:
> 
> ```js
> export const isDesktop = window.plutoDesktop != null
> ```
> 
> And just use a `useEffect` in ExportBanner.js

Yeah I added for any possible future usage, although yes, we can remove it altogether and it would be just fine.

fonsp: Nice, let's remove it for now. 

100% of bugs are caused by code, so less code = less bugs

frontend/components/Editor.js
```diff
@@ -1046,6 +1058,27 @@ patch: ${JSON.stringify(
...
         }
 
+        this.desktop_submit_file_change = async () => {
+            window.plutoDesktop?.ipcRenderer.once("PLUTO-MOVE-NOTEBOOK", async (/** @type {string | undefined} */ loc) => {
+                try {
+                    if (!!loc)
+                        await update_notebook((notebook) => {
+                            notebook.in_temp_dir = false
+                            notebook.path = loc
+                        }, false)
```
fonsp, line 1068: Can you change this to the following, so that we don't need to modify the `update_notebook` function?

```js
await this.setStatePromise(
    immer((notebook) => {
        notebook.in_temp_dir = false
        notebook.path = loc
    })
)
```

frontend/components/Editor.js
```diff
@@ -1046,6 +1058,27 @@ patch: ${JSON.stringify(
...
+                    // @ts-ignore
+                    document.activeElement?.blur()
```
fonsp, line 1070:
```suggestion
                    document.activeElement?.blur?.()
```

then you don't need the `@ts-ignore`

frontend/components/Editor.js
```diff
@@ -1046,6 +1058,27 @@ patch: ${JSON.stringify(
...
+                } catch (error) {
+                    alert("Failed to move file:\n\n" + error.message)
+                } finally {
+                    this.setState({ moving_file: false })
+                }
+            })
+
+            this.setState({ moving_file: true })
```
fonsp, line 1078: Why is it being set to `true` here? Can you add a comment to explain how the value will become `false` later?

frontend/components/ExportBanner.js
```diff
@@ -29,20 +31,44 @@ const Square = ({ fill }) => html`
...
+    return id
+}
+
 export const ExportBanner = ({ onClose, notebookfile_url, notebookexport_url, start_recording }) => {
+    const { isDesktop, conditionalCallback } = usePlutoDesktop(undefined)
+
+    const id = getId(notebookfile_url)
+
+    const exportNotebook = (/** @type {{ preventDefault: () => void; }} */ e, /** @type {0 | 1 | 2 | 3} @note see *types.d.ts* for more info.*/ type) => {
+        e.preventDefault()
```
fonsp, line 45: (Don't `preventDefault` if you are not in desktop mode, or the export in regular Pluto won't work anymore)

frontend/hooks/usePlutoDesktop.js
Illusion47586: >
> 100% of bugs are caused by code, so less code = less bugs
>
Haven't read anything truer.
Fixed it :)

frontend/components/ExportBanner.js
Illusion47586: sorted

frontend/components/ExportBanner.js
```diff
@@ -29,20 +31,44 @@ const Square = ({ fill }) => html`
 //@ts-ignore
 window.enable_secret_pluto_recording = true
 
+const getId = (/** @type {string} */ url) => {
+    const id = url.split("id=")[1]
+    return id
+}
+
 export const ExportBanner = ({ onClose, notebookfile_url, notebookexport_url, start_recording }) => {
```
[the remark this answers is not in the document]
Illusion47586: sorted

frontend/components/Editor.js
```diff
@@ -1046,6 +1058,27 @@ patch: ${JSON.stringify(
             }
         }
 
+        this.desktop_submit_file_change = async () => {
+            window.plutoDesktop?.ipcRenderer.once("PLUTO-MOVE-NOTEBOOK", async (/** @type {string | undefined} */ loc) => {
```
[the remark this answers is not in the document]
Illusion47586: done

frontend/components/Editor.js
Illusion47586: This one is not only related to changing the state, I need the extra parameter for what is happening on line [990](https://github.com/Illusion47586/Pluto.jl/blob/f12e07063a44e61c4909a34bce7428c53687bf99/frontend/components/Editor.js#L990)

frontend/types.d.ts
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * This namespace is meant for [PlutoDesktop](https://github.com/JuliaPluto/PlutoDesktop)
+     * related types and interfaces.
+     */
+    namespace Desktop {
+        /**
+         * This type has to be in sync with the type "Channels"
+         * defined in PlutoDesktop/{branch:master}/src/main/proload.ts
+         */
+        type Channels = "ipc-example"
```
ctrekker, line 11: Could this be changed to something more specific to its function?

```suggestion
        type Channels = "ipc-plutodesktop"
```

frontend/components/Editor.js
fonsp: You need it for what exactly? If line 990 is false, then it will run the snippet from my comment, plus [this](https://github.com/Illusion47586/Pluto.jl/blob/f12e07063a44e61c4909a34bce7428c53687bf99/frontend/components/Editor.js#L944-L985), but I don't think you need it. Does it not work otherwise?

fonsp:
Thanks for looking at my comments! I will try to get this merged by Friday. Do you need this to be released in a Pluto release for the beta release of Pluto desktop? What else do you need from the pluto side?

frontend/types.d.ts
Illusion47586: I mean I can definitely do it, and the fact is that this channel is not actually needed. I just made it so that if anyone wants to use `sendMessage` they can modify the `Channels` to make it type-safe.
Should I remove it?

frontend/components/Editor.js
Illusion47586: Tried it and it is working! pushed it.

frontend/types.d.ts
ctrekker: If it's not used right now I'd say remove it. In the future if someone needs ipc they can add the channel type back in

Illusion47586: done!

frontend/components/ExportBanner.js
```diff
@@ -29,20 +30,37 @@ const Square = ({ fill }) => html`
...
                 <div class="export_title">export</div>
                 <!-- no "download" attribute here: we want the jl contents to be shown in a new tab -->
-                <a href=${notebookfile_url} target="_blank" class="export_card">
+                <${elType} href=${notebookfile_url} target="_blank" class="export_card" onclick=${(e) => exportNotebook(e, 0)}>
                     <header><${Triangle} fill="#a270ba" /> Notebook file</header>
                     <section>Download a copy of the <b>.jl</b> script.</section>
-                </a>
-                <a href=${notebookexport_url} target="_blank" class="export_card" download="">
+                </${elType}>
+                <${elType}
```
fonsp, line 54: I'm changing the nice `eltype` = `isDesktop ? "button" : "a"` back to just an `a` tag, based on my small test that works just fine. Perhaps the problem was using `onclick` vs `onClick` in the attribute.

Illusion47586: got it!

fonsp (approved):

fonsp:
@Illusion47586 Just merged! Can you check whether Pluto desktop works with the `main` branch?

Illusion47586:
It works using `Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="main")`.
Also, is this top text-box is added by you?
image

fonsp:
Ah sorry, can you make a new PR to remove the textbox?

fonsp:
You can just hide the entire FilePicker in the editor on desktop, right?

Illusion47586:
We decided upon hiding only the text part, where it shows the location.

PR #2178, opened 2022-06-16
Track dependency of `skipped_as_script` cells

disberd:
Following the discussion in https://github.com/fonsp/Pluto.jl/discussions/2175

This is a first draft PR to also comment out in the file cells that depend on ones that are directly flagged as _disabled in file_.

Implementation was quite straightforward following the similar dependency on disabled cells, but an update trigger is also needed outside of the `reactive_run_core!` function as this flag might create a notebook update without causing cells running.

Didn't put any styling difference in the cells visuals but modified the click tooltip to differentiate between directly and indirectly skipped cells.


disberd:
@greimel Since you expressed interest, let me know if you find issues if trying this out!

greimel:
Great! I've just tried and it doesn't seem to work. Could it be that the flag is still there?

I don't know anything about HTML and JS, but this doesn't seem to work.

https://user-images.githubusercontent.com/6280307/174314935-627141d0-8800-42ca-b80f-0d4ec81bfd15.mov

Pangoraw:
try this to enable it:
```html
<script>
if (window.localStorage.getItem("ENABLE_SKIP_AS_SCRIPT") !== "true") {
    window.PLUTO_TOGGLE_SKIP_AS_SCRIPT()
}
</script>
```
 as your current code is toggling then reloading then re-toggling, etc...

greimel:
I've played around with it a bit and it seems to work well. Unfortunately it just comes after finishing my course preparations. Otherwise I could have battle-tested it a bit more :-)

greimel (approved):

greimel:
It would be nice if there was a way to jump from the dependent cell to the one that's been skipped. This would be similar to `disable`

image

You can click on the arrow to jump to the cell.

disberd:
Indeed I didn't notice you could do so with disabled cells, implementation it should be straightforward but it would be nice to have a button to trigger this.

One idea could be to re-use the context menu button. Making it only go to the parent cell for a cell that is indirectly disabled

Pangoraw:
I changed the styling of the marker so that the tooltip is only on the left border and not on the entire cell output (which meant that the cell output could not be interacted with).

image

What do you think ?

src/webserver/Dynamic.jl
```diff
@@ -111,6 +111,7 @@ function notebook_to_js(notebook::Notebook)
                 "code" => cell.code,
                 "code_folded" => cell.code_folded,
                 "metadata" => cell.metadata,
+                "depends_on_skipped_cells" => cell.depends_on_skipped_cells,
```
Pangoraw, line 114: I think this makes more sense in `"cell_results"` like `"depends_on_disabled_cells"` meaning that you won't need to initialize it everywhere in the frontend.

disberd: Yeah I didn't realize you would have to initialize it before going into the typescript error.
I put it into the cell_inputs as it seemed more appropriate there logically (I see it more as something related to the cell input rather than the cell result).
But indeed it's probably easier to make it part of the cell results

Pangoraw: > I see it more as something related to the cell input rather than the cell result

The rationale is that cell inputs are set by the user whereas cell results can be determined from all inputs (code, topology, etc...).

disberd: That also makes sense! Just change the code accordingly

disberd:
I implemented a few changes:

I added a stripes to the visual indicator of indirectly skipped cells to differentiate them and changed the text of the tooltip a bit to also provide a jump to the upstream cell causing the indirect skipping:
image

To provide the jump functionality, I reused the code of disabled cells so I moved the relevant functions from RunArea.js into Cell.js and made them generic instead of specific for the `disabled` metadata.
Maybe @lungben have a look as well. 
Both disabled and skipped behavior seem to work correctly but probably the javascript side on Cell.js can be made cleaner, @Pangoraw or @pankgeorg can provide feedback here.

Pangoraw (approved):
Looks good to me! I changed the appearance of dependents on skip as script to use a lighter/darker background instead of the gradient which I think looked too much like the cell running indicator, what do you think ?

disberd:
@Pangoraw that is also OK with me. From my point of view this is ready to be merged

greimel:
It seems to me that the feature is still behind a flag. Shouldn't the flag be removed?

disberd:
You are right! I just removed the flag in the last commit

Pangoraw:
Let's merge when tests pass! It would be great to have a wiki page/sample notebook about this feature and cell disabling.

fonsp:
This is great! Just used it in https://github.com/JuliaPluto/PlutoUI.jl/pull/231 , it makes it so easy to disable all the test cells in the notebook file! Before, with the `@skip_as_script` macro I was just too lazy to put the macro in front of every cell.

In one of the notebooks, I do `using PlutoUI` to get `Slider`, `Button`, etc to test the notebook with, and I can just disable the `using PlutoUI` cell, and it will automatically disable everything I need! I think that would also be really useful with `using PlutoTest`

Issue #2188, opened 2022-06-24
Paste cell after currently selected cell does not work anymore


disberd:
The functionality added in https://github.com/fonsp/Pluto.jl/pull/1263 to paste multiple copied cells at a specific location in the notebook does not work anymore.
The paste command now simply pastes the source code of the cells:

![0c453bf5-f9d0-4be8-aaef-dd0c4487ebe4](https://user-images.githubusercontent.com/12846528/175568896-5a7f8a67-5d0a-4633-950f-0a12518ae6fc.gif)

It seems this functionality was broken by https://github.com/fonsp/Pluto.jl/pull/2131

PR #2189, opened 2022-06-24
fix cell paste at specific cell

disberd:
Fixes https://github.com/fonsp/Pluto.jl/issues/2188

The removed flags inside the serializer are carry-overs from the CM5 era, but should not be needed anymore with the separate CellInput paste handler after migration to CM6.

This should fix the mentioned issue while still avoiding the issue addressed by https://github.com/fonsp/Pluto.jl/issues/2131

I also removed an unused import inside CellInput.js (again coming from the CM5 era)


Pangoraw:
Nice! :raised_hands:

PR #2214, opened 2022-07-13
Allow environment to control custom connects to backend

pankgeorg:
- A new PlutoEvent 🎉 to notify interested parties about when an evaluation cycle is done!
- Small refactor around BinderLaunch button so it's logic is not too convoluted in the Editor.js
- Ability to provide custom launch button if some server also provides a static 🌟✨ 'environment'🌟✨ javascript!
- When an event goes all wrong, stop `@warn` ing all the event (the notebook can easily cause the logging system to crash, e.g. if you're watching and serving the logs over HTTP)


fonsp:
can you fix the big diff? the package lock should stay i think?

pankgeorg:
> can you fix the big diff? did you intend to change the package lock?

I actually did, to support the next version (16/lts vs 14?) of node if I remember correctly. But since, I didn't make it really work so I reverted the update!

fonsp:
I looked over the code so far, ✅ code looks great, I just want to run this code locally (without the extension) to make sure that the Run Local / Run Binder buttons still work

frontend/components/Editor/LaunchBackendButton.js
```diff
@@ -0,0 +1,45 @@
...
+            () => window?.pluto_injected_environment?.environment?.({ client: editor.client, editor, imports: { immer, preact } })?.custom_run_or_edit,
+            [editor.client, editor]
+        )
+        // @ts-ignore
+        if (window?.pluto_injected_environment?.provides_backend) {
+            // @ts-ignore
+            return html`<${EnvRun} editor=${editor} backend_phases=${BackendLaunchPhase} launch_params=${launch_params} />`
+            // Don't allow a misconfigured environment to stop offering other backends
+        }
+    } catch (e) {}
```
fonsp, line 21: Are you sure you don't want to log this?

pankgeorg: Currently I don't; there is no easy way to get these logs from clients and these show a setup issue - not a runtime one; should be fixed in "staging". Also there will be a failed network request along this.

src/notebook/Events.jl
```diff
@@ -83,15 +84,21 @@ FileEditEvent(notebook::Notebook) = begin
...
-# Triggered when a user opens a notebook
+# Triggered when we open any notebook
 struct OpenNotebookEvent <: PlutoEvent
     notebook::Notebook
 end
 
+# Triggered when Pluto completes an evaluation loop
+struct NotebookExecutionDoneEvent <: PlutoEvent
+    notebook::Notebook
+    user_requested_run::Bool
```
fonsp, line 99: do you also want to fire this event for bond updates? maybe add a field  `from_bond_update::Bool`

frontend/editor.js
```diff
@@ -103,6 +103,28 @@ export const empty_notebook_state = ({ notebook_id }) => ({
...
+    // @ts-ignore
+    window?.pluto_injected_environment?.custom_get_statefile?.(read_Uint8Array_with_progress, without_path_entries, unpack) ??
+    (async (launch_params, initial_notebook_state_ref, set_ready_for_editor, set_statefile_download_progress) => {
+        const r = await fetch(new Request(launch_params.statefile, { integrity: launch_params.statefile_integrity ?? undefined }))
+        const data = await read_Uint8Array_with_progress(r, set_statefile_download_progress)
+        const state = without_path_entries(unpack(data))
+        console.log({ state })
+        initial_notebook_state_ref.current = state
+        set_ready_for_editor(true)
+    })
```
fonsp, line 126: I would like to remove `initial_notebook_state_ref` and `set_ready_for_editor` from the arguments of `get_statefile`, and just make `get_statefile` a function that returns `Promse<State>`. 

So I would like to change this to:

```js
const get_statefile = (launch_params, set_statefile_download_progress) => {
	const r = await fetch(new Request(launch_params.statefile, { integrity: launch_params.statefile_integrity ?? undefined }))
    const data = await read_Uint8Array_with_progress(r, set_statefile_download_progress)
    const state = without_path_entries(unpack(data))
	return state
}

// later...

	useEffect(() => {
        if (!ready_for_editor && static_preview) {
            get_statefile(launch_params, set_statefile_download_progress).then((state) => {
				console.log({state})
		        initial_notebook_state_ref.current = state
		        set_ready_for_editor(true)
			})
        }
    }, [ready_for_editor, static_preview, statefile])
```

Does that work with your hook?

fonsp: (I edited my comment with a more complete snippet)

fonsp:
You can merge yourself after addressing/dismissing those comments ✅

src/notebook/Events.jl
pankgeorg: It already runs on bond updates! `user_requested_run::Bool` is a bond update if I'm not mistaken.

PR #2216, opened 2022-07-15
fix async process issue

### LINKED ISSUE #2121: @async run(`sleep 3`) is not async
https://user-images.githubusercontent.com/6933510/169283887-63dda4f6-2dcb-48b2-b0b0-6fba489733ca.mov

cjdoris:
Fixes #2121.

Here's a cell you can test with:
```julia
begin
    cmd = `julia -e 'println(0); sleep(3); println(1);'`
    pipe = pipeline(cmd; stdout, stderr)
    proc = run(pipe; wait=false)
    sleep(0)
end
```

With the old behaviour, this cell takes 3 seconds to run, even though the running process should occur asynchronously.

With this PR, the cell finishes immediately. If you change to `sleep(2)` then you get the terminal viewer popping up with the first message printed ("0") and with `sleep(4)` you get both messages ("0" and "1"), all as you'd expect.

As @fonsp supposed, it was indeed hanging at `wait(buffer_redirect_task)`. I think what's going on is this:
- the stdout from `proc` is redirected to Julia's `stdout`
- `pipe` is created, redirecting Julia's `stdout`
- hence `pipe` does not hit EOF until `proc` terminates
- and `write(output, pipe)` loops until `pipe` hits EOF

Anyway, `close(pipe.out)` forces EOF and solves the problem. Seems to work on Windows and Linux.

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -2209,8 +2209,8 @@ function with_io_to_logs(f::Function; enabled::Bool=true, loglevel::Logging.LogL
         # If we do not close the pipe then the cell
         #   run(`julia -e 'sleep(5)'`; wait=false)
         # takes 5 seconds to complete even though it should run asyncronously. This is
-        # because wait(output, pipe) runs until EOF.
-        close(pipe.out)
+        # because write(output, pipe) runs until EOF.
+        close(pipe)
         wait(buffer_redirect_task)
     end
 
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -2206,6 +2206,11 @@ function with_io_to_logs(f::Function; enabled::Bool=true, loglevel::Logging.LogL
         redirect_stderr(default_stderr)
         close(pe_stdout)
         close(pe_stderr)
+        # If we do not close the pipe then the cell
+        #   run(`julia -e 'sleep(5)'`; wait=false)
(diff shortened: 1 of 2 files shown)


rikhuijzer (commented):
Looks like a nice fix. Can you add a test? With that, the comment can also probably be removed or replaced by a comment linking to this issue. The comment is currently a bit too long in my opinion.

Pangoraw:
Awesome find!! Thanks for looking into this! I added a test and closed the in end of the pipe as well. Let's merge when tests pass!

Pangoraw:
After further testing, there is a problem when closing the pipe the process will crash if it tries to write to its stdout, so maybe it's best to not wait the writing task (or only write from pipe to outbuffer when the cell code is done running).

cjdoris:
> After further testing, there is a problem when closing the pipe the process will crash if it tries to write to its stdout

The original PR didn't have this issue. In the example code in my original post has a process which writes to stdout after the cell terminates. Presumably it's because you changed it to close both ends of the pipe - though I don't understand why this would matter!

cjdoris:
On second thought maybe it did crash and I just never checked the return code.

Pangoraw:
I think your code is working because `run(;wait=false)` does not attach to `stdout` but outputs to `DevNull` instead:

image

cjdoris:
No there's a pipeline redirecting to stdout and stderr. And as the OP says you see the output in the Pluto UI if you sleep for long enough.

PR #2227, opened 2022-07-25
Add base_url option

### LINKED ISSUE #2225: Remove base url path sent in from proxy servers, to not get "Not found!" error?
The design of Pluto is very nice and cool! I've watched a few of the JuliaCon and other videos about it. Thanks for creating this!

My colleagues and I do a lot of collaborative data science work. Many of us use Jupyter Lab running on an HPC system, so we have direct compute access to the same data files, etc. Our university runs a "portal", where a web app helps users set up a compute allocation, then starts a Jupyter Lab session people can connect to.
But I like to use a newer Jupyter version, with my own stuff available (via singularity image, etc.). So I've figured out a way to do that via shell access, and the proxy web server (ngnix, I believe) scheme they use for the portal (still needs/uses university authentication).
Then I can access my custom server via their url scheme:
`https://sub.domain.edu/node/${host}/${port}/`

On an HPC node, where I have Julia available, I can launch . . .
`julia> Pluto.run(;launch_browser=false, host="0.0.0.0")`

There's not much to video, but this is all I see in my client workstation browser window:
Screenshot_20220723_185620

I think (hope) that means I've actually connected to the Pluto web server, but the route being sent is not known (When I try other things, the error page looks more like nginx style).

I suspect Pluto is getting sent the `node/${host}/${port}/` as part of the request (which it would obviously not know what to do with).

With my working Jupyter Lab setup, I have to include the following in my config file:
`c.ServerApp.base_url = '/node/${host}/${port}/'`

I think that is necessary, because otherwise the Jupyter Lab server will always get that `node/${host}/${port}/` as part of the request, and won't know what to do with that (those paths make no sense to it's web url API). I assume that setting just lops that part off, before sending to the Jupyter server.

Is there a similar setting available/possible for Pluto.run()?

Thanks!

P.S. I've confirmed with a local Pluto session that "Not found!" looks identical (same dark background, etc.), when I insert a nonsensical string after the port, in the browser url.
So I have some hope that I am actually able to get to Pluto server running on my HPC node, but maybe just need a way to remove that "base url".

Pangoraw:
```julia
import Pluto
Pluto.run(; base_url = "/my/custom/base/url/")
```

Closes #2225 

 - [x] Add docs in docstring
 - [x] Needs tests
 - [x] Improved 404 handling

### DIFF
--- a/src/Configuration.jl
+++ b/src/Configuration.jl
@@ -28,6 +28,7 @@ function __init__()
 end
 
 const ROOT_URL_DEFAULT = nothing
+const BASE_URL_DEFAULT = "/"
 const HOST_DEFAULT = "127.0.0.1"
 const PORT_DEFAULT = nothing
 const PORT_HINT_DEFAULT = 1234
@@ -74,6 +75,7 @@ The HTTP server options. See [`SecurityOptions`](@ref) for additional settings.
 """
 @option mutable struct ServerOptions
     root_url::Union{Nothing,String} = ROOT_URL_DEFAULT
+    base_url::String = BASE_URL_DEFAULT
     host::String = HOST_DEFAULT
     port::Union{Nothing,Integer} = PORT_DEFAULT
     port_hint::Integer = PORT_HINT_DEFAULT
@@ -218,6 +220,7 @@ end
 
 function from_flat_kwargs(;
         root_url::Union{Nothing,String} = ROOT_URL_DEFAULT,
+        base_url::String = BASE_URL_DEFAULT,
         host::String = HOST_DEFAULT,
         port::Union{Nothing,Integer} = PORT_DEFAULT,
         port_hint::Integer = PORT_HINT_DEFAULT,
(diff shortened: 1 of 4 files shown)


bdklahn:
This worked for me (as long as I made sure to add the '/' to the end of the base url string)!
Awesome! Thanks!
Maybe one can have string libraries or something similar to Python's Pathlib "automatically" add appropriate path separators between path elements, or something?

Otherwise it works fine (and reminds people web servers do somehow need to know or figure out if a path is a directory or file to be served).

Pangoraw:
Thanks for the feedback! I improved the error messages. A pathlib-like Path would be great but more inside the stdlib

bdklahn:
> Thanks for the feedback! I improved the error messages. A pathlib-like Path would be great but more inside the stdlib

Thanks for the quick response!
Yeah . . . I was browsing the code, and saw those error messages added. It seems that should be clear and easy enough for anyone.
Also, I always love to see that/when tests are added at the same as new code! Nice!

PR #2240, opened 2022-08-05
Replace Distributed with Malt soon!

### LINKED ISSUE #300: Allow using `Distributed`

savq:
After this PR, we will no longer use the Distributed stdlib to start and control the notebook Julia process (where your code runs) but we use our home-brewed package [Malt.jl](https://github.com/JuliaPluto/Malt.jl). The advantages are:
- *It is not Distributed*, which means that your code does not run as a **Distributed worker**. This fixes #300.
- The process launches faster, about 1 second with Malt instead of 4 seconds with Distributed.
- Slightly improved developer experience: the API of Malt is more intuitive for our use-case and we need fewer workarounds.
- Room for future improvement: splitting Malt into its own package makes it much easier to improve our multiprocessing functionality (because the codebase and test/benchmark suite is much smaller). This might include:
  - Supporting interrupt on Windows
  - Faster data transfer
  - Crazy stuff like https://github.com/JuliaPluto/Malt.jl/pull/46

This PR replaces Distributed with [Malt.jl](https://github.com/JuliaPluto/Malt.jl) to add support for Distributed itself. Fix #300

# Test release

After this PR is merged, we have a testing period where Pluto still uses Distributed, but you can **enable Malt with**:

```julia
Pluto.run(workspace_use_distributed_stdlib=false)
```

----
This PR is part of GSoC 2022. Some info on that [here](https://github.com/savq/gsoc22-proposal).


Project.toml
```diff
@@ -8,14 +8,14 @@ version = "0.19.11"
...
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
-Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 FileWatching = "7b1f6079-737a-58dc-b8bc-7a2ca5c1b5ee"
 FuzzyCompletions = "fb4132e2-a121-4a70-b8a1-d5b831dcdcc2"
 HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
 HypertextLiteral = "ac1192a8-f4b3-4bfe-ba22-af5b92cd3ab2"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 MIMEs = "6c6e2e6c-3030-632d-7369-2d6c69616d65"
+Malt = "36869731-bdee-424d-aa32-cab38c994e3b"
```
rikhuijzer, line 18: Also needs a `compat` entry. Probably `Malt = "0.5"`.

src/evaluation/WorkspaceManager.jl
```diff
@@ -224,57 +203,52 @@ end
...
 # compiler options, it does not resolve paths for notebooks
 # compiler configurations passed to it should be resolved before this
-function create_workspaceprocess(;compiler_options=CompilerOptions())::Integer
-    # run on proc 1 in case Pluto is being used inside a notebook process
-    # Workaround for "only process 1 can add/remove workers"
-    pid = Distributed.remotecall_eval(Main, 1, quote
-        $(Distributed_expr).addprocs(1; exeflags=$(_convert_to_flags(compiler_options))) |> first
-    end)
+function create_workspaceprocess(;compiler_options=CompilerOptions())::Malt.Worker
+    worker = Malt.Worker(;exeflags=_convert_to_flags(compiler_options)) # FIXME(savq): exeflags
```
rikhuijzer, line 237:
```suggestion
    worker = Malt.Worker(;exeflags=_convert_to_flags(compiler_options)) # TODO(savq): exeflags
```
Let's stick to TODO? Otherwise, people have to search for `TODO`, `FIXME`, and possibly other comments to find tasks. (Even better would be to track the task in an issue for better visibility.)

savq: Sure. I'll make sure to change the other comments.

You can even see that what this comment is referring to (the flags) was already fixed 🙈

rikhuijzer: No problem :+1: Happens to everyone

src/Configuration.jl
```diff
@@ -128,7 +128,6 @@ Note that Pluto is quickly evolving software, maintained by designers, educators
 end
 
 const RUN_NOTEBOOK_ON_LOAD_DEFAULT = true
-const WORKSPACE_USE_DISTRIBUTED_DEFAULT = true
```
fonsp, line 131: Can we continue supporting this feature? It has a lot of cool use cases, and it makes our tests faster!

disberd:
Would this change allow to interrupt cells on windows as well, or is that broken anyhow?

pankgeorg:
Tried to use this with `Dagger.jl` and `Distributed` and for some reason, the `addprocs()` gets stuck. In Malt.jl though, it does work. 👀

fonsp:
> Tried to use this with `Dagger.jl` and `Distributed` and for some reason, the `addprocs()` gets stuck. In Malt.jl though, it does work. 👀

This might have been because this PR won't enable the new Malt backend by default (first some testing period!) (although i temporarily enabled it by default in my previous commit to get CI to test Malt). Use `workspace_use_distributed_stdlib = false` to force using Malt.

fonsp:
😢 This PR is pretty much done but I'm getting strange out-of-memory errors on Windows, but I don't think it is actually running out of RAM, it seems more like a socket/libuv bug.

@pankgeorg or @Pangoraw could you check if this also reproduces locally? (check out the branch and run `test Pluto`) Any insights? It looks like this happens somewhere here:

https://github.com/savq/Pluto.jl/blob/c4bb0e2cfc7b701d466ead2e7cb20275a846d7a7/test/Configuration.jl#L60-L148

(uncomment all other tests in `runtests.jl` to make it faster)

It might be caused by this line, where the shutdown interrupt is triggered async without waiting for it to finish:

https://github.com/savq/Pluto.jl/blob/c4bb0e2cfc7b701d466ead2e7cb20275a846d7a7/test/Configuration.jl#L126

This might be fixed by replacing the `@async` with `try ... catch end` but it's strange that this was not happening before this PR.  (Because we launch the worker process in a way quite similar to Distributed.)

The CI logs also suggest that the worker did not write its port number to stdout, and some of the out-of-memory stack traces go through the `flush` call:

https://github.com/JuliaPluto/Malt.jl/blob/006c4f679fb4f0e822dd734e41aaf527b8469a4b/src/worker.jl#L26-L27

help!

Pangoraw:
Do we have `OPENBLAS_NUM_THREADS=1` (see https://github.com/JuliaLang/julia/pull/47803#issuecomment-1338062937) ?

The issue on VERSION >= 1.9 seems to be different (ref https://github.com/JuliaLang/julia/issues/50325).

Pangoraw:
I added a commit with `OPENBLAS_NUM_THREADS=1` to test, feel free to revert.

Pangoraw:
>  This is currently set on all distributed procs for every OS, right? Let's do the same by default on Malt directly?

yes, this is probably the right move. 

For nightly, we could test `1.10.0-beta2` instead? and for 1.6, I don't know...

fonsp:
Which 1.10 are we testing right now? Testing the beta sounds good!

I also made #2643 to avoid the interrupt call. This means avoiding the problem instead of fixing it, but I don't think that supporting `schedule(pluto_server_task, InterruptException())` on Windows is something we need to test. It's important that Ctrl+C in the terminal shuts down the server, but I think that we are simulating this wrong with `schedule`, so I don't really know how to test for that.

Pangoraw:
`nightly` is actually julia `master` (future 1.11), it would be nice if a docker tag for 1.10 existed so that we can get the alphas/betas & stable release for 1.10 automatically.

fonsp:
Still failing with out-of-memory on Windows Julia 1.6 :(

fonsp:
Maybe we just merge for testing, and eventually only enable Malt as default on Julia 1.8 and up?

fonsp:
Or @savq any ideas why we have this last failure with Malt julia 1.6 windows, but not with Distributed backend or another julia distro?

Pangoraw:
> @Pangoraw wdyt, any other ideas?

I don't have more ideas, we might have to test in real windows and see if we can spot the problem here.

pankgeorg:
> (btw some improvement to Crrl+C on windows in [JuliaLang/julia#51307](https://github.com/JuliaLang/julia/pull/51307))

I do not see an improvement, see https://julialang.slack.com/archives/C6SMTHQ3T/p1694888153763359?thread_ts=1694639864.397389&cid=C6SMTHQ3T (which contains that PR)

fonsp:
Heyy so I think I know what happened!

This initial PR removed the `workspace_use_distributed` option altogether, then later we added it to Malt (https://github.com/JuliaPluto/Malt.jl/pull/39) but some of the tests in the PR still had the `workspace_use_distributed=false` setting removed.

I fixed this in two recent commits: [`7b09e1d` (#2240)](https://github.com/fonsp/Pluto.jl/pull/2240/commits/7b09e1dc91901c8be2d8c6f967bd1ec2e52d950e) and [`a37fd31` (#2240)](https://github.com/fonsp/Pluto.jl/pull/2240/commits/a37fd31b6aafeeea25c3452f61f7c11c971cc967) and then suddenly the tests passed with Malt enabled! [`fd8c68a` (#2240)](https://github.com/fonsp/Pluto.jl/pull/2240/commits/fd8c68a7248a4c98224be1a6c5eebd476ba444ed)

I noticed that we were also getting OOM errors with the Distributed backend as default ([`27307e3` (#2240)](https://github.com/fonsp/Pluto.jl/pull/2240/commits/27307e303d86072703ed088792c404a3ea370a9d)) which means that it's not caused by Malt :)

Yayy so I will merge this with Malt still disabled by default, for the testing period, and then we can enable it by default soon (for all Julia versions)!

Pangoraw:
Awesome!! :tada: :tada:

Issue #2127, opened 2022-05-20
methods on custom structs defined inside macros are not deleted on re-run


disberd:
I noticed that when using macros to generate block of code that contains custom struct and custom methods defined on those, older methods are not deleted upon re-run.
This is not the case when the same codeblock is directly put in a cell rather than inside a macro.

This also creates another instance of the duplicated docstrings (if they are also present in the macro body) like in closed issues #1708 and #1932.

Tagging @Pangoraw directly as he is the macro wizard

Here is a video example:

https://user-images.githubusercontent.com/12846528/169555817-e1cdd435-92e2-4dac-9515-240077be3f74.mp4

and here is the notebook code:
<details>
<summary>Notebook source</summary>

```julia
### A Pluto.jl notebook ###
# v0.19.5

using Markdown
using InteractiveUtils

# ╔═╡ c4311933-7b49-4d3c-bca7-2adbf8a1a029
begin
	struct ASD end
	asdasd(::ASD) = "asd"
end

# ╔═╡ ad3fd6a4-d883-4329-8522-1b1f2304a4b3
methods(asdasd)

# ╔═╡ 142c7ad0-16d4-4af3-924c-a8f633c504a4
macro lol()
	quote
		struct LOL end
		lolol(::LOL) = "lol"
	end |> esc
end

# ╔═╡ 3b345bcb-46fa-4db1-89aa-cbd928c5fb82
@lol

# ╔═╡ 665302b7-5806-4c48-a307-dcebff544dd2
methods(lolol)

# ╔═╡ 00000000-0000-0000-0000-000000000001
PLUTO_PROJECT_TOML_CONTENTS = """
[deps]
"""

# ╔═╡ 00000000-0000-0000-0000-000000000002
PLUTO_MANIFEST_TOML_CONTENTS = """
# This file is machine-generated - editing it directly is not advised

julia_version = "1.7.1"
manifest_format = "2.0"

[deps]
"""

# ╔═╡ Cell order:
# ╠═c4311933-7b49-4d3c-bca7-2adbf8a1a029
# ╠═ad3fd6a4-d883-4329-8522-1b1f2304a4b3
# ╠═142c7ad0-16d4-4af3-924c-a8f633c504a4
# ╠═3b345bcb-46fa-4db1-89aa-cbd928c5fb82
# ╠═665302b7-5806-4c48-a307-dcebff544dd2
# ╟─00000000-0000-0000-0000-000000000001
[...]

PR #2241, opened 2022-08-07
Add macrocall cell_id to LineNumberNodes inside expanded expr

disberd:
In current Pluto the methods are deleted only if the `file` entry in the method signature is marked with the cell_id of the cell where that method is defined (by appending `#==#$(cell_id)` to the LineNumberNode). 
This is done to avoid deleting method signatures that were not defined within the notebook, see:
https://github.com/fonsp/Pluto.jl/blob/2158cab91132ae337e68fb0b708dd6cc46305b88/src/runner/PlutoRunner.jl#L739-L750

This breaks for methods defined inside the expression return by a macro, as the LineNumberNodes there refer to the file where the macro was defined, and no cell_id is attached there by Pluto, making deleting those methods impossible.

If the function being defined inside the macro does not depend on types defined within the macro itself, the problem is avoided as the new method definition simply overwrites the previous one. 
The problem arises when both a type and a method on that type is defined within a macro expression, see https://github.com/fonsp/Pluto.jl/issues/2127.

What this PR does, is going into the expanded macro expression inside `try_macroexpand` and appending the cell_id of the point where the macro is called to the LineNumberNodes.
To differentiate this from the standard cell_id insertion, the prefix I used for this procedure is `#@#==#` rather than `#==#`.

This seemed the *cleanest* approach to fix https://github.com/fonsp/Pluto.jl/issues/2127, and it provides information of the cell calling the macro which might be useful for other things.

I am open discuss alternative implementation if this is not considered acceptable. 

Closes https://github.com/fonsp/Pluto.jl/issues/2127

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -181,6 +181,17 @@ else
     m
 end
 replace_pluto_properties_in_expr(other; kwargs...) = other
+function replace_pluto_properties_in_expr(ln::LineNumberNode; cell_id, kwargs...) # See https://github.com/fonsp/Pluto.jl/pull/2241
+    file = string(ln.file)
+    out = if endswith(file, string(cell_id))
+        # We already have the correct cell_id in this LineNumberNode
+        ln
+    else
+        # We append to the LineNumberNode file #@#==# + cell_id
+        LineNumberNode(ln.line, Symbol(file * "#@#==#$(cell_id)"))
+    end
+    return out
+end
 
 "Similar to [`replace_pluto_properties_in_expr`](@ref), but just checks for existance and doesn't check for [`GiveMeCellID`](@ref)"
 has_hook_style_pluto_properties_in_expr(::GiveMeRerunCellFunction) = true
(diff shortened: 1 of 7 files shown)


Pangoraw:
Interesting approach! Are the error locations from macros still clickable when the error is thrown for a macro in the notebook ?

```julia
macro m()
    quote
        throw("boom")
    end
end

@m() # what does the stack trace looks like?
```

disberd:
@Pangoraw 
I don't seem to be able to get clickable errors from macros with the code you pasted even on latest Pluto:
image

If I put the the `throw` inside a function defined in the macro and call it still within the macro I get the jump-to links:
image
They did not work with the original PR but the latest push fixed that

src/runner/PlutoRunner.jl
```diff
@@ -253,6 +253,23 @@ module CantReturnInPluto
     replace_returns_with_error_in_interpolation(ex) = ex
 end
 
+# This is taken basically from fix_linenumbernodes! in Parse.jl
+function append_macrocellid_linenumbernodes!(ex::Expr, cell_id)
```
Pangoraw, line 257: This could be a method to `replace_pluto_properties_in_expr`

Pangoraw (commented):
>  They did not work with the original PR but the latest push fixed that

Awesome! Nice approach, this should also work for code with methods defined by Memoize.jl because the file ends with the cell_id. It could be great to add tests for this.

src/runner/PlutoRunner.jl
disberd: Nice suggestion!

Pangoraw:
Very cool! :confetti_ball:

PR #2244, opened 2022-08-13
Attach hljs to the window object to allow custom highlighting within code blocks

disberd:
Pluto already uses highlight.js to provide synthax highlighting within code_blocks. This very small changes just attaches the `hljs` object to the window in order to be able to register new languages to it.

The library comes with 36 languages [by default](https://highlightjs.org/download/), but many more can be easily added if we have access to the object

Here is a small example to enable Matlab highlighting:
image

I am not sure whether I attached it in the right part of the code in `Editor.js`, so that might be changed.

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -3,8 +3,6 @@ import * as preact from "../imports/Preact.js"
 import immer, { applyPatches, produceWithPatches } from "../imports/immer.js"
 import _ from "../imports/lodash.js"
 
-import hljs from "../imports/highlightjs.js"
-
 import { empty_notebook_state, set_disable_ui_css } from "../editor.js"
 import { create_pluto_connection } from "../common/PlutoConnection.js"
 import { init_feedback } from "../common/Feedback.js"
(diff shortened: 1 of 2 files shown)


Pangoraw:
I wonder if the assignment really need to be inside `componentDidMount` and can't just be in the main scope :thinking:

disberd:
@Pangoraw It seems not, it also works when moving the assignment to the constructor.

For reference to other potential users, here is the code snippet I put in my notebook with this PR to have the matlab highlighting, which can easily be adapted for other languages:
```html
html"""
<script>
	const {default: matlab} = await import('https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@11.6.0/build/es/languages/matlab.min.js')
	hljs.registerLanguage('matlab',matlab)

	const codeblocks = document.querySelectorAll('.language-matlab')
	for (const el of codeblocks) {
		hljs.highlightElement(el)
	}
</script>
<style>
	.hljs.language-matlab .hljs-comment {
		color: #028009;
	}
	.hljs.language-matlab .hljs-keyword {
		color: #0E00FF;
	}
	.hljs.language-matlab .hljs-string {
		color: #AA04F9;
	}
</style>
"""
```

The difference with the code on the first comment is adding a function to re-highlight blocks after the cell adding the language is run, as markdown is pre-rendered before cells are run, so that re-highlights the one of the added language.

pankgeorg (approved):

disberd:
@fonsp I just tried with new Pluto and I don't seem to have access to hljs from the frontend, do you know why?

fonsp:
It works for me!

disberd:
ok it indeeds worked after a restart, sorry for the bother!

Issue #2118, opened 2022-05-18
Use RegistryInstances.jl to avoid Pkg internal API


fonsp:
We currently use some internal Pkg API to make https://github.com/fonsp/Pluto.jl/wiki/%F0%9F%8E%81-Package-management work, we we should try to avoid as much as possible. Take a look at [src/packages/PkgCompat.jl](https://github.com/fonsp/Pluto.jl/blob/752973ff481568b29a88c60d80fce926a45340df/src/packages/PkgCompat.jl#L24-L42) to learn more, the compatibility issues are pretty well documented.

Some of the internal API usage can be avoid using https://github.com/GunnarFarneback/RegistryInstances.jl

For example: https://github.com/JuliaLang/Pkg.jl/issues/2607#issuecomment-1122481029

PR #2245, opened 2022-08-15
Advanced pkg manager

disberd:
This PR implements a modification to the integrated Package Manager that allows one to specify custom commands for adding packages (in a form of strings as you would type them in the julia PkgREPL), allowing people to have more control on the packages they add to a notebook, like:
- Adding un-registered packages
- Adding specific versions of packages
- Adding packages as developed ones, so that one can exploit Revise to track code changes from within the notebook

 I have wanted these features since a long time (together with multiple people based on discourse/zulip and github discussions) but was always scared away by the amount of understanding of Pluto internals both on the front-end and back-end I needed to implement this.
 As I had somehow more time in these past weeks I slowly tried to understand and implement a change to allow the points above while maintaining as much as possible the convenience of the Pluto PackageManager that we all love.

# Disclaimer regarding #2118 
I initially thought about starting from dealing with issue #2118 as a way to start dwelling into the PkgManager internals but at the time it looked like https://github.com/GunnarFarneback/RegistryInstances.jl was not registered yet.
Given that, it was a bit pointless to start an implementation as the API might have changed and Pluto could only depend on registered packages. When I decided to go for this implementation, I checked [this](https://github.com/GunnarFarneback/RegistryInstances.jl/issues/1) open issue where it seemed that registration didn't happen yet. I only discovered 2 days ago that the RegistryInstances had been registered since 2 weeks, but I was already quite advanced on this PR so I decided to skip #2118 for a later step.

# Some implementation details

I initially though about having a pop-up that would allow one to select an eventual version/repo/path with different fields, but I ended up resorting to a single pop-up that accepts a string that is then parsed to extract package details like the PkgREPL does.
I believe this solution would be more user friendly to users as people are more used to add packages from the PkgREPL.

This new PR internally uses the parsing commands from `Pkg.REPLMode` so that most of the work (with some additional Compat caveats due to changes in Pkg between 1.6 and 1.8) is done from Pkg itself.

In order to ease this, I created a new struct (`PkgData`) inside PkgCompat.jl that keeps useful information about a given installed package, like the `pkg_str` that is needed to parse the Pkg command, the finally installed version in the manifest as well as the function that is used to add the package to the environment (can be only `Pkg.add` or `Pkg.develop`) and the `Pkg.PackageSpec` that is actually used to specify the package details.

On the front-end, a new button has been added to the PkgPopup to allow specifying the custom string to be used using an HTML prompt (that defaults to `add $package_name`). The string is then sent to julia for validation/error_checking and the result is sent back to the notebook to assess whether the given string is acceptable or not. The validation step only looks at the string without trying to install the package immediately, so one could still provide a random github repository that does not contain a valid Julia package and the validation would still succeed at this step.
If the validation fails due to some ill-conditioned string, the prompt is re-issued forcing the user to change the string until a valid one is provided.

# Some usage examples

## Custom package commands

To add packages like you used to before, nothing changes. Instead, if you want to provide custom Pkg commands, you can do so using the newly added button:

https://user-images.githubusercontent.com/12846528/184658981-0ee355a2-b169-4344-a64a-143ff6af3fc4.mp4

You noticed that illegal commands re-prompts the user to fix the problem and try to provide some hints on it.
You also see that as soon as a valid custom string was provided, the PkgStatus marker became orange (randomly selected color), and the finally installed version is 0.4.4, satisfying the constraint provided with the `0.4` version specifier in the command.
The marker remains orange for all packages with a custom command string in order to recognize those more easily.

## Developing a package

One can also specify a command to develop a specific path:

https://user-images.githubusercontent.com/12846528/184660732-3932d8df-20fc-4005-a86f-931de1a0207c.mp4

The color for dev'd path is purple to further differentiate with just added packages.

# Conclusions

This PR still needs some modifications, but it is usable from my small testing in the current state so I would really love feedback and tests/comments from other people.

## TODO
- [ ] add Tests
- [ ] find corner-case things to fix
- [ ] Polish the front-end stuff (Here help/feedback from @fonsp, @pankgeorg, @Pangoraw would be helpful)
- [ ] Understand whether `precompile` statements should be added and where (@rikhuijzer)

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -170,13 +170,20 @@ const first_true_key = (obj) => {
  */
 
 /**
+ * @typedef PkgData
+ * @property {string} installed_version
+ * @property {string} pkg_str The equivalent PkgREPL string to be used to install the package
+ * @property {boolean} has_custom_pkg_str True if the package in question has a non-standard pkg_str, where standard is `add $(package_name)`
+ * @property {boolean} is_dev True if the package has en added with Pkg.develop rather than Pkg.add
+ */
+
+/**
  * @typedef NotebookPkgData
  * @type {{
  *  enabled: boolean,
  *  restart_recommended_msg: string?,
  *  restart_required_msg: string?,
- *  installed_versions: { [pkg_name: string]: string },
- *  installed_pkgstrs: { [pkg_name: string]: string },
+ *  installed_packages: { [pkg_name: string]: PkgData },
  *  terminal_outputs: { [pkg_name: string]: string },
  *  busy_packages: string[],
  *  instantiated: boolean,
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -625,7 +625,7 @@ export class Editor extends Component {
             /** This actions avoids pushing selected cells all the way down, which is too heavy to handle! */
             get_selected_cells: (cell_id, /** @type {boolean} */ allow_other_selected_cells) =>
                 allow_other_selected_cells ? this.state.selected_cells : [cell_id],
-            get_avaible_versions: async ({ package_name, notebook_id }) => {
+            get_available_versions: async ({ package_name, notebook_id }) => {
                 const { message } = await this.client.send("nbpkg_available_versions", { package_name: package_name }, { notebook_id: notebook_id })
                 return message.versions
             },
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -176,6 +176,7 @@ const first_true_key = (obj) => {
  *  restart_recommended_msg: string?,
  *  restart_required_msg: string?,
  *  installed_versions: { [pkg_name: string]: string },
+ *  installed_pkgstrs: { [pkg_name: string]: string },
  *  terminal_outputs: { [pkg_name: string]: string },
  *  busy_packages: string[],
  *  instantiated: boolean,
--- a/frontend/components/PkgStatusMark.js
+++ b/frontend/components/PkgStatusMark.js
@@ -110,7 +110,7 @@ export const PkgStatusMark = ({ package_name, pluto_actions, notebook_id, nbpkg 
     const [available_versions, set_available_versions] = useState(/** @type {string[]?} */ (null))
 
     useEffect(() => {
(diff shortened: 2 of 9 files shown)


greimel:
Something that would be nice to have is relative paths. Right now you cannot really use them, because Pluto stores the tomls in some temporary folder.

e.g.
```
dev ../MyPkg
```

doesn't work.

disberd:
Thanks for the feedback.
I'd have to check the feasibility of this now that I am back after some longish holiday :D

pawbz:
Hi there, it would be awesome to have this feature, I understand that we can start using unregistered packages!! 
Please let us know of the status...
Great work ~~

disberd:
Thanks,
Yeah this has been parked for a while. Next time I manage to speak to @fonsp I'll try to see if it can be pushed forward or it's better to wait for some rewriting of the internals of the pkg manager

disberd:
Closing this as we actually have a nicer implementation now from https://github.com/fonsp/Pluto.jl/pull/3218
Thanks @fonsp for this much wished for feature!

fonsp:
Awesome! @disberd Have you already tried the new project.toml feature with sources, and what do you think?

disberd:
@fonsp I just played around with it very shortly yesterday and it seems to achieve whatever I wanted from this, which was to being able to pin versions and use unregistered dependencies and still allow updating other dependencies in the notebook without completly overwriting the Project.

I didn't try the full `update` button and only updated dependencies via the new interface directly (that may still overwrite the custom project? Didn't check)

fonsp:
In theory, the "update" button will:
1. Remove all "auto" compat entries from the Project.toml, i.e. compat entries that match `r"^~\d+\.\d+\.\d+$"`, like `"~1.2.3"`. This is `PkgCompat.clear_auto_compat_entries!`.
2. Call `Pkg.update(; level=MAJOR)`
3. Auto-add compat entries for packages that have none.

So if you pinned a package using a compat entry `=2.3.4`, then it should not be affected by the update button.

fonsp:
In the future, I'd like to either remove the update button, or implement it as a thin layer around the Project.toml editor.

Maybe it could be a button inside the Proejct.toml editor UI.

This means we can remove some things from the Julia codebase :)

Issue #1999, opened 2022-03-21
More comfortable console output


j-fu:
Hi,
I appreciate the possibilty to print output to the console. 

- In the moment it is "dangerous": the "window" can become quite large. Would it be possible to have a similar "terminal" as with PlutoUI with scrollbars (I could try a PR in 1-2 weeks)
- Is there an option to turn on the old behavior ? E.g. in a cell menu and stored in the new cell metadata

fonsp:
1. Yes! Should be possible with a `max-width: ...` somewhere
2. With the old behaviour, do you mean not capturing stdout at all?

j-fu:
Ok will see if I find time for the PR during the weekend.

For the "old behavior":  probably yes, just seeing stdout in the console. I think this is also a bit related to #1988 .

this-josh:
I'm not sure if this is worth opening its own ticket as it seems to fall under "more comfortable console output". On my M1 mac the console background moves when my cursor hovers over it. Note this occurs with any number of lines, I just used two to make it easier to see.

Versions tested:

- Pluto v0.18.4, Julia  1.8.0-beta3+0~aarch64
- Pluto v0.18.4, Julia  1.7.2+0~aarch64
- Pluto v0.18.4, Julia  1.7.2+0~x64
- Pluto v0.19.0, Julia  1.7.2+0~x64

It's clearly not a big deal, but it's probably worth noting. 

https://user-images.githubusercontent.com/59419126/162152444-1dc4c4f6-4d3f-4c85-ab7f-64c1bc040cb6.mov

disberd:
Regarding 1), a quick hack to avoid the stdout getting huge is easy to get with CSS by adding a cell with this code:
```
html"""
<style>
	pluto-log-dot.Stdout {
		max-height: 300px;
		overflow: auto;
	}
</style>
"""
```

![image](https://user-images.githubusercontent.com/12846528/180982408-b1ff1e82-e9ac-4a7d-9c99-b730866aa773.png)

This has the problem that the retro-console effect is also present on the scrollbar. Avoiding this needs some more investigation/tweaking.

disberd:
> I'm not sure if this is worth opening its own ticket as it seems to fall under "more comfortable console output". On my M1 mac the console background moves when my cursor hovers over it. Note this occurs with any number of lines, I just used two to make it easier to see.
> 
> Versions tested:
> 
> * Pluto v0.18.4, Julia  1.8.0-beta3+0~aarch64
[...]

PR #2247, opened 2022-08-15
Limit height and fix width of stdout "terminal"

j-fu:
Prevent "terminal" output from being long, see #1999

For a test notebook see https://gist.github.com/j-fu/ee7f295de2c665a8b1f145933651cfda


fonsp:
This has been implemented in the meantime

Issue #2196, opened 2022-06-30
Logs messages inside macros go to last cell in the notebook


disberd:
While doing some weird macro stuff, I noticed that logs messages generated inside a macro function body can be sent as logs to cells below the one calling the macro.

Here is a small example:
![269da984-7543-41f5-8b98-a4413d491749](https://user-images.githubusercontent.com/12846528/176722013-5a8f8048-63fe-40ec-b71d-92733ff61d78.gif)

Edit: I tried adding other cells below and the logs always goes to the last cell in the notebook

PR #2249, opened 2022-08-19
Logging: Scope loggers to cell

### LINKED ISSUE #2121: @async run(`sleep 3`) is not async

Pangoraw:
Alternate implementation based on #2216. This allows to use the same logger in an asynchronous manner. The logs
will be displayed under the right cell automatically because we can keep
the writing task for longer. This fixes #2121.

async_io

### TODO:
 - [x] Logs from macro expansion are dropped when running the cell (https://github.com/fonsp/Pluto.jl/blob/338ca1952485190b315680f0fe70faf2b99adc07/src/evaluation/Run.jl#L164). Find a way to keep them around.


disberd:
Just checked this out to see if by chance it would also solve #2196 and I noticed that logs generated inside the macro body (so evaluated at _compile_ time) are completely lost, they are not attached to any cell anymore and they do not appear in the julia session that started Pluto either.
Printing to stdout within the macro body is sent to the julia session where Pluto was started, so the complete loss of logs only happens for stuff like `@info`.

fonsp:
@Pangoraw could you update the TODO list? Is the global logger issue the only thing left before merging?

Pangoraw:
> @Pangoraw could you update the TODO list? 

The most problematic is point raised by @disberd where macro expansion logs are dropped when the cell is run, making them impossible to read.

> Is the global logger issue the only thing left before merging?

From my understanding, we can safely remove the global logger since there should be only a few non-trivial way to escape local cell loggers anyway. What do you think?

Pangoraw:
I added log capture during expansion + replay when the cell is finally executed.

fonsp:
Awesome work Paul! I will take a look soon 🎉

src/runner/PlutoRunner.jl
```diff
@@ -2282,8 +2333,31 @@ function with_io_to_logs(f::Function; enabled::Bool=true, loglevel::Logging.LogL
...
+    function send_output()
+        output_str = String(take!(output))
+        if !isempty(output_str)
+            Logging.@logmsg loglevel output_str
+        end
+    end
+    @async begin
+        pipe_reader = Base.pipe_reader(pipe)
+        try
+            while !eof(pipe_reader)
```
fonsp, line 2347: I expected this short `while` loop to consume 100% CPU, but it doesn't. Nice! Do you know why?

fonsp: Is `readavailable` blocking?

src/runner/PlutoRunner.jl
```diff
@@ -2203,55 +2228,81 @@ const original_stderr = stderr
...
+    stream
+    logger::PlutoCellLogger
+    logs::Vector{Any}
+end
+
+Logging.shouldlog(cl::CaptureLogger, args...) = Logging.shouldlog(cl.logger, args...)
+Logging.min_enabled_level(::CaptureLogger) = min(Logging.Debug, stdout_log_level)
+Logging.catch_exceptions(::CaptureLogger) = Logging.catch_exceptions(cl.logger)
+function Logging.handle_message(pl::CaptureLogger, level, msg, _module, group, id, file, line; kwargs...)
+    Logging.handle_message(pl.logger, level, msg, _module, group, id, file, line; kwargs...)
```
fonsp, line 2252: what is this line for?

fonsp: It looks like it creates this behaviour. It happens very fast, so I made a video and screenshotted three consecutive frames:

Scherm­afbeelding 2022-11-09 om 13 20 45

Scherm­afbeelding 2022-11-09 om 13 20 59

Scherm­afbeelding 2022-11-09 om 13 21 06

fonsp (approved):

src/runner/PlutoRunner.jl
Pangoraw: From my understanding this is blocking until at least one byte is available (with `wait_readnb(this,1)`):
```julia
function readavailable(this::LibuvStream)
    wait_readnb(this, 1) # unlike the other `read` family of functions, this one doesn't guarantee error reporting
    iolock_begin()
    buf = this.buffer
    @assert buf.seekable == false
    bytes = take!(buf)
    iolock_end()
    return bytes
end
```

PR #2251, opened 2022-08-23
:sparkle: Isolated Cell Layout builder :sparkle: 

ghaetinger:
This PR adds a small checkbox button on the side of every cell that contains an output. When selected, the cell will be flagged as `isolated`. All of the cells marked as such will be layed out in an Isolated Cell Layout when a button in the export menu is clicked.

isolated

Selections are saved much like code folding, but for means of practicality, I added an identifier to the right side of the UUID, which I fell like is a big no-no.

### DIFF
--- a/src/notebook/Cell.jl
+++ b/src/notebook/Cell.jl
@@ -40,6 +40,7 @@ Base.@kwdef mutable struct Cell
 
     code::String=""
     code_folded::Bool=false
+    isolated::Bool=false
     
     output::CellOutput=CellOutput()
     queued::Bool=false
@@ -67,10 +68,11 @@ Cell(code) = Cell(uuid1(), code)
 cell_id(cell::Cell) = cell.cell_id
 
 function Base.convert(::Type{Cell}, cell::Dict)
-	Cell(
+    Cell(
         cell_id=UUID(cell["cell_id"]),
         code=cell["code"],
         code_folded=cell["code_folded"],
+        isolated=cell["isolated"],
         metadata=cell["metadata"],
     )
 end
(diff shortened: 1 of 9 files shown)


src/notebook/Cell.jl
```diff
@@ -40,6 +40,7 @@ Base.@kwdef mutable struct Cell
 
     code::String=""
     code_folded::Bool=false
+    isolated::Bool=false
```
disberd, line 43: This is something I did comment about recently here:
https://github.com/fonsp/Pluto.jl/pull/2239#issuecomment-1219630198

It would be nice to have a Set of option/statuses instead of the huge amount of flags we currently have (and we will have to add as new functionality is added like in this PR)

src/webserver/Firebasey.jl
```diff
@@ -638,6 +638,9 @@ function direct_diff(old::Cell, new::Cell)
 	if old.folded ≠ new.folded
 		push!(changes, ReplacePatch([:folded], new.folded))
 	end
+	if old.isolated ≠ new.isolated
+		push!(changes, ReplacePatch([:isolated], new.isolated))
+	end
```
disberd, line 643: This is actually skipped as script in the notebook, so it's not (I guess) used within Pluto but just for testing in the Firebasey notebook. Correct?

src/notebook/saving and loading.jl
```diff
@@ -110,7 +112,7 @@ function save_notebook(io::IO, notebook::Notebook)
     println(io, _cell_id_delimiter, "Cell order:")
     for c in notebook.cells
         delim = c.code_folded ? _order_delimiter_folded : _order_delimiter
-        println(io, delim, string(c.cell_id))
+        println(io, delim, string(c.cell_id) * (c.isolated ? _isolated_suffix : ""))
```
disberd, line 115: If you use metadata for the isolated cell you can also probably avoid this additional delimiter for identifying isolated cells

PR #2267, opened 2022-09-06
Show "Estimate runtime" in static HTML

fonsp:
This is the recorded Pkg install time (from #2265), plus the sum of all cell runtimes, plus 10 times 1.5.

image


fonsp:
I wrote a script that measures the install time, precompile time and loading time for each of the 2000 most downloaded packages:

https://github.com/fonsp/package-loading-times

This can be used for a better estimation of package loading times!

PR #2268, opened 2022-09-07
Allow solving multiple definitions by disabling one

fonsp:
Split out of #1859


fonsp:
It looks like all disabled assigners are also running:

https://user-images.githubusercontent.com/6933510/189002995-86868367-4646-424c-9ee2-98929a7a3706.mov

fonsp:
Superseeded by #2303

PR #2273, opened 2022-09-07
Automatically disable older cells to solve multiple defs

fonsp:
This PR adds some extra code into `update_save_run!` that checks, by comparing the old and new `Topology`s:
1. Did any of the `cells` (i.e. cells that the user wants to run) define a variable that was defined a fewer number of times in the old `Topology`?
2. For each such `cell`:
  a. Does `cell` define exactly one variable?
  b. Are all definers of that variable not in `cells`? i.e. did the other definitions appear in a previous run?
  c. Are no function definitions involved?
  d. Are there no macro calls? (TODO)
  e. Does `cell` not reference itself? (e.g. `x = x+1`)
  f. Then: disable all other definers of the variable.

This is done fully in backend, which makes it accurate!

# Demo

This video demonstrates the basic functionality, and then a series of situations where we purposefully do not use this new feature.

https://user-images.githubusercontent.com/6933510/194586555-a62e5f78-b6f3-47cd-94dd-5923d14dac5a.mov

# Previous/related work
- #2270
- #1859
- #2268
- #2303

# TODO
- [x] #2270
- [x] #2303 
- [ ] Check if this is not too aggressive
- [x] Add GUI to explain (what will happen and) what just happened
- [x] Maybe make the criterea even more strict
- [x] What about `x = x + 1`?
- [x] Macro calls


fonsp:
Now with GUI!

https://user-images.githubusercontent.com/6933510/195908365-d34d0bb2-4102-4962-a004-a06325244d41.mov

src/evaluation/Run.jl
```diff
@@ -507,10 +571,10 @@ end
...
     skipped_cells = filter(is_skipped_as_script, notebook.cells)
     indirectly_skipped = collect(topological_order(topology, skipped_cells))
-    for cell in indirectly_skipped
-        cell.depends_on_skipped_cells = true
-    end
-	for cell in setdiff(notebook.cells, indirectly_skipped)
+    for cell in notebook.cells
         cell.depends_on_skipped_cells = false
     end
+	for cell in indirectly_skipped
```
rikhuijzer, line 577:
```suggestion
    for cell in indirectly_skipped
```

src/evaluation/Run.jl
```diff
@@ -334,10 +330,30 @@ function update_save_run!(
...
 	run_async::Bool=false, 
 	prerender_text::Bool=false, 
+	auto_solve_multiple_defs::Bool=false,
+	on_auto_solve_multiple_defs::Union{Nothing,Function}=nothing,
 	kwargs...
 )
 	old = notebook.topology
 	new = notebook.topology = updated_topology(old, notebook, cells) # macros are not yet resolved
+	
+	# _assume `auto_solve_multiple_defs == false` if you want to skip some details_
```
rikhuijzer, line 340:
```suggestion
```

src/evaluation/Run.jl
```diff
@@ -406,6 +422,54 @@ end
...
+					node = new.nodes[c]
+					
+					# the cell defines only one variable. for more than one, we might confuse the user, or disable more things than we want to.
+					length(node.definitions) == 1 &&
+					# avoid something like `x = x + 1`
+					disjoint(node.references, node.definitions) &&
+					# no function definitions
+					isempty(node.funcdefs_without_signatures) &&
+					# no macro calls					
+					isempty(node.macrocalls)
```
rikhuijzer, line 454:
```suggestion
					# for more than one, we might confuse the user, or disable more things than we want to.
					only_one_definition = length(node.definitions) == 1
					no_self_reference = disjoint(node.references, node.definitions)
					no_function_defs = isempty(node.funcdefs_without_signatures)			
					no_macro_calls = isempty(node.macrocalls)
					return only_one_definition && no_self_reference && no_function_defs && no_macro_calls
```
Code is generally safer than comments because comments tend to become outdated and are often not read

fonsp: I went for a middle ground between the two, and I also removed the confusing `disjoint(cells, other_definers) &&` and just used two `if` statements. Thanks!

PR #2282, opened 2022-09-14
Add query suggestions in live docs

Pangoraw:
live_docs_suggestions

### DIFF
--- a/src/webserver/REPLTools.jl
+++ b/src/webserver/REPLTools.jl
@@ -117,9 +117,9 @@ responses[:docs] = function response_docs(🙋::ClientRequest)
         query = "@$(query[begin:end-1])_str"
     end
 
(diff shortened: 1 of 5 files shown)


src/webserver/REPLTools.jl
```diff
@@ -117,9 +117,9 @@ responses[:docs] = function response_docs(🙋::ClientRequest)
         query = "@$(query[begin:end-1])_str"
     end
 
-    doc_html, status = if REPL.lookup_doc(Symbol(query)) isa Markdown.MD
+    doc_html, status = if (doc_md = Docs.doc(Docs.Binding(Base, Symbol(query)))) isa Markdown.MD &&
```
Pangoraw, line 120: This local fetcher was never actually working (it is now), so maybe we can just get rid of it ?

PR #2284, opened 2022-09-15
frontend bundler: Replace search param with hash in filename

Pangoraw:
This replace the weird URL encoded filename in https://github.com/fonsp/Pluto.jl/tree/fec115f5d55d427b34c1b6dd5ea43fbf6cbc449f/frontend-dist
with a small hash which should not contain weird characters. This way we can still URI decode component in the path.

This should fix https://github.com/JuliaPluto/PlutoSliderServer.jl/issues/88


fonsp (approved):
Yesss well found Paul!!

Pangoraw:
The actual fix for the bug is https://github.com/JuliaPluto/PlutoSliderServer.jl/pull/91.

Issue #65, opened 2020-04-07
Keyboard only use


fonsp:
|  | Which keys? | Action | 
|----------------|---------------|---------------|
| ✅ | `F1` or `Ctrl+Shift+?` | List shortcuts |
| ✅ | `Shift+Enter` | Run cell |
| ✅ | `Ctrl+Enter` | Run cell _if the code changed_ and add a new cell below |
| ✅ | `Ctrl+S` | Run all cells with modified code _in a single batch_ |
| ✅ | `Tab` | Autocomplete |
| ✅ | `Shift+Delete` | Delete cell |
| ✅ | `Page+Up` and `Page+Down` | Edit cell above/below |
| ✅ | `Ctrl+Q` | Interrupt all cells |
| ✅ | `Ctrl+P` | Export to PDF |
| ⚪ | `Ctrl+Z` but it's tricky | Undo delete cell |
| ⚪ | `` | Fold/unfold cell |
| ⚪ | `Ctrl+Up` and `Ctrl+Down` | Move cell(s) up/down |
| ⚪ | `Ctrl+Shift+S` | Move notebook file |
| ⚪ | `` | Wrap in `begin ... end` |
| ⚪ | `` | Wrap in `let ... end` |
| ⚪ | `Ctrl/Cmd+C` and `Ctrl/Cmd+V` | Cut, copy & paste cell(s) |
| ⚪ | `` | Add cell without running |
| ⚪ | `` |  |

fonsp:
Same shortcuts as Jupyter wherever possible.

nilshg:
Just stumbled upon this as I was looking for some place where the available keyboard shortcuts are listed - does this exist? 

One of the nice things about Jupyter imho is the ability to switch between command and cell mode by doing Esc/Enter, and then simple shortcuts for creating/deleting cells, moving cells etc in command mode. Would it be an option to have this in Pluto?

fonsp:
Hey! The full list is in the [FAQ](https://github.com/fonsp/Pluto.jl/wiki), you can also press `F1` inside the editor. If there is anything missing, I will happily add it to the list above!

I am [very reluctant to add a second editor mode](https://dl.acm.org/doi/pdf/10.1145/2212877.2212896) 😬. Part of Pluto's UI design is having all primary functionality (edit, run, add, delete, move and hide) directly available, at the position on your screen where it takes effect.

nilshg:
[...]

PR #2287, opened 2022-09-16
Set keyboard shortcuts for adding a cell below or above selected cell

dgsob:
The default keys for this in Jupyter's command mode are "A" and "B" for cells above and below. Here I set it to be Alt+A and Alt+B (as in #65 the shortcut is not specified).

btw before/after - above/below is a little unfortunate


fonsp:
What are the keyboard shortcuts for VS Code? We try to match the VS Code shortcuts, not necessarily Jupyter.

damiansobecki:
> We try to match the VS Code shortcuts, not necessarily Jupyter.

I know but adding cells seems to be notebook specific action, I'm not aware of any VS Code notebooks.
Do you mean this? https://code.visualstudio.com/docs/datascience/jupyter-notebooks 
If so, the shortcuts are the same as in Jupyter's notebook. If this is not what you meant, sorry, I'm not familiar with VS Code.

fonsp:
Exactly! VS Code has a notebook UI, the shortcuts are different.

damiansobecki:
OK, nice. I meant this two specific shortcuts: "When a code cell is in command mode, the A key can be used to add a cell above and the B can be used to add a cell below the selected cell." That's from their doc I linked.

xgdgsc:
From a previous jupyterlab user this is the first thing you would notice when switching to Pluto. Is there anything blocking this?

dralletje:
VSCode uses <kbd>ctrl</kbd>+<kbd>;</kbd> <kbd>A</kbd> and <kbd>ctrl</kbd>+<kbd>;</kbd> <kbd>B</kbd>.
I guess they use <kbd>ctrl</kbd>+<kbd>;</kbd> as a "enter command mode" thingy.

fonsp:
Right, and we don't want to have a "command mode" or "cell mode" for Pluto shortcuts. @dralletje is working on a way to create new cells with keyboard interactions only, by treating 3 newlines as a cell break. I think we should go with that instead!

PR #2289, opened 2022-09-19
Using RegistryInstances.jl to avoid some Pkg internal API #2118

kcin96:
This pull request attempts to address issue ( #2118 ) by using RegistryInstances.jl. Some of the internal API has been modified to utilize public API functions from [https://github.com/GunnarFarneback/RegistryInstances.jl](https://github.com/GunnarFarneback/RegistryInstances.jl).


src/packages/PkgCompat.jl
```diff
@@ -159,7 +159,7 @@ end
 # (⛔️ Internal API)
 _get_registries() = map(_get_registry_paths()) do r
```
fonsp, line 160: Here, the entire function `_get_registries` can be cleaned up, because `RegistryInstances` means that we no longer need to use the `@static if` with `isdefined` to find out where the API is. It should look something like:

```julia
_get_registries() = map(_get_registry_paths()) do r
	RegistryInstances.RegistryInstance(r)
end
```

Same for `_get_registry_paths`

fonsp (changes requested):

src/packages/PkgCompat.jl
fonsp: We can now also update the comments with ⛔️

kcin96: Good point! It looks much cleaner now!

```julia
# (✅ "Public" API using RegistryInstances)
"Return paths to all installed registries."
_get_registry_paths() = [s.path for s in RegistryInstances.reachable_registries()]

# (✅ "Public" API using RegistryInstances)
_get_registries() = map(_get_registry_paths()) do r
	RegistryInstances.RegistryInstance(r)
end

# (✅ "Public" API)
"Contains all registries as `Pkg.Types.Registry` structs."
const _parsed_registries = Ref(_get_registries())

# (✅ "Public" API)
"Re-parse the installed registries from disk."
function refresh_registry_cache()
	_parsed_registries[] = _get_registries()
end
```

fonsp: Awesome!

fonsp (approved):

kcin96:
No problem!  I'll have a look around the issues tab 😄

PR #2162, opened 2022-06-07
Make publish_to_js official API through AbstractPlutoDingetjes

fonsp:
See https://github.com/JuliaPluto/AbstractPlutoDingetjes.jl/pull/8

TODO:
- [ ] test... somehow. maybe the APD side needs to be merged first?
- [ ] take #2092 into account

Issue #2092, opened 2022-05-04
`publish_to_js` mostly breaks if used inside logs


disberd:
I found this issue while trying to send PlutoPlotly plots into logs via the `@info` macro.
It seems that putting object with `publish_to_js` inside logs mostly breaks.

For scalars, `publish_to_js` sometimes works without error and sometimes it gives a javascript error interrupting execution of the rest of the (JS) code.

When trying to publish more complex objects (like arrays or dicts), it seems to always fail.

Here is a video example (first two `@info` cells did error at first run, a different set of them also errored upon restart of the notebook)

https://user-images.githubusercontent.com/12846528/166691674-927070b4-a6ea-4beb-b4fe-27a9625f3fb7.mp4

and here is the notebook gist:

https://gist.github.com/disberd/3fba685fbb9b6a8478fc1875ef64aa11

disberd:
So I did a bit of further investigation and it seems to be a problem of timing of access to the published objects.

The objects are correctly published to the JS side and can be found within the `editor_state` after cell execution. They are probably just tried to be accessed slightly before they are published for complex objects when used inside logs.

PR #2297, opened 2022-09-27
🍔 Logs: Vertical layout

fonsp:
Here is a much simpler design for the logs 🙃

The exciting ideas in #437 are not really working out: the horizontal timeline was too difficult to develop on my own. The main difficulty is that logs can vary greatly in size, and there are lots of tricky edge cases. The horizontal layout also needed an alternative for smaller screens, which was the vertical layout that we have before this PR. Difficult difficult!

This means that this PR loses the ability to see from which line a log originates, we need to add that back somehow...

Schermafbeelding 2022-09-27 om 15 08 52

Schermafbeelding 2022-09-27 om 15 09 07

Still better than a terminal :)

https://user-images.githubusercontent.com/6933510/192535186-02f25fb9-72bf-4e8e-a0c0-c65f3f8a1b08.mov

Schermafbeelding 2022-09-27 om 15 10 36
Schermafbeelding 2022-09-27 om 15 10 57
Schermafbeelding 2022-09-27 om 15 11 06

# TODO:
- [x] Disable lazy loading...
- [x] I also changed the `! info` markdown callout colors by accident

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -585,6 +585,7 @@ export const CellInput = ({
                     highlighted_line_compartment,
                     global_definitions_compartment,
                     editable_compartment,
+                    highlightLinePlugin(),
 
                     // This is waaaay in front of the keys it is supposed to override,
                     // Which is necessary because it needs to run before *any* keymap,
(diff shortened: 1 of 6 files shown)


disberd:
Cool!
Since you are re-going into the logs code, can you have a quick look at whether this issue #2092 is fixable without too much effort?
Normal plots work in logs but not the ones from PlutoPlotly as they rely on `publish_to_js` which is broken with logs

fonsp:
@disberd I took a look, not really fixable within this PR but I added it to the TODO of #2162 👍

PR #2308, opened 2022-10-10
Fixes "UndefVarError: expr not defined" in replace_returns_with_error_in_interpolation

mickronome:
In PlutoRunner.jl replace_returns_with_error_in_interpolation parameter ex was referred to using both ex and expr.  Changes ex->expr to solve "UndefVarError: expr not defined"

This only fixes this typo caused by the minimal example below, but it seems to allow to run.
```julia
module WAT
	quote
		
		
	end 
end
```

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -254,9 +254,9 @@ module CantReturnInPluto
     replace_returns_with_error(other) = other
 
     "Go through a quoted expression and remove returns"
-    function replace_returns_with_error_in_interpolation(ex::Expr)
-        if ex.head == :$
-            Expr(:$, replace_returns_with_error_in_interpolation(ex.args[1]))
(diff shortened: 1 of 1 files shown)


fonsp:
Thanks! 

I don't completely get the example that you gave, is there something that did not work before, but it works after this PR? Where/when/how did you see the UndefVarError?

disberd:
It looks like the modified function was ill-written from the start:

https://github.com/fonsp/Pluto.jl/blob/36232b497678941f8d8845a4246da8ac281cb045/src/runner/PlutoRunner.jl#L257-L264

If you see on line 262 the function is accessing the variable `expr` while the same variable is called `ex` in the rest of the function.
Indeed when falling onto the `else` branch the code would create an Undefvar error

fonsp:
Alright, thanks everyone!!

PR #2318, opened 2022-10-16
Refactor WorkspaceManager

savq:
No functional changes here. Just a renamed variable and formatting.

- Rename `workspaces` to `active_workspaces`
- Prefer quote blocks to colon quote expressions
- Prefer if blocks to ternary operators
- Prefer function calls instead of pipes
- Minor changes to docstrings and comments
- Remove trailing whitespace

---

Creating this PR instead of wating on #2240 because rebasing was getting tedious.

### DIFF
--- a/src/evaluation/WorkspaceManager.jl
+++ b/src/evaluation/WorkspaceManager.jl
@@ -8,7 +8,11 @@ import ..Pluto.ExpressionExplorer: FunctionName
 import ..PlutoRunner
 import Distributed
 
-"Contains the Julia process (in the sense of `Distributed.addprocs`) to evaluate code in. Each notebook gets at most one `Workspace` at any time, but it can also have no `Workspace` (it cannot `eval` code in this case)."
+"""
+Contains the Julia process (in the sense of `Distributed.addprocs`) to evaluate code in.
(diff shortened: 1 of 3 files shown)


src/evaluation/WorkspaceManager.jl
```diff
@@ -22,37 +26,38 @@ Base.@kwdef mutable struct Workspace
     original_ACTIVE_PROJECT::Union{Nothing,String}=nothing
 end
 
+const SN = Tuple{ServerSession, Notebook}
+
 "These expressions get evaluated whenever a new `Workspace` process is created."
-process_preamble() = quote
```
fonsp, line 26: This needs to be a function because we call project_relative_path, which uses RelocatableFolders.jl, see #1852

fonsp:
Looks fine otherwise!

src/evaluation/WorkspaceManager.jl
savq: I missed the interpolation. I'll update.

fonsp:
force push means that i have to review the whole PR a second time 😢 but I guess it looks fine!

PR #2322, opened 2022-10-17
Print secret to console if main page requested without it

disberd:
This PR was motivated by this discussion on zulip: https://julialang.zulipchat.com/#narrow/stream/243342-pluto.2Ejl/topic/Lost.20the.20secret

I also found myself in a situation where long-running pluto session would result in my losing the secret and not being able to recover it easily.
This is at the moment a very bare-bone implementation, but it serves the purpose.
It is also a fast way to get this seen by @fonsp, @Pangoraw and @pankgeorg for comments.

### DIFF
--- a/src/webserver/Static.jl
+++ b/src/webserver/Static.jl
@@ -170,6 +170,7 @@ function http_router_for(session::ServerSession)
                 end
                 response
             else
+                println("A request to the Pluto server was made without providing a secret, here is the secret: $(session.secret)")
                 error_response(403, "Not yet authenticated", "<b>Open the link that was printed in the terminal where you launched Pluto.</b> It includes a <em>secret</em>, which is needed to access this server.<br><br>If you are running the server yourself and want to change this configuration, have a look at the keyword arguments to <em>Pluto.run</em>. <br><br>Please <a href='https://github.com/fonsp/Pluto.jl/issues'>report this error</a> if you did not expect it!")
             end
         end


src/webserver/Static.jl
```diff
@@ -170,6 +170,7 @@ function http_router_for(session::ServerSession)
                 end
                 response
             else
+                println("A request to the Pluto server was made without providing a secret, here is the secret: $(session.secret)")
```
pankgeorg, line 173: Let's use `@info` here as we mostly do
```suggestion
                @info("A request to the Pluto server was made without providing a secret, here is the secret: $(session.secret)")
```

pankgeorg (approved):
We can do that; it may spam too much of the user's STDOUT and it's maybe susceptible to DDOS but all of that should be handled elsewhere - I think this is the right-tech (not too low [save to file], not too high [sophisticated throttling or something]) solution for now.

fonsp:
I do think that we should do something to prevent DDOS, that's the original reason why I decided not to implement this feature yet. We could do something simple with throttling, maybe just the `throttle` function from our source code?

Also, just printing the secret is a bit technical (you would need to understand what the secret is, how URL queries work, how to add it to the URL). Maybe we could log the requested URL, plus secret, to the user? That way they can just click in the terminal.

disberd:
Maybe we can talk about it at the pluto meeting today to understand what I have to do

disberd:
Implemented throttling following @fonsp suggestion and also printing the full url to use for connecting.
How do we test this feature? I'd need some help/suggestion on that side

PR #2328, opened 2022-10-19
Allow copying vars from Memoize.jl to new workspace

### LINKED ISSUE #2305: `memoize_cache` from Memoize.jl does not work in Pluto
One can use the function `memoize_cache(::Function)` to see the list of cached entries for each function:

```julia
julia> @memoize f(a) = a+1
f (generic function with 2 methods)

julia> f(1)
2

julia> f(2)
3

julia> memoize_cache(f)
Dict{Any, Any} with 2 entries:
  (2,) => 3
  (1,) => 2
```

The same doesn't work in Pluto:
image

disberd:
This PR fixes #2305.

The issue is that when creating a new workspace and moving variables there, Pluto ignores all variables whose name starts with `#`.
Unfortunately Memoize.jl uses such a variable name to save the cache. This PR simply adds another check so that a variable whose name starts with `#` is actually copied over if it matches the name pattern of Memoize.jl

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -632,6 +632,8 @@ end
 # DELETING GLOBALS
 ###
 
+# This function checks whether the symbol provided to it represents a name of a memoized_cache variable from Memoize.jl 
+is_memoized_cache(s::Symbol) = startswith(string(s), "##") && endswidth(string(s), "_memoized_cache")
 
 function do_reimports(workspace_name, module_imports_to_move::Set{Expr})
     for expr in module_imports_to_move
@@ -698,7 +700,7 @@ function move_vars(
             end
         else
             # var will not be redefined in the new workspace, move it over
-            if !(symbol == :eval || symbol == :include || string(symbol)[1] == '#' || startswith(string(symbol), "workspace#"))
+            if !(symbol == :eval || symbol == :include || (string(symbol)[1] == '#' && !is_memoized_cache(symbol)) || startswith(string(symbol), "workspace#"))
                 try
                     getfield(old_workspace, symbol)
 


fonsp:
I think it's cleaner to add this workaround at the macro expansion level, that would also extend more generally to other packages that need special treatment... but @Pangoraw can decide what to do here

fonsp:
Can we fix Memoize.jl directly? Why do they not use a global cache, mapping functions to their memoize cache?

Pangoraw:
> I think it's cleaner to add this workaround at the macro expansion level

It seems that Memoize uses [eval](https://github.com/JuliaCollections/Memoize.jl/blob/beba3c54d91da761824ac0665a49ce276e68f9c3/src/Memoize.jl#L56-L58) to create its cache so there is nothing Pluto can do during expansion.

Pangoraw:
> I think it's cleaner to add this workaround at the macro expansion level, that would also extend more generally to other packages that need special treatment... but @Pangoraw can decide what to do here

We could maybe set a whitelist instead of a blacklist:
```julia
symbol ∈ (
    Symbol("#eval"),               # builtin
    Symbol("#include"),            # builtin
    Symbol("#__this_module_name"), # from https://github.com/fonsp/Pluto.jl/pull/1966
)
```

On the top of my head, those are the symbols we want to avoid moving, are there more?

src/runner/PlutoRunner.jl
```diff
@@ -698,7 +700,7 @@ function move_vars(
             end
         else
             # var will not be redefined in the new workspace, move it over
-            if !(symbol == :eval || symbol == :include || string(symbol)[1] == '#' || startswith(string(symbol), "workspace#"))
+            if !(symbol == :eval || symbol == :include || (string(symbol)[1] == '#' && !is_memoized_cache(symbol)) || startswith(string(symbol), "workspace#"))
```
dralletje, line 703: This double negation screws with my brain hahaha

fonsp:
@Pangoraw Good idea, let's do that!

I looked into it, I have no idea why I wrote it as `string(symbol)[1] == '#'`. I traced it back to [this commit](https://github.com/fonsp/Pluto.jl/commit/cd3dcdfcde4382439b5aec5f44addfe9fcbf0835#diff-19cd52ed7f505f98eb8f7ab7816e5292d106ef88a37b42748093d5aa76916a10) in Pluto's 3rd month, so it was probably not super motivated, a random guess based on seeing  `#eval` and `#include`.

fonsp:
Implemented the suggestion by @Pangoraw , it works now

Scherm­afbeelding 2022-11-09 om 15 51 27

fonsp (approved):
Merge after tests pass

Pangoraw:
hmm, it seems that Julia also populates the module with `var"#..."` so maybe it does not play nicely with redefinitions:
```julia
julia> f() = nothing
f (generic function with 1 method)

julia> names(@__MODULE__; all = true)
8-element Vector{Symbol}:
 Symbol("##meta#58")
 Symbol("#f")
 :Base
 :Core
 :InteractiveUtils
 :Main
 :ans
 :f

julia> var"#f"
typeof(f) (singleton type of function f, subtype of Function)

julia> var"#f" = :redefining_f
ERROR: invalid redefinition of constant #f
Stacktrace:
 [1] top-level scope
   @ REPL[4]:1
```
My proposal was unfortunately incorrect :grimacing:

dralletje:
🙃

PR #2341, opened 2022-10-25
Draft: Guard exceptions caused by `last(Tables.rows(table))`

pankgeorg:
Let's not merge this until we understand why `applicable(lastindex, true)` is true, but `last(table)` explodes.


pankgeorg:
(From the pluto call:) Let's write a test and put this in!

PR #2343, opened 2022-10-26
On using other languages indirectly, e.g. Clojure

PallHaraldsson:
[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 _Writing a notebook is not just about writing the final document — Pluto empowers the experiments and discoveries that are essential to getting there._
 
-**Explore models and share results** in a notebook that is
+**Explore models and share results** in a notebook that is:
 
 - **_reactive_** - when changing a function or variable, Pluto automatically updates all affected cells.
 - **_lightweight_** - Pluto is written in pure Julia and is easy to install.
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
I just found using Cljuto.jl intriguing, using Clojure[Script] with Pluto. I suppose this is trivia, but it might be good trivia to include an obscure/non-mainstream language, to show anything is really possible. When I used PythonCall. similar could be done, py"<Python code>", and I suppose also possible in same or similar way with PythonCall.jl.

dralletje:
I don't think focussing users on PythonCall and such is something we want right now...
We can in no way provide the experience we want in Python or Clojurescript.

PallHaraldsson:
What's the experience we (you) want for e.g. Python possibly? I realize Pluto is Julia-first, since it is Julia source code, so I don't think multi-language support can be done much differently (and that might be a perceived downside vs. alternatives Jupyter, Quatro, and if people just knew of the possibility, then not). Just close if you think this is too much trivia. I just found it a bit interesting that people wanted and did support Clojure in Pluto, already. Most wouldn't know if not documented, I even think Clojure isn't too easily supported in Julia except in Pluto!

I was just thinking documenting the possibility,* since it is a possibility (just as in Julia's REPL to work with Python), not thinking of this as implicit endorsement or for you to support/maintain in any way. The info needs not be front-and-center, if considered trivia, could be in FAQ, or IFAQ (InFrequently...).

[ * If the issue with PythonCall is the slow startup I hope if will catch up to PyCall, and thinking about it, maybe neither works, since you need libpython installed (which PyCall downloads for you, but I've not confirmed it actually works in Pluto, should it?). Cljuto.jl is a tiny package however with one macro should just work (points to online JavaScript code, i.e. ClojureScript). The package isn't registered (yet) however, so you need to know how to install it.]

Issue #2314, opened 2022-10-13
Variable assigned through splat never updates


yha:
For example, here `y` doesn't update, even when manually re-running the cell referring to it.

https://user-images.githubusercontent.com/4170948/195596667-17b22585-5163-45e3-9d77-2f6b9094bef4.mp4

Pangoraw:
Note the console warning:
```
┌ Warning: unknown use of `=`. Assignee is unrecognised.
│   ex = :(y...)
└ @ Pluto.ExpressionExplorer ~/.julia/packages/Pluto/rBF1P/src/analysis/ExpressionExplorer.jl:152
```

PR #2365, opened 2022-11-11
Add case for handling splat assignments.

kcin96:
This PR attempts to fix Issue #2314 by adding a case for handling splat assignments.

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -148,6 +148,10 @@ function get_assignees(ex::Expr)::FunctionName
         get_assignees(ex.args[1])
     elseif ex.head == :ref || ex.head == :(.)
         Symbol[]
+    elseif ex.head == :...
+        # Handles splat assignments. e.g. _, y... = 1:5
+        args = ex.args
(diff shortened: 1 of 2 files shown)


rikhuijzer (changes requested):
Can a test be added to avoid future problems?

kcin96:
Yes I can add a test.

kcin96:
Besides adding this test under test/ExpressionExplorer.jl

```julia
  @testset "Multiple assignments" begin
        #other tests ...
        @test testee(quote
            a, b... = 0:5
        end, [],[:a, :b], [[:(:)]], [])

  end
```

are there any other tests that you can think of that can be helpful? e.g. test for y... value assignment?

Pangoraw:
Thanks for tackling this one!
You can also test the case where this pattern is used as a function argument:
```julia
f((a,b...)) = b # only defines f, no reference to a or b
```

Pangoraw:
Thanks!! :confetti_ball: 

I added an extra test with destructuring an argument:
https://github.com/fonsp/Pluto.jl/blob/54ff21d960905faa3714aad461a070066b457aa8/test/ExpressionExplorer.jl#L319-L321

PR #2368, opened 2022-11-14
embed_display: await inner scripts to finish

fonsp:
Fix https://github.com/JuliaPluto/PlutoUI.jl/issues/236

After this PR, the `<script>` element that powers the `embed_display` will `await` all embedded `<script>` elements to finish.

# TODO:
- [ ] frontend test would be *great* but takes some effort
- [ ] I had to use a `requestIdleCallback` to keep this from getting crazy complicated, which might lead to slower rendering when using lots of embed_display in one cell... needs to be investigated.


fonsp:
Hmpf I tricked myself into thinking that this PR worked because of #1776 , it's not actually working yet.

PR #2377, opened 2022-11-17
Fix cell input scrollbars showing on Chrome Linux

### LINKED ISSUE #2373: Unnecessary vertical scrollbar appears in the input area (Chrome & Linux )
At some point (probably 2022-11-04), unnecessary vertical scrollbars began to appear in the input areas.
See the following screenshot. (The scrollbar will appear after the notebook is loaded and the input area is activated)

input_area_vertical_scrollbar

My environment is as follows:
* Pluto v0.19.15
* Linux (Xubuntu 22.04)
* Chrome (version 107.0.5304.87)

Firefox (version 107.0) does not have this problem.

This problem started happening when I updated Pluto from commit  eca8648f to commit 20b3287b. I thought it might be caused by this update, but reverting to commit eca8648f did not fix the problem.

I then updated to commit 4af5a569 and the problem is still there.
I have a feeling this issue is related to Issue #2363, but my problem was not fixed in commit c741b243.

It seems that chrome was updated to version 107 at the same time the problem started occurring, so chrome may be the cause.

I currently hide this scrollbar by adding the following line to `editor.css`.

```css
pluto-input .cm-editor .cm-scroller {
    overflow-x: visible;
}
```

I don't really understand the significance of this setting.
I checked with chrome's DevTools and found that the setting `"overflow-x: auto;"` on elements with `class="cm-scroller"` seemed to be causing this scrollbar to appear, so I added the above setting to disable it.

fonsp:
Fix #2373

I am _fairly_ sure that this has no unintended side effects. The following still work:

- auto sizing cells
- wrapping
- the Pkg bubble popping out of the cell on the right

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -2779,7 +2779,7 @@ Based on "Paraíso (Light)" by Jan T. Sott:
 }
 
 pluto-input .cm-editor .cm-scroller {
-    overflow-x: hidden;
+    overflow-y: hidden;
 }
 
 pluto-input .cm-editor .cm-content,
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -649,6 +649,7 @@ nav#at_the_top > div.desktop_picker span {
 
 pluto-filepicker .cm-scroller {
     scrollbar-width: none; /* Firefox */
+    overflow-x: hidden;
 }
 pluto-filepicker .cm-scroller::-webkit-scrollbar {
     display: none; /* Safari and Chrome */
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -649,7 +649,6 @@ nav#at_the_top > div.desktop_picker span {
 
 pluto-filepicker .cm-scroller {
     scrollbar-width: none; /* Firefox */
-    overflow-x: hidden;
 }
 pluto-filepicker .cm-scroller::-webkit-scrollbar {
     display: none; /* Safari and Chrome */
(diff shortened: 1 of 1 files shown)


frontend/editor.css
```diff
@@ -649,6 +649,7 @@ nav#at_the_top > div.desktop_picker span {
 
 pluto-filepicker .cm-scroller {
     scrollbar-width: none; /* Firefox */
+    overflow-x: hidden;
```
holomorphism, line 652: This setting must be moved to `pluto-input .cm-editor .cm-scroller`.
Also, the value must be `visible`, not `hidden`.

holomorphism: In my case, I added the following lines near line 2789 of this file and the scrollbar disappeared.
```css
pluto-input .cm-editor .cm-scroller {
    overflow-x: visible;
}
```

Pangoraw: @holomorphism I updated the PR with `overflow-x: hidden` with the right cm instance selector. Have you tried with `overflow-y: hidden` on the same selector (since the scrollbars are vertical) ?

holomorphism: @Pangoraw Sorry, you are right. When I set `overflow-y: hidden` instead of `overflow-x: visible`, the vertical scrollbar disappeared as well. (Note that scrollbars are still visible with the current `overflow-x: hidden`)

I think it would be appropriate to simply write only `overflow-y: hidden` there. Could you please fix this?

After checking with DevTools in chrome and finding `.cm-scroller` set to `overflow-x: auto`, I changed this value to `visible` and the vertical scrollbars disappeared, so I was happy with that.
I don't remember why I didn't try `overflow-y: hidden`, but I'm a little embarrassed!

Pangoraw: No problems, thanks for the extensive investigation!

fonsp:
Thanks everyone!

PR #2379, opened 2022-11-19
expose a function to return the bond names

disberd:
This PR is a companion of the PlutoSliderServer one https://github.com/JuliaPluto/PlutoSliderServer.jl/pull/97.
On Pluto side, we expose a function that returns a vector of all the variable names (Symbols) currently bound in the notebook

### DIFF
--- a/src/evaluation/RunBonds.jl
+++ b/src/evaluation/RunBonds.jl
@@ -51,6 +51,11 @@ function set_bond_values_reactive(;
 end
 
 """
+Returns the names of all defined bonds
+"""
+get_bond_names(session::ServerSession, notebook::Notebook) = WorkspaceManager.get_bond_names((session,notebook))
+
+"""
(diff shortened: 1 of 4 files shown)


fonsp:
I think this should be scoped to the notebook, because we sometimes run multiple notebooks on the same process, so on the same PlutoRunner module.

An idea is to add `cell_id` as argument in the PlutoRunner function:

PlutoRunner.jl:
```julia
function get_bond_names(cell_id::UUID)
    get(cell_registered_bond_names, cell_id, Set{Symbol}())
end
```

And then in RunBonds.jl, call this function for each cell in the notebook, and then `union` the results.

disberd:
> Awesome! This needs a test, can you find an existing test case that uses `@bind` and insert a line to test the output of this new function?

Ah didn't know this was possible!
Let's see if this rogue modification without testing passes the tests or I have to fix something :D

PR #991, opened 2021-03-11
Direct communication link between user JS and JL with WebIO support

dralletje:
Some things are still needed before merging, like writing the small HTTP-interop types for making AssetRegistry more generic.

~Also, I just copied in Require to prevent a dependency, but I admit this is not the most elegant.~ EDIT: we decided that this is the simplest way to go

Also, tests

PR #2392, opened 2022-11-24
Direct communication link between user JS and JL for JSServe support

fonsp:
Split off from #991. This PR is a subset of #991, only adding the simple JS-JL link, without WebIO support, and without AssetRegistry support. 

We decided to work on supporting JSServe instead of WebIO because JSServe is being actively maintained and developed (https://github.com/SimonDanisch/JSServe.jl/pull/118), while WebIO and its dependencies are not developed/maintained as much, and they are currently not compatible with HTTP 1.0, which is painful for the Julia ecosystem. 

JSServe will also be easier to support, especially in Pluto's HTML Export, as it does not require hooking into the HTTP server. Instead, it uses data URIs, and maybe #1124 to transfer data.

This will add stable out-of-the-box support for WGLMakie!!

Thanks @dralletje for the initial work for this PR!

# TODO:
- [ ] The API choice for the Julia side is still difficult. It would great to expose the Julia API via AbstractPlutoDingetjes, or perhaps a new package. It would help if we first finish:
  - [ ] https://github.com/JuliaPluto/AbstractPlutoDingetjes.jl/pull/8 and 
  - [ ] https://github.com/JuliaPluto/AbstractPlutoDingetjes.jl/pull/7
- [ ] The API for the JS side also needs to be decided. We currently define a global `Pluto` object, but we could also pass it into the script directly, like we do for `invalidation` and `currentScript`.
- [ ] We need a basic end-to-end test added to our puppeteer tests.


disberd:
@fonsp is there any way to help with this PR?
I think it would really enable a lot of cool functionality (look at all the internal refs to this PR) and it's a pity that is basically stale since quite a while

fonsp:
Replaced by https://github.com/fonsp/Pluto.jl/pull/2726

PR #2412, opened 2022-12-13
Make featured notebook sources a frontend launch param

fonsp:
This is necessary to render the featured notebooks directly on our WIP website :)

You can test it:
```
http://localhost:1234/?featured_source_url=https%3A%2F%2Fmit-c25.netlify.app%2Fpluto_export.json
```

This will show the regular index page, but with the notebooks from https://mit-c25.netlify.app/ as source! (These don't have images, because no image URLs are given in the frontmatters in that repository.)

Matching PlutoSliderServer PR: 
- https://github.com/JuliaPluto/PlutoSliderServer.jl/pull/100

### DIFF
--- a/frontend/components/welcome/Welcome.js
+++ b/frontend/components/welcome/Welcome.js
@@ -26,7 +26,6 @@ import default_featured_sources from "../../featured_sources.js"
 /**
  * @typedef LaunchParameters
  * @type {{
- * featured_static: boolean,
  * featured_direct_html_links: boolean,
  * featured_sources: import("./Featured.js").FeaturedSource[]?,
  * featured_source_url?: string,
@@ -55,8 +54,6 @@ export const Welcome = ({ launch_params }) => {
     const client_ref = useRef(/** @type {import('../../common/PlutoConnection').PlutoConnection} */ ({}))
 
     useEffect(() => {
-        if (launch_params.featured_static) return
-
         const on_update = ({ message, type }) => {
             if (type === "notebook_list") {
                 // a notebook list updates happened while the welcome screen is open, because a notebook started running for example
--- a/frontend/components/welcome/Welcome.js
+++ b/frontend/components/welcome/Welcome.js
@@ -16,19 +16,21 @@ import default_featured_sources from "../../featured_sources.js"
 /**
  * @typedef NotebookListEntry
  * @type {{
- *  notebook_id: String,
- *  path: String,
- *  in_temp_dir: Boolean,
- *  shortpath: String,
+ *  notebook_id: string,
+ *  path: string,
+ *  in_temp_dir: boolean,
+ *  shortpath: string,
  * }}
  */
 
 /**
  * @typedef LaunchParameters
  * @type {{
- *  featured_sources: import("./Featured.js").FeaturedSource[]?,
- * featured_source_url?: String,
- * featured_source_integrity?: String,
+ * featured_static: boolean,
+ * featured_direct_html_links: boolean,
+ * featured_sources: import("./Featured.js").FeaturedSource[]?,
+ * featured_source_url?: string,
+ * featured_source_integrity?: string,
  * }}
  */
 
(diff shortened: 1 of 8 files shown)


frontend/components/welcome/Welcome.js
```diff
@@ -93,41 +108,60 @@ export const Welcome = () => {
...
+     * @type {import("./Featured.js").FeaturedSource[]}
+     */
+    const featured_sources = preact.useMemo(
+        () =>
+            // Option 1: configured directly
+            launch_params.featured_sources ??
+            // Option 2: configured through url and integrity strings
+            (launch_params.featured_source_url
+                ? [{ url: launch_params.featured_source_url, integrity: launch_params.featured_source_integrity }]
+                : // Option 3: default
```
fonsp, line 122: @pankgeorg You could add an option between 2 and 3 to overload the sources with those from juliahub. Something like:

```js
 // Option 1: configured directly
            launch_params.featured_sources ??
            // Option 2: configured through url and integrity strings
            (launch_params.featured_source_url
                ? [{ url: launch_params.featured_source_url, integrity: launch_params.featured_source_integrity }]
                : // Option 3: overloaded by environment
				environment.featured_sources ??
				  // Option 4: default
                  default_featured_sources.sources

fonsp: Because this PR adds the `launch_params.featured_sources` config, but for JH integration the `environment` seems better suited, right?

BioTurboNick:
How do we use this? I was thinking it would be an argument to `Pluto.run()` somewhere but I'm not seeing anything like that. Do we need to hand-generate the json file being used? Or I'm misunderstanding what this is for.

I was hoping it could be used to present a series of read-only templates that people can save and then edit.

BioTurboNick:
@fonsp Just checking on this - is this parameter exposed to the `Pluto.run()` method somehow?

fonsp:
Hey Nick!

This is a "launch parameter" to the Pluto GUI, which is different from the `Pluto.run(...)` options, which mostly control the server. (The reason that there is a second set of settings is that you can also use the Pluto GUI without a server, like in an HTML export.)

The list of launch parameters is here: 
- For the editor: https://github.com/fonsp/Pluto.jl/blob/v0.19.31/frontend/editor.js#L28-L61
- For the main menu: https://github.com/fonsp/Pluto.jl/blob/v0.19.31/frontend/index.js#L8-L28

There are three ways you can set the launch parameters:
- As global javascript variables, defined before the editor loads. This is used when you click the "Export HTML" button inside Pluto. Generate an HTML and open it with a text editor to see how it works.
- As URL parameters: one place where this is used is when you click on a featured notebook in the Pluto main menu. It opens `editor.html`, but the path to the statefile, notebook file, etc is given with URL query paramaters. Open a featured notebook and look at the URL to see it.
- As HTML attributes to the `<pluto-editor>` web component. This is used in https://computationalthinking.mit.edu/ . Check out `view-source:https://computationalthinking.mit.edu/Fall23/images_abstractions/abstraction/` to see an example. 

---

For you use case, you probably want to generate a URL like I did in my original post and give that to people (use `encodeURIComponent` in JS or URIs.jl), or you can create an online website like featured.plutojl.org where people download the template files from.

The JSON file is generated automatically by PlutoSliderServer since https://github.com/JuliaPluto/PlutoSliderServer.jl/pull/68 . https://github.com/JuliaPluto/PlutoSliderServer.jl/pull/100 makes it possible to run a website like featured.plutojl.org yourself, where you see an index page in the style of Pluto's main menu. Check out https://github.com/JuliaPluto/featured to learn more about frontmatter and `pluto_export_configuration.json`.

BioTurboNick:
Thanks for the all the details! I'm not great with JS/HTML so I'm struggling a bit.

I'd rather not change what URL people use, I just want to provide a different default for the main page. Is it that the frontend can't really be modified by the server? Can I just change the source file, like `.julia/packages/Pluto/xxxxx`? I guess I'd just have to beware of updates reverting it...

fonsp:
Maybe you could just fork Pluto and edit the `featured_sources.jl` file?

BioTurboNick:
I'd rather not have to maintain a repo just for this.

fonsp:
Alright, let me know if I can help in another way

Issue #2190, opened 2022-06-25
DataFrames display truncates string columns, column much narrower than space provided


BioTurboNick:
![image](https://user-images.githubusercontent.com/1438610/175777510-b138e445-b360-4ae9-8aae-7246ab8a7fc3.png)

Why can't I see the full value? I would expect the table to take up the full width, have a scroll bar, and/or wrap.

Pluto 0.19.9
Julia 1.8.0-rc1

fonsp:
Some sample code:

```julia
# ╔═╡ bb939694-98da-11ed-2baf-3d6a075388ba
using DataFrames

# ╔═╡ 07eeffd7-713f-428f-b2ee-fd529e8f191d
DataFrame(:filename => ["DataFrames_display_truncates_string_columns,_column_much_narrower_than_space_provided"])

# ╔═╡ b72fdca2-f97b-4fd2-a972-71d0ede64b3c
DataFrame(
	:filename => ["DataFrames_display_truncates_string_columns,_column_much_narrower_than_space_provided", "2"],
	:b => [Dict(:a => [1,2,3], :b => [sin, tan, cos], :c => names(Base)), "3"]
)

# ╔═╡ a005493c-ea8e-4b6e-8a8b-243ab153aa98
DataFrame(
	:filename => ["DataFrames_display_truncates_string_columns,_column_much_narrower_than_space_provided", "2"],
	:b => [Dict(:a => [1,2,3], :b => [sin, tan, cos], :c => names(Base)), "3"],
	:c => 1:2,
	:d => 1:2,
	:e => 1:2,
	:f => 1:2,
)

# ╔═╡ 10eeff5b-2edc-4757-b36a-152496793379
DataFrame(
	:filename => ["DataFrames_display_truncates_string_columns,_column_much_narrower_than_space_provided", "2"],
	:b => [Dict(:a => [1,2,3], :b => [sin, tan, cos], :c => names(Base)), "3"],
	:DataFrames_display_truncates_string_columns => 1:2,
)

```

genericallyterrible:
Looks great on Chrome, but Firefox doesn't want to play nice. Not sure where to raise the issue since it's not really Pluto's fault.

| Browser | Result |
| --- | --- |
| Chrome | ![image](https://github.com/fonsp/Pluto.jl/assets/11425944/9154c81f-1cda-4fcf-983d-3ee3354c418b) |
[...]

PR #2425, opened 2022-12-26
Add column resizing functionality to Pluto tables.

kcin96:
This PR proposes a potential fix for #2190 and adds the following features to pluto tables :

1. Allows user to resize columns by dragging a div between column headers. 
2. Double clicking on the div between columns resizes that column to fit the longest string/item in that column.

https://user-images.githubusercontent.com/26337751/209561873-95df729d-2cf5-4e33-aa20-a0313d108fbd.mov

### DIFF
--- a/frontend/components/TreeView.js
+++ b/frontend/components/TreeView.js
@@ -154,8 +154,79 @@ export const TableView = ({ mime, body, cell_id, persist_js_state }) => {
             ? null
             : html`<thead>
                   <tr class="schema-names">
-                      ${["", ...body.schema.names].map((x) => html`<th>${x === "more" ? more(2) : x}</th>`)}
-                  </tr>
+                      ${["", ...body.schema.names].map((x, i) => html`<th>${x === "more" ? more(2) : x}
+                      <div class="table-col-resizer" onmousedown=${(e) => {
+                            const table = node_ref.current;
+                            const columns = Array.from(table.querySelector(".schema-names").querySelectorAll("th"));
+                            let x = e.clientX;  //initial mousedown x value
+                            let table_width = parseFloat(window.getComputedStyle(table).width);
+                            
+                            //Initialize and sets each column width to rendered column width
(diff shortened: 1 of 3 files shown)


fonsp:
Hi @kcin96, thanks for your PR! Sorry for the late response, it slipped from me during the winter break.

The fix to #2190 could be quite simple: we can change `overflow: hidden` to `overflow: auto`:

https://github.com/fonsp/Pluto.jl/blob/bedc7767d76439477bae8a5165f4f39906f9064c/frontend/treeview.css#L279-L282

To make better use of screen space, we could also tweak the `max-width` declaration. We could make it depend on the number of columns, viewport width, or perhaps something dynamic, where you can click to expand a column.

For resizing, did you try the CSS `resize` property? If so, why did it not work?

Did you consider a simpler solution like this already before ending up with the current solution? What advantages/disadvantages did you find?

The current implementation is a cool feature, but the disadvantages are:
- Complex implementation: Using tools like measuring computed width, estimating maximum width and directly analysing and modifying DOM (outside of react) has a lot of surface area for bugs, because tables can contain all sorts of user data. By keeping the logic simple (and CSS based), it is more robust for unpredictable use. 
- Maintenance cost: there is quite a bit of new code here, which will require more maintenance in the future from our side. 
- Some users might be confused by the feature. The behaviour might not be self-explanatory.

I understand that this isn't the reaction you are hoping for! We really appreciate contributions from the community, especially to our JavaScript code. I would be happy to collaborate on your next PR, and let's make sure to communicate more on the design process before and during your work on a problem.

kcin96:
Hey @fonsp, thanks for your feedback!

For resizing I did initially try the resize property but I found that with long tables (multiple columns spanning the full width of the Pluto cell) resizing doesn't expand past the right table boundary (The amount of column resizing is limited by the initially rendered table width.

I totally agree with your points on the disadvantages. Thanks for taking the time to provide feedback! :)

Issue #2309, opened 2022-10-11
Tables.jl object not triggering Pluto table viewer


juliohm:
The [attached notebook](https://julialang.zulipchat.com/user_uploads/7178/scisJPpde4SIT56FWspa-R6n/test.jl) (see also below) defines a `geotable`, which is a Tables.jl object. The Pluto's built-in table viewer is not being triggered:

![image](https://user-images.githubusercontent.com/3345261/195069434-7058990c-5a69-4e69-8324-3684f6983e5d.png)

Can you help understand why that is the case?

pankgeorg:
Hey! I can't reproduce this with `julia 1.8.2`, `pluto 0.19.12` or *13 running on WSL. Can you share more about your setup?

juliohm:
I am on the same setup, Julia v1.8.2, Pluto v0.19.13, Linux. What is the current logic to display the table viewer? I could try to help debug what is happening with this specific table type.

pankgeorg:
The logic resides in PlutoRunner.jl, mostly.

- The type must not have specialized show methods: https://github.com/fonsp/Pluto.jl/blob/main/src/runner/PlutoRunner.jl#L1077-L1095

- You must allow the mime (not override this https://github.com/fonsp/Pluto.jl/blob/main/src/runner/PlutoRunner.jl#L1103)
-

juliohm:
@pankgeorg it is very strange that it only happens with that particular dataset I shared. If you open the same notebook and create a different object without using the dataset, then it displays properly:

![image](https://user-images.githubusercontent.com/3345261/195092186-b6151b36-01f5-4753-8cad-5f397e4925d3.png)

I wonder why given that these two objects have the same type in theory. Is there any logic on the number of columns for example?

juliohm:
Here you can find the show methods:

https://github.com/JuliaGeometry/Meshes.jl/blob/master/src/traits/data.jl#L201

rikhuijzer:
For future reference and people who cannot login to Zulip, this is the notebook:

```julia
### A Pluto.jl notebook ###
# v0.19.9

using Markdown
using InteractiveUtils

# ╔═╡ 7bf2c6b2-48e0-11ed-03a1-713809e45593
using GeoStats

# ╔═╡ caf64249-e6d7-494c-873c-c8658dd5a2d6
using CSV

# ╔═╡ 81a68a55-5874-4efa-b1ce-4ea0467393e6
table = let
[...]

PR #2429, opened 2023-01-02
fix pluto_showable for Tables.jl without iterate

pankgeorg:
Fixes https://github.com/fonsp/Pluto.jl/issues/2309#issuecomment-1342774073 @juliohm can you check it?

`isempty` depends on `iterate` on the general case, which is _not_ guaranteed to exist for Table objects; rather, we're guaranteed to be able to get an iterator by requesting `Table.rows` on the object.

@fonsp do you have any reason in particular to avoid that, except performance? I'm a bit worried that for _some_ objects (e.g. database cursors) this will move a cursor or something; but in the general case it should be fine.

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -1573,7 +1573,7 @@ const integrations = Integration[
             end
 
 
-            pluto_showable(::MIME"application/vnd.pluto.table+object", x::Any) = try Tables.rowaccess(x)::Bool && !isempty(x) catch; false end
+            pluto_showable(::MIME"application/vnd.pluto.table+object", x::Any) = try Tables.rowaccess(x)::Bool && !isempty(Tables.rows(x)) catch; false end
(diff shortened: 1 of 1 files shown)


juliohm:
Thank you @pankgeorg I will give it a try soon.

juliohm:
I confirm that the PR fixes the issue 👍🏽

fonsp:
Discussed this in the Pluto call today, the issue was caused by https://github.com/fonsp/Pluto.jl/commit/309d29be62ba5050629522778a11edcb7dd7dc10 . This fix was a bit hacky, the proper solution is to implement this in the frontend: if there are no rows, display the text "Table with no rows". That means that we can undo https://github.com/fonsp/Pluto.jl/commit/309d29be62ba5050629522778a11edcb7dd7dc10 and fix https://github.com/fonsp/Pluto.jl/issues/2309#issuecomment-1342774073

Issue #2427, opened 2023-01-02
Bug with anonymous functions in Julia 1.9


greimel:
Some my old notebooks stopped working on Julia 1.9.

 I reduced it to the following MWE

![image](https://user-images.githubusercontent.com/6280307/210222074-218ae320-3192-4969-9ad6-1ce217f116df.png)

Pluto doesn't find bindings when they are used inside an anonymous function.

The code is here

https://gist.github.com/greimel/e4f87e0f7162ddb5bbf5aac7ed477867

Notebook runs fine in Julia 1.8 and in the 1.9 REPL.

PR #2433, opened 2023-01-04
Test Julia 1.9

Pangoraw:
- Test 1.9
- Potential fix to #2427
    This seems to be a weird issue with splatting or `union!` but I
    could not manage to reproduce outside.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -34,7 +34,7 @@ jobs:
             fail-fast: false
             matrix:
                 # We test quite a lot of versions because we do some OS and version specific things unfortunately
-                julia-version: ["1.6", "1.7", "1.8"] #, "~1.9.0-0"] #, "nightly"]
+                julia-version: ["1.6", "1.7", "1.8", "~1.9.0-0"] #, "nightly"]
                 os: [ubuntu-latest, macOS-latest, windows-latest]
 
         steps:
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -567,7 +567,7 @@ function explore_function_macro!(ex::Expr, scopestate::ScopeState)
     # Macro are called using @funcname, but defined with funcname. We need to change that in our scopestate
     # (The `!= 0` is for when the function named couldn't be parsed)
     if ex.head == :macro && length(funcname) != 0
-        funcname = Symbol[Symbol("@$(funcname[1])")]
+        funcname = Symbol[Symbol('@', funcname[1])]
         push!(innerscopestate.hiddenglobals, only(funcname))
     elseif length(funcname) == 1
         push!(scopestate.definedfuncs, funcname[end])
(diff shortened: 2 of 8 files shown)


fonsp:
The tests take 50% longer to run on 1.9 🤔 cc @rikhuijzer

rikhuijzer:
> The tests take 50% longer to run on 1.9 thinking cc @rikhuijzer

Yes I'm very happy with Paul's work because now we can hopefully use the new package compilation better. I'm hope to find some time soon to switch `Pluto.jl` from `PrecompileSignatures.jl` to `SnoopPrecompile.jl`. That should probably reduce TTFX.

greimel:
That's great! Do you plan to release this anytime soon?

fonsp:
Yep! Rik is working on #2441 , we will release both PRs together probably. Just releasing this PR makes me worries about the 50% performance loss

rikhuijzer:
> The tests take 50% longer to run on 1.9 thinking cc @rikhuijzer

Likely due to a bug in Julia-1.9.0-beta2 which was fixed in Julia-1.9.0-beta3. In beta2, `import Pluto` took 59 seconds while it takes only 2 seconds in beta3.

pankgeorg:
> > The tests take 50% longer to run on 1.9 thinking cc @rikhuijzer
> 
> Likely due to a bug in Julia-1.9.0-beta2 which was fixed in Julia-1.9.0-beta3. In beta2, `import Pluto` took 59 seconds while it takes only 2 seconds in beta3.

For the first time, I get ~80 on master. Is this with a clean cache on `main`? (I noticed `julia` is smart enough to cache precompilations through branches)

rikhuijzer:
> > > The tests take 50% longer to run on 1.9 thinking cc @rikhuijzer
> > 
> > 
> > Likely due to a bug in Julia-1.9.0-beta2 which was fixed in Julia-1.9.0-beta3. In beta2, `import Pluto` took 59 seconds while it takes only 2 seconds in beta3.
> 
> For the first time, I get ~80s on `main`. Is this with a clean cache on `main`? (I noticed `julia` is smart enough to cache precompilations through branches)

What do you do to get that time and which Julia version are you running?

fonsp:
I think the number from @pankgeorg included precompilation (you get this when you run `import Pluto` directly after switching branches or Julia versions), which can take 80 seconds on Julia 1.9 😵‍💫

pankgeorg:
Yes this is the precompilation time, without any cache.

fonsp:
I have been discussing this @rikhuijzer , precompilation went from 15s to 60s (!) in Julia 1.9, and TTFX went from 30s to 20s. I'm a bit worried and I think we should open a Julia issue, Rik is mostly happy with the TTFX gains :)

60 seconds seems extremely long to me, that's 120 seconds on an older laptop. After seeing most other packages precompile in 3 seconds, people might assume that something is broken and restart Julia...

greimel:
Right, but you already get the regression with the current release of Pluto on Julia 1.9 beta (in addition to Pluto bugs). Isn't releasing this already a strict improvement? The precompile time is orthogonal to the bug fixed here and can be fixed in a later release, can't it?

fonsp:
Publishing a release is [more complex than it sounds](https://youtu.be/o_4EX4dPppA?t=144), but yes, a release is coming 👍

Issue #1385, opened 2021-08-11
Failing to display DataFrame() output


anuraag94:
I have a Pluto notebook where I'm running a simulation, and the output is in the form of a dataframe from the Julia package DataFrames. However, I only see this blank field as output. 

<img width="828" alt="image" src="https://user-images.githubusercontent.com/6537752/129077134-00a29e9c-405a-4cbf-8309-5cf1a36f1b91.png">

My output does print as expected in a Jupyter notebook. If DataFrame output is currently unsupported, are there plans to bring it to Pluto? If not, what would be a good workaround for me?

fonsp:
Could it be that the table is empty? Can you use the `Text` function to compare it to plaintext output?

PR #2438, opened 2023-01-09
fix: `pluto_showable` for objects without implementation of `length(rows(::Tablelike))`

pankgeorg:
Fix https://github.com/fonsp/Pluto.jl/issues/2309, supersedes https://github.com/fonsp/Pluto.jl/pull/2429

As discussed in https://github.com/fonsp/Pluto.jl/pull/2429#issuecomment-1370061793 , this PR is a cleaner solution to https://github.com/fonsp/Pluto.jl/issues/1385, replacing the original workaround, https://github.com/fonsp/Pluto.jl/commit/309d29be62ba5050629522778a11edcb7dd7dc10 .


src/runner/PlutoRunner.jl
```diff
@@ -1516,7 +1516,13 @@ const integrations = Integration[
...
             function table_data(x::Any, io::Context)
                 rows = Tables.rows(x)
-
+                if isempty(rows)
+                    return Dict{Symbol,Any}(
+                        :objectid => objectid2str(x),
+                        :schema => [],
+                        :rows => [],
+                    )
+                end
```
fonsp, line 1525: We don't need to add this, right? If the table is empty, it should already return this dict.

fonsp:
Looks good otherwise! Remember to link & close the relevant issues and commit

frontend/components/TreeView.js
```diff
@@ -135,7 +135,12 @@ export const TreeView = ({ mime, body, cell_id, persist_js_state }) => {
 export const TableView = ({ mime, body, cell_id, persist_js_state }) => {
     let pluto_actions = useContext(PlutoActionsContext)
     const node_ref = useRef(null)
-
+    console.log(body.rows, body.rows?.length ?? 0)
```
fonsp, line 138: console.log

PR #2449, opened 2023-01-18
Show hint when you click on a global without Ctrl down

fonsp:
https://user-images.githubusercontent.com/6933510/213256615-8cbcddaa-cdf4-4b0c-ace0-1a607a7a5e08.mov

### DIFF
--- a/frontend/components/CellInput/go_to_definition_plugin.js
+++ b/frontend/components/CellInput/go_to_definition_plugin.js
@@ -2,6 +2,8 @@ import { Facet, ViewPlugin, Decoration, EditorView } from "../../imports/Codemir
 import { ctrl_or_cmd_name, has_ctrl_or_cmd_pressed } from "../../common/KeyboardShortcuts.js"
 import _ from "../../imports/lodash.js"
 import { ScopeStateField } from "./scopestate_statefield.js"
+import { open_pluto_popup } from "../Popup.js"
+import { html } from "../../imports/Preact.js"
 
 /**
  * @param {any} state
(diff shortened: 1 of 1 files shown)


ederag:
I agree that it might quickly become too distracting.
Would it be possible to have a "do not show again" (even when restarting the Pluto server) ?
Maybe as a cookie ?

ederag:
Another stance could be to only show the wiggling underlines (that make the variable a link)
when <kbd>Ctrl</kbd> is pressed.
This way, the first time the user hits <kbd>Ctrl</kbd>, it should be clearer that something new can be done.
And the code display would be cleaner (those underlines _slightly_ alter my reading) when <kbd>Ctrl</kbd> is released.

fonsp:
> Another stance could be to only show the wiggling underlines (that make the variable a link) when Ctrl is pressed. This way, the first time the user hits Ctrl, it should be clearer that something new can be done. And the code display would be cleaner (those underlines _slightly_ alter my reading) when Ctrl is released.

Thanks for the feedback! This was exactly my thought when I made the underlines bright pink when you press `Ctrl`! 😅

ederag:
> This was exactly my thought when I made the underlines bright pink when you press Ctrl

:rofl: Being so used to the effect, I forgot about it !?
Maybe the effect would be stronger if the underlines were normally invisible ?
(a change of color is not as strong as something appearing)

fonsp:
:) I like the fat underlines, since globals play the lead role in our reactivity. But good to hear your feedback! I will take it into consideration the next time this is up for a restyle.

greimel:
I like a lot that globals stand out visually!

PR #2468, opened 2023-02-07
Improve print view styling

tecosaur:
Here are a few small tweaks that I think improve the print style.

- Remove (useless) symbol definition underlines. These could be linked in the PDF (as in HTML), but they aren't, and so they end up being visual noise.
- Re-add the left cell border, in case the browser is hiding background colors, otherwise cells can look a bit funky.
- Hide the gutter element, it serves no purpose here.
- Hide the automatic page/url/date annotations

### Before (Firefox default print settings)

image

### After (Firefox default print settings)

image

### DIFF
--- a/frontend/hide-ui.css
+++ b/frontend/hide-ui.css
@@ -22,15 +22,9 @@ pluto-runarea,
     pkg-status-mark {
         display: none !important;
     }
-    [data-pluto-variable] {
-        text-decoration-color: transparent;
-    }
     pluto-input .cm-editor {
         border-left: 1px solid var(--normal-cell-color);
         border-radius: 4px !important;
-    }
-    .cm-gutterElement {
-        display: none;
     }
 }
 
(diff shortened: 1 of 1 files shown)


fonsp:
I want to keep the underlines because of #2450 and the gutters because of https://github.com/fonsp/Pluto.jl/commit/33cb224892eb5cbc0b7caeb35bf3de276a320930 . Good point about making them not selectable!

tecosaur:
> I removed the `@page` rules because we probably just want the user/browser to choose these settings.

Eh, I'm not sure about that.

> It looks like this is [a trick](https://stackoverflow.com/questions/8228088/remove-header-and-footer-from-window-print) to make the content flow on top of the headers, not an API to change the setting directly.

In a way it is, but you could then describe it as "a trick without side effects". The reality is, it seems to be the only way you can do so currently, and so I'd think worthwhile.

PR #2469, opened 2023-02-07
Change "Static PDF" export title to just "PDF"

tecosaur:
Just a minor tweak with the export banner: PDFs are already a static format, and it's stated in the description, and so this is implicit and redundent 😛.

image


fonsp:
Thanks!

PR #2471, opened 2023-02-07
Add plaintext export options

tecosaur:
This provides the ability to export to plaintext formats, namely:
- Markdown, which is commonly accepted by other tools
- Org, another plaintext interactive notebook format (with support for cross-references, citations, bibliography generation, inline LaTeX, and other publication-related things Pluto probably doesn't want to concern itself with)

image

This gives more options for post-processing and use with other tools, for instance turning some initial experiments in Pluto into a publication.

Since further use as a computational notebook is possible with Org, it has been fairly straightforward to also add bidirectional support for the embedded project/manifest, via a pluto export backend on the Org side of things.

#### Change breakdown
- 45 lines of boilerplate
- 5 lines of minor tweaks
- 5 lines for the pentagon shape
- 30 lines of generic plaintext export code
- 35 lines of format-specific code


tecosaur:
Hmmm, not sure what's causing the CI errors with Julia 1.6. I see there's a TOML parser error and a method error.
```
┌ Error: Failed to load notebook files: Project.toml+Manifest.toml parse error. Trying to recover Project.toml without Manifest.toml...
│   exception =
│    Could not parse manifest: TOML Parser error:
│    /tmp/jl_GB8bfv/Manifest.toml:3:-1 error: invalid bare key character: '['
│      [afsd[
│           ^ 
...

│    MethodError: no method matching project_rel_path(::Pkg.Types.Context, ::Nothing)
│    Closest candidates are:
│      project_rel_path(::Pkg.Types.Context, ::String) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:866
```

I can't see how these could relate to this PR, I must say.

tecosaur:
Ok, looks like there was an issue with `Returns(x)` on 1.6, I've replaced that pattern with `_ -> x` in the PR. Hopefully the tests will be happier now.

Pangoraw:
Thanks for looking into this!
Maybe it would be easier to maintain the multiple export formats by going the way of [PlutoStaticHTML.jl](https://github.com/rikhuijzer/PlutoStaticHTML.jl) and creating PlutoPlainMarkdown.jl, PlutoPlainOrgMode.jl, PlutoWordDocument.jl... packages? This would be quite straightforward since these export methods don't currently include the runtime outputs of cells.

tecosaur:
> Maybe it would be easier to maintain the multiple export formats by going the way of [PlutoStaticHTML.jl](https://github.com/rikhuijzer/PlutoStaticHTML.jl) and creating PlutoPlainMarkdown.jl, PlutoPlainOrgMode.jl, PlutoWordDocument.jl... packages?

Hmmm, to me this feels like an over-complication. With the `generate_plaintext` helper function you're essentially proposing a four-line package and a thirty-line package? Leaving word aside for a moment (as I can imagine that one actually being more involved), for the main plaintext export cases I think having them simply be in Pluto makes a fair bit of sense.

Regarding "easier to maintain", one of the nice things about plaintext export (as opposed to HTML, word, etc.) is that the formats are very simple, and settled* (not a moving target). So if this works (which from my testing, it does seem to :stuck_out_tongue:) the only "maintenance" required is if Pluto makes major changes to its internals (e.g. the structure of `Cell`).

> This would be quite straightforward since these export methods don't currently include the runtime outputs of cells.

I actually think it would be nice to include outputs when exporting, however I don't know how to handle "large" outputs (e.g. a massive table/matrix), and so have left this idea alone.

Also, regarding having this in pluto itself I've thought of a (potentially) nice way of compacting the two plaintext export options into one UI "block", which I thought I'd explore in another PR if/when this one is sorted.

image

fonsp:
Hey @tecosaur !

Thanks again for your PR and sorry for taking a long time to respond. I had a hard time finding the motivation to write a comment that you probably won't be satisfied with :( When reading this, remember that our **target audience** is beginner programmers and their teachers, and try to empathise with our [core values](http://plutojl.org/).

You definitely convinced me that some users would really appreciate a Markdown and Org export option, but I still don't think that they should be offered as export options from the GUI because:
1. They add options (Pluto users really appreciate how minimal our UI is; this is very important for beginners) and these options are complicated, and lead to more questions: 
  - what is Markdown/Org? what can i do with these files? (this last question does not have a simple answer)
  - with 5+ export options, which one should I choose?
2. They break our WYSIWYG guarantee that our other options have:
  - cells that output data structures like vectors, dicts, structs will display differently
  - special HMTL outputs like HypertextLiteral, plotly plots, PlutoVista.jl, PlutoUI etc will probably not display correctly in the end result
  - Interpolating values into Markdown macros will not work
3. They don't promote reproducibility, one of our core values:
  - You can't (currently) convert back. For a PDF this is expected, but Markdown/Org are editable, not final products.
    - This also means: If you edit the Markdown/Org file by hand, the final document is not reproducible.
  - Pluto does not execute cells top-to-bottom, but the consumer of the Markdown/Org probably will
  - Embedded Pkg info is (currently) not included in Markdown. There is (currently) no system to execute a Pluto-generated Org file with the Pkg info.


So these are some reasons why I don't think this feature is a good fit: it does not seem useful to our target audience, and it makes Pluto more difficult to use. Of course, this code can still be really useful to other more advanced users! This is very similar to https://github.com/rikhuijzer/PlutoStaticHTML.jl, and I think the same approach would work well for Markdown and Org, exactly in the way that Paul suggested https://github.com/fonsp/Pluto.jl/pull/2471#issuecomment-1460668933 . I don't see registering packages as an over-complication, it's actually a great way to distribute and develop a project like this.

tecosaur:
Hi Fons,

Thanks for taking the time to compose a thoughtful response.

> You definitely convinced me that some users would really appreciate a Markdown and Org export option

As one of those users, I'm very glad to hear this 🙂. Ultimately, I'm not tied
to the particular way this PR makes that possible, but I'd really like to see
something that is comparatively easy.

----

> I still don't think that they should be offered as export options from the GUI because ...

I have a few comments here, if I may.

> 1. They add options (Pluto users really appreciate how minimal our UI is; this is
> very important for beginners)

A minimal UI is very nice, but so long as it isn't at the expense of
functionality 😛.

> [...] and these options are complicated, and lead to more questions: (1.1) what is
> Markdown/Org? what can i do with these files? (this last question does not
> have a simple answer)

I feel like plaintext markup is pervasive enough that beginner programmers and
teachers of technical subjects will at least have a passing familiarity — at
least my experience at my local university has been consistent with this.

If you shift to a school level, I can see this won't hold as strongly.

> (1.2) with 5+ export options, which one should I choose?

I feel like this actually isn't a complicated question:
- For web stuff/exactly-as-seen sharing: HTML
- For a static document, containing everything, that's highly portable: PDF
- For further work: Markdown/Org, you'll know which if you want this option

> 2. They break our WYSIWYG guarantee that our other options have:

I'd suggest that when wanting to work on a document further, maintaining Pluto
WYSIWYG is not actually key, since the intent will likely be to modify the
content.

> (2.1) cells that output data structures like vectors, dicts, structs will display differently

Currently, there's no output whatsoever, so it's not _incorrect_, it's
_omitted_, which I think are somewhat different.

> (2.3) Interpolating values into Markdown macros will not work

Indeed, there's not much that I think can be done about this though. If we care
to try, interpolation is possible in Org though I'll note.

> 3. They don't promote reproducibility, one of our core values:

> (3.1) You can't (currently) convert back.

Actually, for Org at least, you can.
[ox-pluto](https://github.com/tecosaur/ox-pluto) provides bi-directional
compatibility.

> (3.1.1) This also means: If you edit the Markdown/Org file by hand, the final
> document is not reproducible.

The _original_ exported document is of course not recoverable form a modified
export (just as if you modified the HTML or PDF), however that does not mean
it's not reproducible.

In the case of the Org export, because the Project.toml and Manifest.toml are
embedded, the Org export is actually exactly as reproducible as pluto `.jl`
file.

> (3.3.1) Embedded Pkg info is (currently) not included in Markdown.

Mmm, I'm not aware of a good solution to this.

> (3.3.2) There is (currently) no system to execute a Pluto-generated Org file
> with the Pkg info.

Actually, the current systems will use the Project/Manifest.toml just fine 🙂.

----

> Of course, this code can still be really useful to other more advanced users!
> This is very similar to https://github.com/rikhuijzer/PlutoStaticHTML.jl, and
> I think the same approach would work well for Markdown and Org, exactly in the
> way that Paul suggested #2471 (comment) . I don't see registering packages as
> an over-complication, it's actually a great way to distribute and develop a
> project like this.

A key difference I see between these options and PlutoStaticHTML is that (as I
see it) PlutoStaticHTML fits in with a website generation pipeline, whereas
plaintext export options are "oh, I want this document as an X" like the
HTML/PDF export. These two use-cases/situations seem distinctly different to me.

From what I can tell, both of our thoughts on this can almost be summed up like so:
1. HTML/PDF export are good options to have for most people
2. Some people may want to have a plaintext export option
3. Adding more options clutters the UI
4. Plaintext export's won't work the same way as a pluto notebook if/when edited
  further, and may (in the case of Markdown) loose reproducibility properties

Personally, I don't see (3) and (4) as overriding the usefulness of (2).

Regarding making a package, similar to `PlutoStaticHTML`, my main problem with 
this is how user-unfriendly I think it is. Consider the following situation:

Say I have a pluto file open, and I think "oh, I want this as plaintext". With the PR the process was:
- Open the export tab
- Click on the Markdown/Org option

With a package, the procedure instead becomes something like this:
- Open a new Julia session, in a terminal
- Load `PlutoPlaintextExport` (and wait for TTFX)
- Switch back to the Pluto window, copy the file path
- Type `PlutoPlaintextExport.do("<file path>")` and paste the file path

This is significantly more involved, and I'm not happy about this degradation in
usability.

Perhaps there's some scope for a best-of-both-worlds option? Either hiding the
UI for the plaintext export behind a toggle/setting, or making it possible for a
package to add an export that is not much harder to get to than the HTML/PDF
options.

I'd be keen to hear your thoughts on how a good experience could be possible for
people interested in this option.

fonsp:
Hey @tecosaur 

It can be frustrating to hear that a feature proposal is rejected, but please remember that we have to focus on the core values of our project, and we have to be mindful of the time we have available for that. Here is a talk that really helped me better understand difficult situations like this one, I highly recommend it! https://www.youtube.com/watch?v=o_4EX4dPppA

Sorry for not answering your follow-up questions, I hope you can find yourself in some of the answers we have given so far.

tecosaur:
Hi Fons, I'll write a shorter question :stuck_out_tongue:: is there any interest in exploring ways to allow for an easy plaintext export, other than this PR?  For example, adding something that another package could hook into to provide more options for users that want them.

tecosaur:
I must admit that at this point I'm feeling slightly disappointed. I hope to be proven wrong, but it looks like there's no willingness to engage constructively on this topic :pensive:.

PR #2496, opened 2023-03-09
A few random cleanups

sjkelly:
This fixes/eases the demands on inference in a few cases, by eliding some union splits and enforcing more concrete types. Also a few minor typo corrections.

### DIFF
--- a/src/analysis/Errors.jl
+++ b/src/analysis/Errors.jl
@@ -23,16 +23,19 @@ function MultipleDefinitionsError(topology::NotebookTopology, cell::Cell, all_de
 	)
 end
 
-hint1 = "Combine all definitions into a single reactive cell using a `begin ... end` block."
-hint2 = "Wrap all code in a `begin ... end` block."
+const hint1 = "Combine all definitions into a single reactive cell using a `begin ... end` block."
 
 # TODO: handle case when cells are in cycle, but variables aren't
 function showerror(io::IO, cre::CyclicReferenceError)
-	print(io, "Cyclic references among $(join(cre.syms, ", ", " and ")).\n$hint1")
+	print(io, "Cyclic references among ")
+	println(io, join(cre.syms, ", ", " and "))
+	print(io, hint1)
 end
 
 function showerror(io::IO, mde::MultipleDefinitionsError)
-	print(io, "Multiple definitions for $(join(mde.syms, ", ", " and ")).\n$hint1") # TODO: hint about mutable globals
+	print(io, "Multiple definitions for ")
+	println(io, join(mde.syms, ", ", " and "))
+	print(io, hint1) # TODO: hint about mutable globals
 end
 
 "Send `error` to the frontend without backtrace. Runtime errors are handled by `WorkspaceManager.eval_format_fetch_in_workspace` - this function is for Reactivity errors."
--- a/src/evaluation/WorkspaceManager.jl
+++ b/src/evaluation/WorkspaceManager.jl
@@ -530,9 +530,9 @@ function collect_soft_definitions(session_notebook::SN, modules::Set{Expr})
     end)
 end
 
-function macroexpand_in_workspace(session_notebook::SN, macrocall, cell_id, module_name = nothing; capture_stdout::Bool=true)::Tuple{Bool, Any}
+function macroexpand_in_workspace(session_notebook::SN, macrocall, cell_id, module_name = Symbol(""); capture_stdout::Bool=true)::Tuple{Bool, Any}
     workspace = get_workspace(session_notebook)
-    module_name = module_name === nothing ? workspace.module_name : module_name
+    module_name = module_name === Symbol("") ? workspace.module_name : module_name
 
     Distributed.remotecall_eval(Main, workspace.pid, quote
         try
--- a/src/packages/Packages.jl
+++ b/src/packages/Packages.jl
@@ -12,6 +12,8 @@ const tiers = [
 
 const pkg_token = Token()
 
+_default_cleanup() = nothing
+
 # This list appears multiple times in our codebase. Be sure to match edits everywhere.
 function use_plutopkg(topology::NotebookTopology)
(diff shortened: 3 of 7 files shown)


src/analysis/Errors.jl
```diff
@@ -23,16 +23,19 @@ function MultipleDefinitionsError(topology::NotebookTopology, cell::Cell, all_de
...
+	print(io, "Cyclic references among ")
+	println(io, join(cre.syms, ", ", " and "))
+	print(io, hint1)
 end
 
 function showerror(io::IO, mde::MultipleDefinitionsError)
-	print(io, "Multiple definitions for $(join(mde.syms, ", ", " and ")).\n$hint1") # TODO: hint about mutable globals
+	print(io, "Multiple definitions for ")
+	println(io, join(mde.syms, ", ", " and "))
+	print(io, hint1) # TODO: hint about mutable globals
```
rikhuijzer, line 38: Is this TODO something that can be done immediately or can an issue be opened? TODO's in code tend to not lead to any action

src/packages/Packages.jl
```diff
@@ -363,7 +365,7 @@ function sync_nbpkg(session, notebook, old_topology::NotebookTopology, new_topol
 
 		save && save_notebook(session, notebook)
 	finally
-        isnothing(cleanup[]) || cleanup[]()
+        cleanup[]()
```
rikhuijzer, line 368: Nice.
```julia
julia> @time cleanup[]()
  0.000007 seconds
```

rikhuijzer (approved):
Nice finds. It looks good and these parts should be covered pretty well by the tests, so it looks good to me

src/analysis/Errors.jl
sjkelly: Hmm I'm not sure, though I do think that line should go back to a single `print` statement

rikhuijzer: Oh sorry my bad. I didn't see that you just copied the existing TODO.

rikhuijzer:
I'll leave merging this PR to @fonsp. He'll probably take a look at some point.

Thanks @sjkelly!

src/Configuration.jl
```diff
@@ -66,8 +66,8 @@ The HTTP server options. See [`SecurityOptions`](@ref) for additional settings.
 # Keyword arguments
 
 - `host::String = "$HOST_DEFAULT"` Set to `"127.0.0.1"` (default) to run on *localhost*, which makes the server available to your computer and the local network (LAN). Set to `"0.0.0.0"` to make the server available to the entire network (internet).
-- `port::Union{Nothing,Integer} = $PORT_DEFAULT` When specified, this port will be used for the server.
-- `port_hint::Integer = $PORT_HINT_DEFAULT` If the other setting `port` is not specified, then this setting (`port_hint`) will be used as the starting point in finding an available port to run the server on. 
+- `port::Union{Nothing,Int} = $PORT_DEFAULT` When specified, this port will be used for the server.
```
fonsp, line 69: Unfortunately this would be a breaking change (this is public API). I would say to leave it at `Integer` unless we have good evidence that it would improve performance.

src/analysis/Parse.jl
```diff
@@ -83,7 +83,7 @@ fix_linenumbernodes!(::Any, actual_filename) = nothing
 `expression_boundaries("sqrt(1)\n\n123") == [ncodeunits("sqrt(1)\n\n") + 1, ncodeunits("sqrt(1)\n\n123") + 1]`
 
 """
-function expression_boundaries(code::String, start=1)::Vector{<:Integer}
+function expression_boundaries(code::String, start=1)
```
fonsp, line 86: The return type is a form of documentation here

src/evaluation/WorkspaceManager.jl
```diff
@@ -15,7 +15,7 @@ Each notebook gets at most one `Workspace` at any time, but it can also have no
 (it cannot `eval` code in this case).
 """
 Base.@kwdef mutable struct Workspace
-    pid::Integer
+    pid::Int
```
fonsp, line 18: Could you check that `Distributed.addprocs` indeed returns an `Int`, not `Int64` on a 32 bit system?

fonsp:
Thanks! Some small comments

src/evaluation/WorkspaceManager.jl
sjkelly: Yes it is `Int` on 1.8~x86:
```
julia> using Distributed; typeof(Distributed.addprocs(1))
Vector{Int32} (alias for Array{Int32, 1})

```

src/analysis/Parse.jl
sjkelly: Ah yes. I see:

https://github.com/fonsp/Pluto.jl/blob/de6a312b09d09081d549681d05d286033d894ce7/src/precompile.jl#L80-L81

src/Configuration.jl
sjkelly: I see. I will drop this commit, since the benefit is marginal, mostly on precompilation time.

As an aside, I think there could be some better inference and shorter precompilation time if we remove more Union splitting. But since this is breaking I think it would be good to attempt it all at once, and see what kind of improvements are possible and if such a change is worth is.

src/analysis/Parse.jl
fonsp: I meant more that this is a form of documentation for developers: I want to communicate that this function returns a `Vector{<:Integer}`, to make it more clear what the function does.

Just like how `find_first_vowel(::String)` would be ambiguous, but `find_first_vowel(::String)::Char` is clear.

fonsp: I think PrecompileSignatures works with the *input signatures* of functions, Rik will know more :)

fonsp (approved):
Thanks!! Can you fix the merge conflicts?

sjkelly:
Should be fixed! Thanks!

PR #2499, opened 2023-03-11
Julia 1.9: Ensure that precompile caches match by running `Pkg.precompile()` in an isolated Julia process

fonsp:
Since Julia 1.9, precompilation has gotten more advanced. There is a new command line flag for precompilation: https://github.com/fonsp/Pluto.jl/pull/2456, but there is a problem with our current system:

Because we run all Pkg operations on the server process, precompilation will use the `--pkgimages` setting of the server process, regardless of what the notebook process will use. If the two don't match, then the server will generate precomiled images for one setting, and then the notebook will miss the cache and precompile everything again with the other setting. 

Since 1.9, you also see [this message](https://user-images.githubusercontent.com/6933510/224487322-a1966954-1eb3-4933-a232-b1f57e828e94.png). What the message says is technically true, but the advice to restart Julia does not apply and can be confusing.

# Solutions

There is no Pkg API to control for which setting you want to precompile, it has to be the global setting of the Julia process. Our two options are:
- Run `Pkg.instantiate()` in the notebook process
- Run `Pkg.instantiate()` in a new Julia process spawned for this purpose.

I think the second option is cleaner, because it maintains the separation between Pkg operations and code evaluation. I can also imagine cases where the notebook process is not/should not be running, but you still want to setup the environment.

The downside is overhead: about 1 second for starting the process, and about 200MB additional memory usage.

**EDIT:** In this PR we now run `Pkg.instantiate(allow_autoprecomp=false)` in the server, and `Pkg.precompile()` in a new Julia process.

# TODO

- [x] Call the function
- [x] Test that the PR works: do we generate the right precompile files when the flags dont match
- [x] Make sure that command line arguments are not propagated https://github.com/JuliaLang/julia/pull/48425
- [x] Also use it for updating
- [x] This adds a ~1.5~ EDIT: 0.9 second overhead to every Pkg operation, including things like adding `import Dates` or removing that statement. Not too bad but might be nice to avoid. 
  - Correctness is more important than speed
  - Overhead is not too big, it's constant
  - Overhead happens at a place where people expect things to take some time (Pkg operations)
  - We can improve it in a separate PR


fonsp:
I added a test and the test fails on `main` 👍

src/packages/PkgCompat.jl
```diff
@@ -227,13 +227,12 @@ end
...
 ###
 
-# ⚠️✅ Internal API with fallback
-function instantiate(ctx; update_registry::Bool)
-	@static if hasmethod(Pkg.instantiate, Tuple{}, (:update_registry,))
-		Pkg.instantiate(ctx; update_registry)
-	else
-		Pkg.instantiate(ctx)
-	end
+# ⚠️ Internal API
```
fonsp, line 230: (This never actually had a fallback)

PR #2500, opened 2023-03-11
🎡 Fancyprint Pkg logs

fonsp:
Enable TTY fancy printing features for TTY logs: loading bars and spinners in terminal UI, notably during precompilation.

# Before

<details>
<summary>Details</summary>

https://user-images.githubusercontent.com/6933510/224378733-e1892b51-5eae-4d89-821c-9716dc9d8f55.mov

</details>

Note that during precompilation, packages are printed with a checkmark when they are done, but you don't see the fancy spinners and **progress bar**. (This is also what you see in GitHub CI logs.)

# After

https://user-images.githubusercontent.com/6933510/224486930-1adb62e5-9d80-4897-abdf-43f73ba5a4a5.mov

Now we see the fancy spinners and **progress bar**. 

# Difficult

This turned out to be difficult to implement because:

## Existing lib
There is no existing Julia lib to handle ANSI control characters, and only one option for JS: xterm.js, which I don't want to use because of its poor scrolling and dpi scaling. A Julia lib would also be better because of the large buffer (see next point). Syncing the input buffer on each update would require an extra layer of optimization, like [`AppendonlyMarker` for the large vector of logs](https://github.com/fonsp/Pluto.jl/commit/0e5f8cebbe893c399c29f85858fdd732de72d777).

I ended up implementing a minimal TTY emulator in Julia myself ~and it has some glitches~. (I think I fixed the glitches) This has no tests currently, but it is in a `try catch` block to make it safe. It might be nice to publish this? Maybe some Terminal UI packages like https://github.com/kdheepak/TerminalUserInterfaces.jl could use and improve it?

## Large buffer
Because the spinners update every 100ms or so, the total captured buffer becomes quite large. (Every spinner update works by moving the cursor and reprinting the list of precompiling packages.) 

The simplest solution would be a function `raw_buffer_content::String -> emulated_terminal_content::String` that we run on the buffer content on every update. But with a large buffer, this can take 5ms, which will end up consuming a larger and larger part of your CPU. 

So I had to solve this by storing the *emulated terminal state*, and incrementally updating it whenever new data is available. This works but adds some complexity.

## Misc
- Pkg just checks `io isa Base.TTY` to decide whether to prettyprint. I did some type piracy to override that and allow prettyprinting to an `IOBuffer`.
- I realized that for https://github.com/fonsp/Pluto.jl/pull/2499 we will run precompilation in a separate process, which might make capturing IO and overriding the `can_prettyprint` logic more difficult. EDIT: this was pretty easy!


fonsp:
I was concerned that this system might overflow the websocket, or the queue of messages to send to clients, when the connection is poor. I tested this (with `simulated_lag`) and I verified the code: this is not a problem! It "throttles" itself automatically, see my latest commit for explaination.

PR #2501, opened 2023-03-11
🦆 Fix #2416 maybe by avoiding race condition

### LINKED ISSUE #2416: Crash with AppendonlyMarker
My notebook crashed today running some cells, (triggered by the file watching cell)

Notebook: https://gist.github.com/fonsp/cbb184f87323184afbbeb372eb5727e5

```
┌ Error: Failed to relay self-update
│   exception =
│    Not really supposed to diff AppendonlyMarker with the original being longer than the next version (you know, 'append only' and al)
│    Stacktrace:
│      [1] diff(a::Pluto.FirebaseyUtils.AppendonlyMarker{Dict{String, Any}}, b::Pluto.FirebaseyUtils.AppendonlyMarker{Dict{String, Any}})
│        @ Pluto.FirebaseyUtils ~/.julia/packages/Pluto/JxOQL/src/webserver/FirebaseyUtils.jl:185
│      [2] diff(o1::Dict{String, Any}, o2::Dict{String, Any}) (repeats 3 times)
│        @ Pluto.Firebasey ~/.julia/packages/Pluto/JxOQL/src/webserver/Firebasey.jl:378
│      [3] send_notebook_changes!(🙋::Pluto.ClientRequest; commentary::Nothing, skip_send::Bool)
│        @ Pluto ~/.julia/packages/Pluto/JxOQL/src/webserver/Dynamic.jl:183
│      [4] send_notebook_changes!
│        @ ~/.julia/packages/Pluto/JxOQL/src/webserver/Dynamic.jl:178 [inlined]
│      [5] #242
│        @ ~/.julia/packages/Pluto/JxOQL/src/evaluation/Run.jl:137 [inlined]
│      [6] #32
│        @ ~/.julia/packages/Pluto/JxOQL/src/evaluation/Throttled.jl:23 [inlined]
│      [7] lock(f::Pluto.var"#32#35"{Pluto.var"#242#248"{Pluto.ServerSession, Pluto.Notebook}, Base.RefValue{Bool}}, l::ReentrantLock)
│        @ Base ./lock.jl:185
│      [8] flush
│        @ ~/.julia/packages/Pluto/JxOQL/src/evaluation/Throttled.jl:21 [inlined]
│      [9] throttled_f
│        @ ~/.julia/packages/Pluto/JxOQL/src/evaluation/Throttled.jl:41 [inlined]
│     [10] run_reactive_core!(session::Pluto.ServerSession, notebook::Pluto.Notebook, old_topology::Pluto.NotebookTopology, new_topology::Pluto.NotebookTopology, roots::Vector{Pluto.Cell}; save::Bool, deletion_hook::typeof(Pluto.WorkspaceManager.move_vars), user_requested_run::Bool, already_run::Vector{Pluto.Cell}, bond_value_pairs::Base.Iterators.Zip{Tuple{Vector{Symbol}, Vector{Any}}})
│        @ Pluto ~/.julia/packages/Pluto/JxOQL/src/evaluation/Run.jl:167
│     [11] #237
│        @ ~/.julia/packages/Pluto/JxOQL/src/evaluation/Run.jl:22 [inlined]
│     [12] withtoken(f::Pluto.var"#237#238"{Bool, typeof(Pluto.WorkspaceManager.move_vars), Bool, Base.Iterators.Zip{Tuple{Vector{Symbol}, Vector{Any}}}, Pluto.ServerSession, Pluto.Notebook, Pluto.NotebookTopology, Pluto.NotebookTopology, Vector{Pluto.Cell}}, token::Pluto.Token)
│        @ Pluto ~/.julia/packages/Pluto/JxOQL/src/evaluation/Tokens.jl:19
│     [13] #run_reactive!#236
│        @ ~/.julia/packages/Pluto/JxOQL/src/evaluation/Run.jl:21 [inlined]
│     [14] start_relaying_self_updates(::Tuple{Pluto.ServerSession, Pluto.Notebook}, run_channel::Distributed.RemoteChannel{Channel{Base.UUID}})
│        @ Pluto.WorkspaceManager ~/.julia/packages/Pluto/JxOQL/src/evaluation/WorkspaceManager.jl:143
│     [15] (::Pluto.WorkspaceManager.var"#4#6"{Pluto.Notebook, Pluto.ServerSession, Distributed.RemoteChannel{Channel{Base.UUID}}})()
│        @ Pluto.WorkspaceManager ./task.jl:484
└ @ Pluto.WorkspaceManager ~/.julia/packages/Pluto/JxOQL/src/evaluation/WorkspaceManager.jl:148
```

fonsp:
Be sure to read this PR with whitespace changes disabled.

Fix #2416 which looks like it was a race condition: multiple updates were happening at the same time, and there was a race condition around updating `current_state_for_clients`. Our code is not multithreaded, but it does run in many async `Task`s.

Since I can't reproduce #2416, this is some guesswork. "Fixed" it by:
- adding a lock around the logic that updates `current_state_for_clients`. 
- instead of doing the blocking websocket IO during the update loop (inside the lock), I first compute all updates, (then release the lock), and then send them in a single batch afterwards.

Questions:
- I used `ReentrantLock` but I don't know what it is. 
	- Is this the right choice? what about `Threads.SpinLock`? If we want to multithread in the future, will this also lock among threads?
	- Is there a performance cost?
	- Is there a risk of deadlock?

### DIFF
--- a/src/evaluation/Run.jl
+++ b/src/evaluation/Run.jl
@@ -177,7 +177,7 @@ function run_reactive_core!(
         cell.running = true
         # Important to not use empty! here because AppendonlyMarker requires a new array identity.
         # Eventually we could even make AppendonlyArray to enforce this but idk if it's worth it. yadiyadi.
-        cell.logs = []
+        cell.logs = Vector{Dict{String,Any}}()
         send_notebook_changes_throttled()
 
         if any_interrupted || notebook.wants_to_interrupt || !will_run_code(notebook)
--- a/src/webserver/Dynamic.jl
+++ b/src/webserver/Dynamic.jl
@@ -173,30 +173,39 @@ end
 For each connected client, we keep a copy of their current state. This way we know exactly which updates to send when the server-side state changes.
 """
 const current_state_for_clients = WeakKeyDict{ClientSession,Any}()
+const current_state_for_clients_lock = ReentrantLock()
 
 """
 Update the local state of all clients connected to this notebook.
 """
 function send_notebook_changes!(🙋::ClientRequest; commentary::Any=nothing, skip_send::Bool=false)
-    notebook_dict = notebook_to_js(🙋.notebook)
-    for (_, client) in 🙋.session.connected_clients
-        if client.connected_notebook !== nothing && client.connected_notebook.notebook_id == 🙋.notebook.notebook_id
-            current_dict = get(current_state_for_clients, client, :empty)
-            patches = Firebasey.diff(current_dict, notebook_dict)
-            patches_as_dicts::Array{Dict} = Firebasey._convert(Array{Dict}, patches)
-            current_state_for_clients[client] = deep_enough_copy(notebook_dict)
+    const outbox = Set{Tuple{ClientSession,UpdateMessage}}()
+    
+    lock(current_state_for_clients_lock) do
+        notebook_dict = notebook_to_js(🙋.notebook)
+        for (_, client) in 🙋.session.connected_clients
+            if client.connected_notebook !== nothing && client.connected_notebook.notebook_id == 🙋.notebook.notebook_id
+                current_dict = get(current_state_for_clients, client, :empty)
+                patches = Firebasey.diff(current_dict, notebook_dict)
+                patches_as_dicts::Array{Dict} = Firebasey._convert(Array{Dict}, patches)
+                current_state_for_clients[client] = deep_enough_copy(notebook_dict)
 
-            # Make sure we do send a confirmation to the client who made the request, even without changes
-            is_response = 🙋.initiator !== nothing && client == 🙋.initiator.client
+                # Make sure we do send a confirmation to the client who made the request, even without changes
+                is_response = 🙋.initiator !== nothing && client == 🙋.initiator.client
 
-            if !skip_send && (!isempty(patches) || is_response)
-                response = Dict(
-                    :patches => patches_as_dicts,
-                    :response => is_response ? commentary : nothing
-                )
-                putclientupdates!(client, UpdateMessage(:notebook_diff, response, 🙋.notebook, nothing, 🙋.initiator))
+                if !skip_send && (!isempty(patches) || is_response)
+                    response = Dict(
+                        :patches => patches_as_dicts,
+                        :response => is_response ? commentary : nothing
+                    )
+                    push!(outbox, (client, UpdateMessage(:notebook_diff, response, 🙋.notebook, nothing, 🙋.initiator)))
+                end
             end
         end
+    end
+    
+    for (client, msg) in outbox
+        putclientupdates!(client, msg)
     end
     try_event_call(🙋.session, FileEditEvent(🙋.notebook))
 end
--- a/src/webserver/Dynamic.jl
+++ b/src/webserver/Dynamic.jl
@@ -179,7 +179,7 @@ const current_state_for_clients_lock = ReentrantLock()
 Update the local state of all clients connected to this notebook.
 """
 function send_notebook_changes!(🙋::ClientRequest; commentary::Any=nothing, skip_send::Bool=false)
-    const outbox = Set{Tuple{ClientSession,UpdateMessage}}()
+    outbox = Set{Tuple{ClientSession,UpdateMessage}}()
     
     lock(current_state_for_clients_lock) do
         notebook_dict = notebook_to_js(🙋.notebook)


fonsp:
A bit more info about what I think the race condition is:

Two parallel threads both call https://github.com/fonsp/Pluto.jl/blob/6950747e6670da4f9c6054c4ed9025ad0de2a6c3/src/webserver/Dynamic.jl#L180 . This function computes the current state (first line), and then loops through connected clients, compares this with their last known state, updates their last known state, and sends them the diff.

The following happened in order:

- Cell x has no logs
- There is only one connected client.
- A run is triggered by PlutoHooks. This starts an async Task 1 on the server to run the cell and eventually execute `send_notebook_changes!` (you can see this in the stack trace of the issue).
- Task 1 calls https://github.com/fonsp/Pluto.jl/blob/6950747e6670da4f9c6054c4ed9025ad0de2a6c3/src/webserver/Dynamic.jl#L181 . This state has no logs stored for cell x. The `AppendonlyMarker` has `length_at_time_of_creation=0`.
- A log comes in (this is async). This Task 2 calls `send_notebook_changes!`. It happens at the same time because the log comes from the cell that was just executed.
- Task 2 calls https://github.com/fonsp/Pluto.jl/blob/6950747e6670da4f9c6054c4ed9025ad0de2a6c3/src/webserver/Dynamic.jl#L181 . This state has a single log for cell x. The `AppendonlyMarker` has `length_at_time_of_creation=1`, with the same underlying `Vector`.
- Task 2 executes `send_notebook_changes!` until completion, updating `current_state_for_clients_lock` with its state.
- Task 1 gets control again, and executes `send_notebook_changes!` until completion. For the current state of the client it finds the state computed by Thread 2.
- When diffing, the "old" state has the `AppendonlyMarker` with `length_at_time_of_creation=1`, while the current state has the same underlying vector but with `length_at_time_of_creation=0`.
- Error!

tecosaur:
> I used ReentrantLock but I don't know what it is.

My understanding is you basically want to use a `SpinLock` when you're looking at multiple threads wanting to briefly touch something, where it's desirable for the waiting thread to "spin in place" constantly checking if it can go. If in doubt as to how brief the delay will be, or if anything like IO is involved, go for a `ReentrantLock`. Note that you cannot deadlock a `ReentrantLock` within a single task.

fonsp:
Thanks! It sounds like SpinLock and ReentrantLock would both be fine in this case, and ReentrantLock is a good choice?

Does a lock have a performance cost in general?

tecosaur:
Glancing at the patch itself, I'd be inclined to just go with `ReentrantLock` and see if any issues pop up (likely none).

Locks of course have some cost, but not much, and if you need them you need them...

fonsp:
Looks like the overhead is very small:

```julia
julia> l = ReentrantLock()
ReentrantLock(nothing, 0x00000000, 0x00, Base.GenericCondition{Base.Threads.SpinLock}(Base.InvasiveLinkedList{Task}(nothing, nothing), Base.Threads.SpinLock(0)), (0, 0, 0))

julia> using BenchmarkTools

julia> x = rand(10)
10-element Vector{Float64}:
 0.19438967322046696
 0.507077497607146
 0.26051489368260183
 0.5571815732377596
 0.3581092258664019
 0.024027844514510677
 0.760799509399507
 0.49865164009155216
 0.8988749180195333
 0.48456254513387287

julia> @benchmark sum(sqrt.(x))
BenchmarkTools.Trial: 10000 samples with 194 evaluations.
 Range (min … max):  489.691 ns …   7.365 μs  ┊ GC (min … max): 0.00% … 92.94%
 Time  (median):     496.995 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   509.527 ns ± 201.318 ns  ┊ GC (mean ± σ):  1.17% ±  2.77%

   ▃▇██▇▆▄▃▂▂▃▃▄▄▃▃▂▂▂▂▁▁                                       ▂
  ▇██████████████████████████████▇█▇▇▇▆▇▆▅▅▅▂▅▄▅▇▇▆▆▃▂▄▅▆▄▃▄▄▅▅ █
  490 ns        Histogram: log(frequency) by time        581 ns <

 Memory estimate: 224 bytes, allocs estimate: 5.

julia> @benchmark lock(() -> sum(sqrt.(x)), l)
BenchmarkTools.Trial: 10000 samples with 185 evaluations.
 Range (min … max):  564.411 ns …   7.718 μs  ┊ GC (min … max): 0.00% … 92.22%
 Time  (median):     571.622 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   583.600 ns ± 207.757 ns  ┊ GC (mean ± σ):  1.06% ±  2.76%

   ▅▇█▇▅▃▂▂▃▄▃▃▂▂▂▁▁                                            ▂
  ████████████████████▇▇▆▇▇▅▆▅▅▆▅▅▆▆▄▆▆▅▅▅▆▆▅▄▆▅▄▅▅▃▅▄▄▅▅▂▄▂▄▄▄ █
  564 ns        Histogram: log(frequency) by time        682 ns <

 Memory estimate: 224 bytes, allocs estimate: 5.

julia> 
```

tecosaur:
> Looks like the overhead is very small

It should be :stuck_out_tongue: fast locks are pretty important for multithreading.

Issue #2488, opened 2023-03-02
Precompilation when updating packages


aplavin:
Pluto doesn't precompile packages when they are updated in a notebook. This leads to very long `using` time, because there precompilation happens sequentially.
I'm not sure why is this the case, while Julia REPL of course runs precompile automatically on `]update`.
Would be nice to run actual parallel precompilation in Pluto as well!

The only relevant line of code I found in Pluto is this:
https://github.com/fonsp/Pluto.jl/blob/776cbf7cb37770dbc60bc037737ed143a1fa0f58/src/runner/Loader.jl#L32
but not sure if really related.

PR #2504, opened 2023-03-13
Instantiate after auto fixed resolve

fonsp:
Fixes the same problem as #2488 but with auto-fixed resolve (when you open a notebook with a corrupt manifest, we will automatically fix it for you (e.g. #2294 #2298)). After this PR, when we deploy an auto-fix, we also make sure that the notebook is instantiated (hence precompiled in parallel) afterwards.


fonsp:
Hmm, this does not seem to fix the issue i was facing:

Opening the Plots.jl featured notebook and running it on Julia 1.9, it will delete the manifest during the auto-fixed instantiate (I thought it was during resolve), and then succeed. But then the cell `using Plots` takes many minutes, because it is precompiling everything again. Might be related to https://github.com/JuliaLang/Pkg.jl/issues/3403

fonsp:
My issue was caused by https://github.com/JuliaLang/Pkg.jl/issues/3407

fonsp:
Tracking this other issue in #2508

PR #2516, opened 2023-03-30
Hide server start and stop logs from HTTP.jl

fonsp:


fonsp:
Will merge this after https://github.com/JuliaWeb/HTTP.jl/pull/1022 to make sure that we are no using undocumented API that might be removed in a future release

PR #2523, opened 2023-04-09
Add binder progress and logs to status tab

fonsp:
Screenshot (1)

I used the `backend_launch_phase` for the subtasks of the backend launch, and I capture the JSON logs from the streaming mybinder.org API as logs in the little terminal, similar to #2498

### DIFF
--- a/frontend/common/Binder.js
+++ b/frontend/common/Binder.js
@@ -6,6 +6,7 @@ export const BackendLaunchPhase = {
     wait_for_user: 0,
     requesting: 0.4,
     created: 0.6,
+    responded: 0.7,
     notebook_running: 0.9,
     ready: 1.0,
 }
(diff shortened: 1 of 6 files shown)


fonsp:
@pankgeorg If you set the backend_launch_phase for JH integration then this feature should already work. It might be nice to include the logs if they are easily available to the frontend, use https://github.com/fonsp/Pluto.jl/pull/2523/files#diff-c0a453bced0e2ce81412403f646a1e2bbfd19300fab615ce77b039a417c82dceR109-R113 to set the value of the terminal. You could also use this to print something like

```
Launching a 4-core 8GB computer on thisandthat account. Go to asdf.com/asdf to see all your job launches!
```

fonsp:
Updated PR description

PR #2526, opened 2023-04-09
integrate JuliaSyntax.jl

Pangoraw:
image

 - [x] Investigate time to first parse (seems negligable on v1.9+ https://github.com/JuliaLang/JuliaSyntax.jl/pull/213)
 - [x] Prevent roundtrip to notebook process by making `ParseError <: ReactivityError` and using `relay_reactivity_error!`.
 - [x] Make each diagnostic clickable and selects the problematic span.
 - [x] Find and resolve the CM bug which lead to the update being reverted.
 - [x] Cleanup the javascript (replace custom events with hooks?)
 - [x] z-index for lint boxes
 - [ ] parse (debounced) as we type and not just on submit?
 - [ ] Map highlighted ranges up to date with content?


Pangoraw:
https://github.com/JuliaLang/julia/pull/46372 has been merged :confetti_ball: and JuliaSyntax will become the default parser in 1.10 ! I will come back to this PR and probably drop the JuliaSyntax dependency (make it work only on >=1.10).

Pangoraw:
The last problem is z-index positioning of the tooltip which is wrong when the cell is not in focus or with the run button.

image

I tried setting the `z-index` for `.cm-tooltip` but this does not seems to work. Maybe this can be fixed by [changing the positioning from `fixed` to `absolute`](https://codemirror.net/docs/ref/#view.tooltips^config).

The same problem happens with autocompletion tooltip but it is less of a problem (notice the run button over):
image


Pinging CSS Expert @pankgeorg, you may have an idea on how to fix that ? :smiley:

Pangoraw:
One future improvement is identifying error messages we can propose quick fixes for (like "replace `else if` with `elseif`) and add fix actions to the tooltip:

quick_fix_example

I am going to merge once the CI pass as it only affects 1.10 users for now (and slightly improves the display of syntax errors for VERSION < 1.10).

PR #2536, opened 2023-04-25
Migrate from SnoopPrecompile to PrecompileTools

timholy:
This pull request migrates the package from [SnoopPrecompile](https://github.com/timholy/SnoopCompile.jl/tree/master/SnoopPrecompile) to [PrecompileTools](https://github.com/JuliaLang/PrecompileTools.jl).
PrecompileTools is **nearly a drop-in replacement** except that there are **changes in naming and how developers locally disable precompilation** (to make their development workflow more efficient). These changes are described in [PrecompileTool's enhanced documentation](https://julialang.github.io/PrecompileTools.jl/stable/), which also includes instructions for users on how to set up custom "Startup" packages, handling precompilation tasks that are not amenable to workloads, and tips for troubleshooting.

Why the new package? It meets several goals:

- The name "SnoopPrecompile" was easily confused with "SnoopCompile," a package designed for *analyzing* rather than *enacting* precompilation.
- SnoopPrecompile/PrecompileTools has become (directly or indirectly) a dependency for much of the Julia ecosystem, a trend that seems likely to grow with time. It makes sense to host it in a more central location than one developer's personal account.
- As Julia's own stdlibs migrate to become independently updateable (true for DelimitedFiles in Julia 1.9, with others anticipated for Julia 1.10), several of them would like to use PrecompileTools for high-quality precompilation. That requires making PrecompileTools its own "upgradable stdlib."
- We wanted to change the [use of Preferences](https://github.com/timholy/SnoopCompile.jl/issues/356) to make packages more independent of one another. Since this would have been a breaking change, it seemed like a good opportunity to fix other issues, too.

For more information and discussion, see this [discourse post](https://discourse.julialang.org/t/ann-snoopprecompile-precompiletools/97882).

### DIFF
--- a/src/packages/ANSIEmulation.jl
+++ b/src/packages/ANSIEmulation.jl
@@ -424,7 +424,7 @@ uuid = "4536629a-c528-5b80-bd46-f80d51c5b363"
 version = "0.3.21+4"
 
 [[deps.Parsers]]
-deps = ["Dates", "SnoopPrecompile"]
+deps = ["Dates", "PrecompileTools"]
 git-tree-sha1 = "478ac6c952fddd4399e71d4779797c538d0ff2bf"
 uuid = "69de0a69-1ddd-5017-9359-2bf0b02dc9f0"
 version = "2.5.8"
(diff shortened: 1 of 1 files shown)


timholy:
Yes, these are arriving via an automated script, sorry for the noise. I'll close.

Pangoraw:
Dependent on https://github.com/JuliaData/Parsers.jl/pull/164 and then the notebook Manifest can be updated when Parsers.jl is updated.

fonsp:
Thanks for your work on PrecompileTools, Tim!

That notebook manifest isn't used by users of the Pluto package (it is `include`d), it's only used when a Pluto developer opens that notebook using Pluto.

Issue #2554, opened 2023-05-02
MethodError: no method matching pretty_stackcall


Keno:
The attached Pluto notebook with a single cell containing `Main.foo` spits out the following error:

```
On worker 3:

MethodError: no method matching pretty_stackcall(::Base.StackTraces.StackFrame, ::Module)

Closest candidates are:

pretty_stackcall(::Base.StackTraces.StackFrame, !Matched::Nothing)

@ Pluto ~/.julia/packages/Pluto/mULuw/src/runner/PlutoRunner.jl:1075

pretty_stackcall(::Base.StackTraces.StackFrame, !Matched::Core.CodeInfo)

@ Pluto ~/.julia/packages/Pluto/mULuw/src/runner/PlutoRunner.jl:1087

pretty_stackcall(::Base.StackTraces.StackFrame, !Matched::Core.MethodInstance)

@ Pluto ~/.julia/packages/Pluto/mULuw/src/runner/PlutoRunner.jl:1091

Stacktrace:

[1] #67

@ ~/.julia/packages/Pluto/mULuw/src/runner/PlutoRunner.jl:1045

[2] iterate

@ ./generator.jl:47 [inlined]

[3] collect_to!

@ ./array.jl:885

[4] collect_to_with_first!

@ ./array.jl:863 [inlined]

[5] _collect

@ ./array.jl:857

[6] collect_similar

@ ./array.jl:756

[7] map

@ ./abstractarray.jl:3267

[8] #format_output#63

@ ~/.julia/packages/Pluto/mULuw/src/runner/PlutoRunner.jl:1044

[9] format_output

@ ~/.julia/packages/Pluto/mULuw/src/runner/PlutoRunner.jl:1026 [inlined]

[10] #53

@ ~/.julia/packages/Pluto/mULuw/src/runner/PlutoRunner.jl:911 [inlined]

[11] #with_io_to_logs#119

@ ~/.julia/packages/Pluto/mULuw/src/runner/PlutoRunner.jl:2468

[12] #123

@ ~/.julia/packages/Pluto/mULuw/src/runner/PlutoRunner.jl:2492

[13] with_logstate

@ ./logging.jl:515

[14] with_logger

@ ./logging.jl:627 [inlined]

[15] #with_logger_and_io_to_logs#122

@ ~/.julia/packages/Pluto/mULuw/src/runner/PlutoRunner.jl:2491 [inlined]

[16] #formatted_result_of#50

[...]

PR #2560, opened 2023-05-15
add `pretty_stackcall(::Base.StackFrame, ::Module)`

pankgeorg:
Patch https://github.com/fonsp/Pluto.jl/issues/2554

tested on 

```
Version 1.10.0-DEV.1293 (2023-05-14)
Commit 344f1f5a9cd (0 days old master)
```

Would be nice to understand what changed in julia before merging

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -1096,6 +1096,10 @@ function pretty_stackcall(frame::Base.StackFrame, linfo::Core.MethodInstance)
     end
 end
 
+function pretty_stackcall(frame::Base.StackFrame, linfo::Module)
+    sprint(Base.show, linfo)
+end
+
 "Return a `(String, Any)` tuple containing function output as the second entry."
 function show_richest_withreturned(context::IOContext, @nospecialize(args))
     buffer = IOBuffer(; sizehint=0)


Pangoraw:
The corresponding Julia change is https://github.com/JuliaLang/julia/pull/41099.

https://github.com/JuliaLang/julia/commit/2cd0149fc6034d8785b66a1e0e2cadf26c4d1ca3#diff-cf1a3652a2d8ed35925fb266a1c1d0018d5c393819bb67784e7a5f005e54ce65R55-R57

It seems, we also need to add `pretty_stackcall(::Base.StackFrame, ::Method)`.

src/runner/PlutoRunner.jl
```diff
@@ -1096,6 +1096,10 @@ function pretty_stackcall(frame::Base.StackFrame, linfo::Core.MethodInstance)
     end
 end
 
+function pretty_stackcall(frame::Base.StackFrame, linfo::Union{Module,Method})
```
Pangoraw, line 1099: To be consistent with the case where `linfo.def isa Method` in `pretty_stackcall(::Base.StackFrame, ::MethodInstance)` maybe?

```suggestion
function pretty_stackcall(frame::Base.StackFrame, linfo::Method)
    sprint(Base.show_tuple_as_call, linfo.name, linfo.specTypes)
end

function pretty_stackcall(frame::Base.StackFrame, linfo::Module)
```

Pangoraw: oops, `Method` does not have specTypes :grimacing:

src/runner/PlutoRunner.jl
```diff
@@ -1096,6 +1096,14 @@ function pretty_stackcall(frame::Base.StackFrame, linfo::Core.MethodInstance)
     end
 end
 
+function pretty_stackcall(frame::Base.StackFrame, linfo::Method)
+    sprint(Base.show_tuple_as_call, linfo.name, linfo.specTypes)
```
Pangoraw, line 1100:
```suggestion
    sprint(Base.show_tuple_as_call, linfo.name, linfo.sig)
```

pankgeorg: cool! Should we do the same for the other block a few lines above?

Do we expect `Core.MethodInstance`.`specTypes` to equal `Core.MethodInstance`.`def`.`sig`?

image

Pangoraw: The `MethodInstance.specTypes` is more specialized than `Method.sig`, one `Method` can have multiple `MethodInstance`. For example:

```julia
julia> method = first(methods(+))
+(x::T, y::T) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}
     @ Base int.jl:87

julia> method_instance = first(method.specializations)
MethodInstance for +(::Int64, ::Int64)

julia> Base.show_tuple_as_call(stdout, method.name, method.sig) # abstract signature
+(::T, ::T) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}

julia> Base.show_tuple_as_call(stdout, method.name, method_instance.specTypes) # one specialization (::Int64, ::Int64)
+(::Int64, ::Int64)
```
so when we have the MethodInstance using specTypes gives more information.

fonsp:
I'm surprised that this wasn't caught by our tests and nanosoldier, because we test stack traces... Is there something specific about the `Main.foo` example?

fonsp:
And thanks Panagiotis!!

fonsp (approved):

fonsp:
@pankgeorg can you run the tests before this PR on Julia 1.10? (and play some guitar because it takes 30 minutes)

pankgeorg:
> can you run the tests before this PR on Julia 1.10? (and play some guitar because it takes 30 minutes)

yes, doing

> I'm surprised that this wasn't caught by our tests and nanosoldier, because we test stack traces... Is there something specific about the `Main.foo` example?

Well, does anything run Julia 1.10? It's only on nightly build, not yet available in our CI or juliaup.

pankgeorg:
Tests break on 1.10:

image

I think the perpetrator is indeed this, but to make sure, I'll rerun on main
image
image

goedman:
It's working fine now on the latest update of Pluto(v0.16.26)!

`Main.foo` was just a good, short example. I ran into it in many other programming errors.

Thanks a lot!!!!!

PR #2563, opened 2023-05-15
👀 Open notebooks without running

fonsp:
After this PR, you can open a notebook without running it! You can even make edits (change code, add/delete cells) that are saved to the file.

When you Shift+Enter, cells still won't run: you need to click the big "Start notebook", and then it will be as if you just started the notebook in old Pluto: Pkg initializes and all cells will run.

https://github.com/fonsp/Pluto.jl/assets/6933510/fec23b79-a162-43e8-a474-b04b404d8dc5

Implement #259

# TODO

- [x] Open notebook without running
- [x] Also disable Pkg instantiate/etc without permission
- [x] New Pkg popup message
- [x] Check all Dynamic.jl entry points
- [x] See changes when editing md-only cells
- [x] Make it more clear that you need to run the notebook
- [x] Move warning when opening from url to the run button
- [x] Opening a URL, then closing the session, then starting the session does not show a warning.
- [x] Configuration to disable warning (for binder etc)
- [x] Should the notebook stay in your session? Maybe reword the "shutdown?" prompt in the main menu.
  - [x] Keep the prompt because you are used to press enter.
- [x] What happens when you add/remove packages while waiting for permission?
  - [x] Adding packages works
  - [x] If you add `Pkg.activate()` statically, then start the notebook, you immediately get a "Restart required"
  - [x] If you remove a package import statically, then start the notebook, you immediately get a "Restart recommended" message
- [x] Remove old config `run_notebook_on_load` or rename to `run_all_cells_on_start` or sth
- [x] Tests
- [x] Fix existing tests
- [x] Disable scripts (and more? maybe [sanitize HTML](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API))
- [x] Switch to DOMPurify: https://github.com/mozilla/sanitizer-polyfill/pull/234#issuecomment-1551374733
- [x] "Start notebook in background" no longer working
- [x] Test notebookupload
- [x] Test modifying in Safe mode
- [x] Test fallback samples
- [x] Test "start notebook in background" for risky source
- [x] What should happen to the "run in background" buttons?
- [x] Make cells with hidden code more visible.

# FUTURE TODO

- [ ] From URL: offer to run with Binder/JuliaHub instead?
- [ ] Review: walk through all non-md cells one by one?
- [ ] Try to preserve the cell order that was used in the saved file?


fonsp:
Opening a notebook from a URL:

https://github.com/fonsp/Pluto.jl/assets/6933510/3a7b6d14-1b6b-4faf-bccd-d21cebcebc8b

BioTurboNick:
This feature looks great! A little encouragement to finish it, and also offer help if possible :-)

Issue #2567, opened 2023-05-21
Pluto.run(notebook="mynotebook.jl")  fails to start notebook on 0.19.26


j-fu:
Hi,
when I give the notebook name in the Pluto.run command, I get a "not yet authenticated" page instead of the notebook. This happens since 0.19.26.

pankgeorg:
Should be addressed in https://github.com/fonsp/Pluto.jl/pull/2569/files

PR #2569, opened 2023-05-22
Fix: #2567 by echoing the secret from the request URL

pankgeorg:
The secret is not taken from the session settings; rather we paste whatever existed in the initial request. So this should be as secure as the initial one.


fonsp:
Thanks @pankgeorg ! I found a shorter solution in https://github.com/fonsp/Pluto.jl/commit/20515dd46678a49ca90e042fcfa3eab1e5c8e162 by reusing the existing cookie-setting logic

PR #2573, opened 2023-05-26
Feat: serialize `Pluto`.`activate_notebook_environment` to allow running Pluto notebooks as scripts

pankgeorg:
Towards: https://github.com/fonsp/Pluto.jl/discussions/2572#discussioncomment-6008940

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -2105,10 +2105,12 @@ end"""
 
 # At some point it would be nice to move
 # activate_notebook_environment either to Pkg or PlutoPkg.jl
-const script_pkg = """
+script_pkg(version) = """
 if !isdefined(Main, :PlutoRunner)
-    import Pluto
     import Pkg
+    Pkg.activate(;temp=true)
+    Pkg.add(Pkg.Types.PackageSpec(name="Pluto", version="$version"))
+    import Pluto
     Pluto.activate_notebook_environment(@__FILE__)
     Pkg.instantiate()
 end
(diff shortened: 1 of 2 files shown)


src/runner/PlutoRunner.jl
```diff
@@ -2103,6 +2103,16 @@ const fake_bind = """macro bind(def, element)
...
 end"""
 
+# At some point it would be nice to move
+# activate_notebook_environment either to Pkg or PlutoPkg.jl
+const script_pkg = """
+if !isdefined(Main, :PlutoRunner)
+    import Pluto
+    import Pkg
+    Pluto.activate_notebook_environment(@__FILE__)
+    Pkg.instantiate()
```
pankgeorg, line 2113:
```suggestion
    import Pkg
    Pkg.activate(;temp=true)
    Pkg.add("Pluto", rev=PLUTO_VERSION)
    import Pluto 
    Pluto.activate_notebook_environment(@__FILE__)
    Pkg.instantiate()
```

disberd:
Wouldn't this break packages that use notebooks as included files (unless they explicitly avoid using PlutoPkg).
I didn't try the PR but it would seem from the code that when `included` outside of Pluto this would create, activate and instantiate the notebook environment instead of the one of the `including` package

pankgeorg:
Yes that is a good point. We probably want something more sophisticated here.

pankgeorg:
Closing in favor of the new julia syntax for scripts with embedded Project/Manifest tomls

PR #2580, opened 2023-06-06
Call Pkg.upgrade_manifest() when updating packages

fonsp:
A bit annoying that `Pkg.upgrade_manifest()` throws an error when the manifest is already up to date... so i just put it in an empty try catch

untested


fonsp:
Moved to https://github.com/JuliaPluto/GracefulPkg.jl/pull/12

PR #2598, opened 2023-07-09
Trigger updating from file upon changes to folded state

torfjelde:
Currently, changing the folded state of a cell in the underlying file does not trigger an update in the notebook. This is because the folded state is not currently checked for changes.

This PR makes it so we also check for changes in the folded state:)


Pangoraw:
Thanks for the PR! In this case, I think adding cells where only the `code_folded` properties has changed to the `changed` array has the unwanted effect of re-running them which makes some tests fail. I opened #2602 as an alternative.

torfjelde:
Aaah yeah I didn't think of that. Great! Feel free to close this in favour of #2602:)

PR #2604, opened 2023-07-12
ExEx: Track function calls in assignment lhs.

### LINKED ISSUE #2595: Assigning to an object indexed with a function returns "not defined" error
Here is a simple Pluto notebook example of the issue: [test_func.jl](https://gist.github.com/azraq27/b85913584cd86e9f48929f2e3fac306b)

Briefly, I'm getting an error when I try to index an object using the return of a function, and then subsequently (in the same line) assigning a value to it (e.g., `test_array[simple_func(1)] = 10`). The error says that `simple_func` isn't defined, but it works in the other cells (e.g., `test_array[simple_func(1)]` returns the correct value). Also, running this file through the Julia interpreter doesn't produce this error.

I don't know the details of how Pluto parses the cells, but it seems like this assignment is taking place in a different context, which isn't inheriting the definition of the `simple_func` function.

Breaking this up into separate lines fixes the problem. For example, this functionally equivalent block works fine:

    begin
    	i = simple_func(1)
    	test_array[i] = 10
    end

Appreciate your work! Pluto is great!

Pangoraw:
Fixes #2595

### DIFF
--- a/src/analysis/ExpressionExplorer.jl
+++ b/src/analysis/ExpressionExplorer.jl
@@ -146,7 +146,7 @@ function get_assignees(ex::Expr)::FunctionName
             # e.g. (x, y) in the ex (x, y) = (1, 23)
             args = ex.args
         end
-        union!(Symbol[], Iterators.map(get_assignees, args)...)
+        mapfoldl(get_assignees, union!, args; init=Symbol[])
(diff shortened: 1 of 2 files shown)


fonsp:
🎉🎉🎉 Nice!!

PR #2607, opened 2023-07-16
💁📩 Support publish_to_js inside logs

fonsp:
Fix #2092

I had to change the dynamics of `publish_to_js`, but these changes will be done anyways by https://github.com/fonsp/Pluto.jl/pull/2162 : the published object will be stored during the render to JavaScript, not during the call of `publish_to_js`. (But when using it inside HypertextLiteral, this is the same. Actually it's hard to think of a case where it is not the same.)

PR #2608, opened 2023-07-16
Make `published_to_js` (now `publish_to_js`) official API through AbstractPlutoDingetjes.jl (v2)

disberd:
This resurrects #2162 and adds some front-end tests.

Second tests is currently commented out as it's broken due to #2092, but @fonsp had a very productive afternoon in Graz fixing that out in #2607

Works in combination with https://github.com/JuliaPluto/AbstractPlutoDingetjes.jl/pull/11


fonsp:
I think I broke embed_display in #2607 so lets fix that here

disberd:
What is broken in embed_display?
I tried a simple call in the output and in the logs but I can't find problems:
image

fonsp:
Yep looks like it still works!

fonsp (approved):

PR #2613, opened 2023-07-17
🚦Slider warning messages when requests encounter problems

ctrekker:
If a slider server is still starting and the user interacts with a slider, a HTTP 503 error is thrown and the user is not notified. This PR adds a pop-up which informs the user of the slider server's status.
sliders_loading

Otherwise, if the client expects a slider server but for whatever reason (user is offline, slider server crashed, etc.) the request fails, this message is shown:
image

Finally, a bug was fixed in this PR, where if the initial bondconnections request fails but subsequent staterequests succeed, the client sends faulty state patches. These changes force the client to continue fetching bondconnections until the request succeeds.

### DIFF
--- a/frontend/common/SliderServerClient.js
+++ b/frontend/common/SliderServerClient.js
@@ -8,6 +8,23 @@ const assert_response_ok = (/** @type {Response} */ r) => (r.ok ? r : Promise.re
 
 const actions_to_keep = ["get_published_object"]
 
+const get_start = (graph, v) => Object.values(graph).find((node) => Object.keys(node.downstream_cells_map).includes(v))?.cell_id
+const get_starts = (graph, vars) => new Set([...vars].map((v) => get_start(graph, v)))
+const recursive_dependencies = (graph, starts) => {
+    const deps = new Set(starts)
+    const ends = [...starts]
+    while (ends.length > 0) {
+        const node = ends.splice(0, 1)[0]
+        _.flatten(Object.values(graph[node].downstream_cells_map)).forEach((child) => {
+            if (!deps.has(child)) {
+                ends.push(child)
+                deps.add(child)
+            }
+        })
+    }
+    return deps
+}
+
 export const nothing_actions = ({ actions }) =>
     Object.fromEntries(
         Object.entries(actions).map(([k, v]) => [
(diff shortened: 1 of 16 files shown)


frontend/common/SliderServerClient.js
```diff
@@ -39,6 +40,23 @@ export const nothing_actions = ({ actions }) =>
...
+    const el = previously_set ? document.querySelector(`bond[def=${previously_set.values().next().value}]`) : null
+    if (e.status === 503) {
+        open_pluto_popup({
+            type: "warn",
+            body: "Sliders are still starting, check back later for interactivity!",
+            source_element: el,
+        })
+    } else {
+        open_pluto_popup({
+            type: "warn",
```
ctrekker, line 53: I'm thinking to possibly change this from `warn` to `error`

frontend/common/SliderServerClient.js
```diff
@@ -53,30 +71,33 @@ export const slider_server_actions = ({ setStatePromise, launch_params, actions,
...
-        )
-    })
+    let bond_connections = null
 
     const mybonds = {}
     const bonds_to_set = {
         current: new Set(),
     }
     const request_bond_response = debounced_promises(async () => {
+        if (!bond_connections) {
```
ctrekker, line 81: This part will continue to fetch bondconnections until they are received successfully

frontend/common/SliderServerClient.js
fonsp: I think it already looks a bit too scary... I see the sliderserver functionality more as a bonus that people get when reading the article, and when PSS fails it just means that there is no bonus today, not that there is something to get worried about. Also because it is not something that the user could fix.

frontend/common/SliderServerClient.js
```diff
@@ -53,30 +71,33 @@ export const slider_server_actions = ({ setStatePromise, launch_params, actions,
...
+                        state.slider_server.interactive = Object.keys(bond_connections).length > 0
+                    })
+                )
+            }
+        }
+
         const base = trailingslash(launch_params.slider_server_url)
         const hash = await notebookfile_hash
-        const graph = await bond_connections
+        const graph = bond_connections || {}
```
fonsp, line 100:
```suggestion
        const graph = bond_connections ?? {}
```

fonsp:
Nice! You should be able to see those typescript errors directly in your IDE, maybe check your setup. You could put the bugfix in a separate PR or a commit to main.

fonsp:
When no sliderserver url is given, we show bonds greyed out, and we show a message about running the notebook when you click on them. Maybe this UI could also be used when the sliderserver is not available for this notebook? Maybe with some extra info at the bottom of the popup?

If the PSS will be available soon for a notebook, I think it's nice to say "check back later" because people might actually do that, but if it doesn't seem like it will work in the next hour, we might as well give the "static notebook" UX, maybe with some extra info in a `<detail>` for the notebook author.

PR #2626, opened 2023-08-09
Test Julia 1.10

Pangoraw:
It seems that 1.10 is around the corner, let's test it!

Note: There is a problem with FuzzyCompletions.jl which should be adressed in:
 - https://github.com/JunoLab/FuzzyCompletions.jl/pull/14

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -34,7 +34,7 @@ jobs:
             fail-fast: false
             matrix:
                 # We test quite a lot of versions because we do some OS and version specific things unfortunately
-                julia-version: ["1.6", "1.7", "1.8", "1.9"] #, "nightly"]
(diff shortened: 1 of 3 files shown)


Pangoraw:
Tests fail because #2560 is not included in `Pluto@0.14.7`:

https://github.com/fonsp/Pluto.jl/blob/1d2c3038032108fff3a9c565eaa84609f1e75b4e/test/packages/Basic.jl#L482-L484

fonsp:
@Pangoraw can you check that https://github.com/fonsp/Pluto.jl/commit/394be9033e327e870bf0d56715938dc8d4b276e3 is correct?

Pangoraw:
oops, the previous check disabled the test for 1.9.xx too, should be good now, thanks!

PR #2631, opened 2023-08-23
switch any-pointer:coarse to pointer:coarse

disberd:
The current CSS keeps the hide cell buttons always on when Pluto is opened on laptops which also have a touchscreen.

Using pointer instead of any-pointer should only make the button always on only on devices that have the touchscreen as primary.

With the proposed change the button will still be always on on devices whose primary pointer is a touchscreen, even if they have a secondary pointer connected with fine accuracy as the media query for pointer:coarse comes after the one with any-pointer:fine

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -1519,7 +1519,7 @@ pluto-input > button > span {
     }
 }
 
-@media screen and (any-pointer: coarse) {
+@media screen and (pointer: coarse) {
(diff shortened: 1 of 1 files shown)


Pangoraw:
Cool! I think it makes sense, it would be nice if browsers could change the primary pointer to coarse when you use the touchscreen kind of like mouse/controller switch in video games.

PR #2632, opened 2023-08-25
fix EmbeddableDisplay

### LINKED ISSUE #2619: Click "more" for embeddable element doesn't work
The "more" button doesn't seem to work as expected (and reported in #1126) when embedded in an `@htl` block. It works normally for direct output.
[more_bug.jl.txt](https://github.com/fonsp/Pluto.jl/files/12188487/more_bug.jl.txt)
[more_bug.webm](https://github.com/fonsp/Pluto.jl/assets/12857162/ac47e995-75f9-4c12-a2f1-808acaef7707)

jbrea:
Fixes #2619

### DIFF
--- a/frontend/components/Cell.js
+++ b/frontend/components/Cell.js
@@ -337,7 +337,7 @@ export const Cell = ({
                 onerror=${remount}
             />
             ${show_logs && cell_api_ready
-                ? html`<${Logs} cell_id=${cell_id} logs=${Object.values(logs)} line_heights=${line_heights} set_cm_highlighted_line=${set_cm_highlighted_line} />`
+                ? html`<${Logs} logs=${Object.values(logs)} line_heights=${line_heights} set_cm_highlighted_line=${set_cm_highlighted_line} />`
                 : null}
             <${RunArea}
                 cell_id=${cell_id}
--- a/frontend/components/Cell.js
+++ b/frontend/components/Cell.js
@@ -337,7 +337,7 @@ export const Cell = ({
                 onerror=${remount}
             />
             ${show_logs && cell_api_ready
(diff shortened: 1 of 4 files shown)


disberd:
I just tried out this pull request but unfortunately it does not seem to fix #2619 for logs. *more* inside `@info` does not seem to work in any situation (with or without `embed_display` and with or without `@htl`)

jbrea:
Ah, yes, indeed. I only looked at my original bug report. This is the only functionality I need for the class starting mid September.
If I find the time I can try to have a look at `@info` and friends.

jbrea:
This takes too much time for me, sorry. I fixed one issue, which was that all logs had a hand-written cell-id "adsf".
But now I would have to mess with `set_output!` in Run.jl and I don't yet really see the best solution. Maybe it would be best to treat logs as special `published_outputs`? Or copy the mechanism of `published_outputs` to `logs`? No clue, and I don't have the time to figure it out and fix this this now.

disberd:
Yeah don't worry I'll just open a new issue for logs.
I think you can also just revert your latest commit so that you only fix the lingering error in embed display.
That way it's really just a single line fixing a specific issue clearly identified issue. Hopefully that way it will be much faster to get merged :)

jbrea:
OK, thanks.

disberd:
Tagging @Pangoraw for review/merge. This is important to get merged before the next release as `embed_display` will break since #2608 has been merged (and missed changing the code of `show` in this fix)

Pangoraw:
Thanks @jbrea and @disberd! I added a test which could have detected the problem and opened a new issue to track the problem in logs: #2633.

PR #2637, opened 2023-09-03
change selector so that CSS applies correctly

### LINKED ISSUE #2583: Variable name "info" causes unexpected css formatting
I noticed some weird formatting issue occuring on a specific cell in one of my notebooks. I narrowed it down to the definition of anything (variable or function does not matter) with the name "info" in that specific cell. The cell where the definition is will always have additional whitespace and a displaced "Add cell above" button.

https://github.com/fonsp/Pluto.jl/assets/38977690/587cf899-601b-4655-b71d-515a8bd71313

grafik

When looking at the resulting page with "Inspect Element", it seems like the output of most Pluto cell produces one or more (invisible) spans with the variable name as its element id. In the case of a variable called `info` this results in a `<span id=info></span>` tag. The problem is now that the footer element containing the FAQ link and feedback has the same `id`. Meaning that the CSS formatting meant for the footer also applies to the span above the cell defining a variable called `info` causing the additional whitespace. https://github.com/fonsp/Pluto.jl/blob/9d7e050bdbbd07958d7fe1e252968649a25bc37f/frontend/components/Editor.js#L1590-L1600
https://github.com/fonsp/Pluto.jl/blob/9d7e050bdbbd07958d7fe1e252968649a25bc37f/frontend/editor.css#L2658-L2669

This seems like it could be easily fixed either by renaming the id in the footer to something less likely to be picked as a variable name or change the CSS to make sure that it only applies to the correct element.

Thank you very much for your work on this awesome tool, we had a really great experience last semester teaching an introduction into scientific computing using Julia and Pluto!

vladflore:
change selector so that CSS applies only to element with id info inside footer element

Fixes #2583

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -2665,7 +2665,7 @@ footer button {
     font-size: 0.75rem;
 }
 
-#info {
+footer #info {
     max-width: 9400px;
     margin: 0 auto;
     padding: 1rem;


pankgeorg:
ah that's true! We add an `id` with the variable name on each variables, so you can link to that with a fragment (`#info` in this case) in the URL.

PR #2648, opened 2023-09-15
Fix typo in docstring

LilithHafner:
The primary change here is:
> ...snoop out all methods that **we** defined by us, and not by another package...

to
> ...snoop out all methods that **were** defined by us, and not by another package...

My editor also automatically stripped out trailing whitespace on file save, I hope that's okay. I can put back the trailing whitespace if necessary.

### DIFF
--- a/src/runner/PlutoRunner.jl
+++ b/src/runner/PlutoRunner.jl
@@ -9,7 +9,7 @@
 # SOME EXTRA NOTES
 
 # 1. The entire PlutoRunner should be a single file.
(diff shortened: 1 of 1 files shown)


pankgeorg:
Thanks for contributing! I'm not sure what our policy on whitespace is (it kind of messes up the git history); but if this PR only had the one changed line I'll approve it right away!

pankgeorg (approved):
Thanks! You can make a new white space PR if you want and we can discuss that there! Or maybe even a formatter one. (It would be super cool if the commit message for each line referenced the previous commit of each line, but I don't know if there is any tool that actually does that)

PR #2653, opened 2023-09-19
add `msgpack_type` method Expr and related 

### LINKED ISSUE #2652: `workspace_custom_startup_expr` kwarg seems broken
This is based on the zulip discussion here https://julialang.zulipchat.com/#narrow/stream/243342-pluto.2Ejl/topic/pass.20arguments.20into.20Pluto

I didn't even know that the `workspace_custom_startup_expr` was a configuration variable at all, but it seems that it can not really be used (at least on windows).

I just tried calling the example code put on zulip `Pluto.run(workspace_custom_startup_expr=:(ARGS = ["ena", 2, :three]))` but this breaks the Pluto server (even before creating or opening notebooks) with this error:

```
 Warning: Failed to write to WebSocket of 15p5r1q
│   exception =
│    no non-`AnyType` MsgPack mapping found for Expr; please overload `msgpack_type` for this type.
│    Stacktrace:
```

While this is just printed as a warning, it basically breaks the Pluto server.

This also seems to be tested here https://github.com/fonsp/Pluto.jl/blob/27b23a408ebe12e2ef335405b2f3899fd40775b5/test/Configuration.jl#L128-L133

but it does not seem to create a CI error

disberd:
This PR should fix #2652

### DIFF
--- a/src/webserver/MsgPack.jl
+++ b/src/webserver/MsgPack.jl
@@ -30,6 +30,8 @@ MsgPack.msgpack_type(::Type{Configuration.EvaluationOptions}) = MsgPack.StructTy
 MsgPack.msgpack_type(::Type{Configuration.CompilerOptions}) = MsgPack.StructType()
 MsgPack.msgpack_type(::Type{Configuration.ServerOptions}) = MsgPack.StructType()
 MsgPack.msgpack_type(::Type{Configuration.SecurityOptions}) = MsgPack.StructType()
+# We also want expression to be packable due to the `workspace_custom_startup_expr` configuration variable
+MsgPack.msgpack_type(::Type{T}) where T <: Union{Expr, LineNumberNode, QuoteNode} = MsgPack.StructType()
 
 # Don't try to send callback functions which can't be serialized (see ServerOptions.event_listener)
 MsgPack.msgpack_type(::Type{Function}) = MsgPack.NilType()


pankgeorg:
Can you add a test? (I'm just curious how it gets serialized 😅)

disberd:
@pankgeorg serialization is just a Dict:
```julia
julia> MsgPack.pack(:(LOL = [1, "string", :symbol])) |> MsgPack.unpack
Dict{Any, Any} with 2 entries:
  "head" => "="
  "args" => Any["LOL", Dict{Any, Any}("head"=>"vect", "args"=>Any[0x01, "string", Dict{Any, Any}("value"=>"symbol")])]
```

The problem is in de-serialization. It is not straightforward to reconstruct the original Expr like this.

But I am thinking, isnt' MsgPack just used to send data to the front-end? Since the test in https://github.com/fonsp/Pluto.jl/blob/27b23a408ebe12e2ef335405b2f3899fd40775b5/test/Configuration.jl#L128-L133 already shows that this kwarg is correctly executed in the notebook workspace, maybe here is the problem is that this kwarg at least should not be packed and sent via MsgPack at all to the HTTP server?

Adding a `MsgPack.pack` method seems to just avoid the error but sending this Expr via MsgPack should probably be avoided in the first place?

Pangoraw:
I think the msgpack serialization happens here when sending server config to the client: 

https://github.com/fonsp/Pluto.jl/blob/871517cdb359110d4e9c4c8c27c6321a9fadea4b/src/webserver/Dynamic.jl#L397-L397

Maybe it is simpler to send the config option as a string or simply drop it from the config when sending the config to the client instead of a serialized Expr?

Pangoraw:
> I think making that change would not be breaking since this setting is currently unusable.

@rikhuijzer is using it in PlutoStaticHTML, without client connections it works:

https://github.com/rikhuijzer/PlutoStaticHTML.jl/blob/234b907a1aecea1071146b5126594382a33efdf2/src/build.jl#L369-L373

fonsp:
A string is also necessary when using a TOML config like in PSS: https://github.com/JuliaPluto/PlutoSliderServer.jl#2-plutodeploymenttoml

Lemme see if I can do some hack with overloading `Base.set!` to support both...

fonsp:
https://github.com/fonsp/Pluto.jl/pull/2654

rikhuijzer:
> > I think making that change would not be breaking since this setting is currently unusable.
> 
> @rikhuijzer is using it in PlutoStaticHTML, without client connections it works:
> 
> https://github.com/rikhuijzer/PlutoStaticHTML.jl/blob/234b907a1aecea1071146b5126594382a33efdf2/src/build.jl#L369-L373
> 
> But a string would also make it clear that you cannot embed arbitrary code with packages not loaded on the notebook like:
> 
> ```julia
> workspace_custom_startup_expr = quote
>    MyPackage.do_something() # <- will not de-serialize on the notebook process
> end
> ```

@Pangoraw No worries about breaking PlutoStaticHTML 😄 . As long as another option remains for handling the situation then it's fine. PlutoStaticHTML is pinned to an exact Pluto version so is very robust to breaking changes.

PR #2660, opened 2023-10-02
Button to toggle presentation mode in export menu

fonsp:
A new button in the export menu to toggle presentation mode

image

See also #2611


fonsp:
@disberd @pankgeorg going with an easy-to-implement solution since the fullscreen detection in #2611 turned out to be difficult. Do you like this approach or do you still want to explore another idea?

PR #2666, opened 2023-10-18
Use featured.plutojl.org as the featured source for interactive previews!

fonsp:
This means that we can use our sliderserver when you view featured notebooks. Depends on https://github.com/JuliaPluto/featured/pull/17

This uses https://github.com/fonsp/Pluto.jl/commit/3b375490a2e2984a9cb81f50f23a71427269a83a for the new `valid_until` attribute (because we won't have permanent access to the plutojl.org domain) and the `id` attribute (this means that we can use the CDN featured source as a backup for the live one).


fonsp:
Oops we also need https://github.com/JuliaPluto/PlutoSliderServer.jl/pull/128

fonsp:
Now it only needs the featured server to be updated with the latest PSS version

PR #2672, opened 2023-10-21
⏱️ Estimate for package install and load time

fonsp:
image

Package installation & precompilation can take a long time. In Julia 1.9, this takes significantly longer (in exchange for shorter 2nd load times), making it even more important to display this information to the user. 

# How the estimate is calculated
I wrote a CI script that finds the (1000) most downloaded packages and their dependencies. For each package, it measures the installation, precompilation and load time, not including its dependencies. Using this data (timings + dependency graph) you can make an estimate for how long a set of packages will take to install.

See how it works at https://github.com/fonsp/package-loading-times

# Overestimation
The time given by this GUI will probably be an overestimation: it gives the time for an **average computer** with an **empty cache**. For example, if you have already installed & loaded Plots last week, then some of your cache is still usable by the new version of Plots and its deps. Or if you imported Plots already, then StatsPlots will be fast to install.

# TODO
- [ ] plutojl.org URL for the sources
- [ ] improve the automation of https://github.com/fonsp/package-loading-times
- [ ] Choose the right source depending on the Julia version?

# FUTURE IMPROVEMENTS
- [ ] Run some local benchmarks to scale the estimate results by
- [ ] Take cache into account for more realistic estimate

### DIFF
--- a/frontend/common/InstallTimeEstimate.js
+++ b/frontend/common/InstallTimeEstimate.js
@@ -83,8 +83,17 @@ export const time_estimate = (/** @type {PackageTimingData} */ data, /** @type {
     let sum = (xs) => xs.reduce((acc, x) => acc + (x == null || isNaN(x) ? 0 : x), 0)
 
     return {
-        install: sum(times.map(_.property("install"))),
-        precompile: sum(times.map(_.property("precompile"))),
-        load: sum(times.map(_.property("load"))),
+        install: sum(times.map(_.property("install"))) * timing_weights.install,
+        precompile: sum(times.map(_.property("precompile"))) * timing_weights.precompile,
+        load: sum(times.map(_.property("load"))) * timing_weights.load,
     }
+}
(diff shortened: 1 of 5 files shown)


fonsp:
The feature disables itself if the URL is not available so let's merge as an experiment!

rikhuijzer:
> Package installation & precompilation can take a long time. In Julia 1.9, this takes significantly longer (in exchange for shorter 2nd load times), making it even more important to display this information to the user.

Agreed! I am learning a lot from your focus on great UX! 

> The time given by this GUI will probably be an overestimation: it gives the time for an average computer with an empty cache.

Depends also a bit on the runner. My guess is that GitHub Runners are basically old CPUs that no one wants to use inside Azure, but sometimes you get an amazing runner with proper clock speeds.

Maybe Pluto has access to a self-hosted runner somewhere with some consistent performance?

Issue #2674, opened 2023-10-23
Safe preview still modifies the Pluto file


ederag:
The new safe preview is real nice because it allows inspection before running code,
but it is not "safe" for the file itself, which can be modified even before "Run notebook code" is clicked.

For instance, opening a notebook  starting with
```
### A Pluto.jl notebook ###
# v0.19.27
```
e.g. [Pluto_lost_disabled_23803aa.zip](https://github.com/fonsp/Pluto.jl/files/13068930/Pluto_lost_disabled_23803aa.zip)

results in this diff:
```diff
-# v0.19.27
+# v0.19.30
```
even if still in "safe preview mode":
![Screenshot_20231023_122536](https://github.com/fonsp/Pluto.jl/assets/10962809/18d33563-ce4c-49c9-acf5-45774e1a0bc7)

I'd prefer the file to stay pristine while in safe preview mode,
to allow copy/paste from an old notebook without touching the underlying git repo.

fonsp:
Thanks! Just to be clear, safe preview *should* let you modify the file, but it is a nice suggestion to not touch the file if no edits were done.

ederag:
Indeed it's important to be able to fix the notebook before running it.

About the title edit: I also see comment blocks being removed and code being moved around
at least for one notebook (not just the Pluto number changes).

klwlevy:
Dear Plutonians and Julians,
I understand "Safe preview" is something new. I am relatively new to Julia. At the moment, this new feature broke my otherwise working Pluto notebook, i.e. certain parts do not render anymore even after clicking "run the notebook". Is there anything I can do to remedy this? Downgrade Pluto, or something better? I am confident my notebook is safe for me to run.
I am attaching a screenshot with a minimal (non-) working example. The 4 cells for the grid are each working individually, but not this div grid. I would be very greatful for any helping pointers to fix my notebook again. The cell not working in the div is an embedded Youtube video, but the same problem also applies for my sidebar (not shown in screenshot).
 
![image](https://github.com/fonsp/Pluto.jl/assets/65867864/6eab2b1c-d5ab-49ec-9eba-cfd71bef7278)

klwlevy:
Addendum: Pinning Pluto to version 0.19.27 did solve my problem with "safe preview".

ederag:
It is probably unrelated to the "safe preview".
Please file your own issues rather than hijacking others,
to reduce noise, and make it easier for developers to keep track.
Thanks and welcome !

[...]

PR #2677, opened 2023-10-25
Fix: Don't drop disabled cells

### LINKED ISSUE #2676: Disabled cell content sometimes disappears from file
The content of disabled cells sometimes disappear from the file
This is an issue, as it can get lost when the notebook is closed.

This has been happening "randomly" for months, so it is not related to the latest release.
But I finally found a mwe, tested with `Pluto v0.19.30` and `julia 1.9.2`.

Note: in the following video, every time the `git gui` window appears, it shows the previous state,
until I hit F5 to get the current state (still have to figure out how to show the keystrokes...)
The last cell is in a "disabled" state.
The sequance shown in this video is
1. Click on the  `using BenchmarkTools` cell, which modifies the file.
2. The modification is not visible in the `git gui` that gets focus (previous state, no diff),
    then I press F5 in the `git gui` window and the view is updated, showing the actual diff.
3. Then I click on the "save code without running" on the disabled cell,
and do the same actions in the `git gui` window to show that the file gets fixed (no diff).
4. Execute the `using BenchmarkTools` cell, which modifies the file,
and do the same actions in the `git gui` window to show that the cell content got lost gain.

https://github.com/fonsp/Pluto.jl/assets/10962809/a74bd6a5-77ea-491b-bf26-ad220c651350

Here is the notebook:
https://gist.github.com/ederag/c6628a895b10b973d7fdc184d8236232#file-losing_disabled-jl

Note that `BenchmarkTools` is _not_ the culprit. The issue is the same with ` TestEnv `
or any other package present in the global environment.

To reproduce, just run the notebook from a fresh julia session.

Pangoraw:
Fixes #2676.

### DIFF
--- a/src/notebook/saving and loading.jl
+++ b/src/notebook/saving and loading.jl
@@ -56,10 +56,11 @@ function save_notebook(io::IO, notebook::Notebook)
     println(io)
 
     cells_ordered = collect(topological_order(notebook))
-    
+    @assert length(cells_ordered) == length(notebook.cells) "Cells were dropped"
+
     for c in cells_ordered
         println(io, _cell_id_delimiter, string(c.cell_id))
-        
+
         let metadata_toml = strip(sprint(TOML.print, get_metadata_no_default(c)))
             if metadata_toml != ""
                 for line in split(metadata_toml, "\n")
@@ -67,7 +68,7 @@ function save_notebook(io::IO, notebook::Notebook)
                 end
             end
         end
-        
+
         if must_be_commented_in_file(c)
             print(io, _disabled_prefix)
             print(io, replace(c.code, _cell_id_delimiter => "# "))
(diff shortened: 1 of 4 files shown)


src/notebook/saving and loading.jl
```diff
@@ -56,18 +56,24 @@ function save_notebook(io::IO, notebook::Notebook)
     println(io)
 
     cells_ordered = collect(topological_order(notebook))
-    
+    # TODO: add test for this case
```
fonsp, line 59: Can you add a comment here to explain when this can happen?

ederag:
While in "safe preview" the file is modified (that's #2674),
but after "Run notebook code", the file is back to normal (only the Pluto version changed from 0.19.29).
So this PR does fix #2676 indeed, even for larger, more involved files.
Thanks !

Pangoraw:
Thanks a lot for testing @ederag and for the detailed report!

fonsp:
@Pangoraw merging this because I am doing a release, but can you still add that comment? (But don't push to `main` until the release is tagged 🙈)

PR #2686, opened 2023-10-30
Show error message when statefile fails to load

fonsp:
image

You can test this PR by clicking on a featured notebook, and then editing the URL to make the statefile a bad URL.


fonsp:
@pankgeorg this is probably compatible with JH but you might want to make sure. It just adds a `.catch` to the promise returned by your overloaded `get_statefile` function.

PR #2700, opened 2023-11-01
fix: activate a temp environment when a sysimage is enabled

pankgeorg:
Starting with Julia 1.8, the sysimage internal manifest (here on `Ⅿ₀`) is included in every Pkg resolution. So, a new, fresh manifest is equivalent to `Ⅿ₀` ∪ `∅`. This is strictly better than the "default" manifest that can be inconsistent with the `Ⅿ₀`, or may include packages that will interfere with the new packages we want to add (e.g. an old version of a package that the package you're adding is depending on).

To address that, the proposal is to instantiate a fresh manifest (`temp=true`) instead of the default.

### DIFF
--- a/src/webserver/SessionActions.jl
+++ b/src/webserver/SessionActions.jl
@@ -221,7 +221,7 @@ function new(session::ServerSession; run_async=true, notebook_id::UUID=uuid1())
     notebook = if session.options.compiler.sysimage === nothing
         emptynotebook()
     else
-        Notebook([Cell("import Pkg"), Cell("# This cell disables Pluto's package manager and activates the global environment. Click on ? inside the bubble next to Pkg.activate to learn more.\n# (added automatically because a sysimage is used)\nPkg.activate()"), Cell()])
+        Notebook([Cell("import Pkg"), Cell("# This cell disables Pluto's package manager and activates the global environment. Click on ? inside the bubble next to Pkg.activate to learn more.\n# (added automatically because a sysimage is used)\nPkg.activate(;temp=true)"), Cell()])
     end
 
     # Run NewNotebookEvent handler before assigning ID


fonsp:
Oooh and can we get that internal manifest? is there also a project? in that case we should activate that internal environment, right?

pankgeorg:
> Oooh and can we get that internal manifest?
I don't think we can _get_ the manifest, but the Pkg knows about it.

> is there also a project? in that case we should activate that internal environment, right?
Hmm that's a good point. The most important repercussion I think is "What to serialize in cells "0000-...0001" and "0000-...0002", and I don't have a good answer, but I'll ask around

fonsp:
I think the original assumption was that JH will open Pluto notebooks with a sysimage, and with the Manifest-and-Project-that-was-used-to-generate-the-sysimage stored as the global environment. If people were to use a Julia REPL, we want them to use the global environment, and this way we match that on Pluto.

Is that no longer the case? Happy to merge either way because JH is probably the only user.

fonsp:
@pankgeorg

pankgeorg:
> I think the original assumption was that JH will open Pluto notebooks with a sysimage, and with the Manifest-and-Project-that-was-used-to-generate-the-sysimage stored as the global environment. If people were to use a Julia REPL, we want them to use the global environment, and this way we match that on Pluto.
> Is that no longer the case? Happy to merge either way because JH is probably the only user.

I'm not sure that this is the case currently. Pluto runs _with_ the sysimage and opens the notebooks with the sysimage. But `Pkg.activate()` doesn't necessarily activate the _sysimage_ environment. And I think we don't _need_ to either. Starting with Julia 1.9, I noticed that the internal pluto package manager does work with sysimages; it's just terribly slow. Maybe the right fix here is to teach it to be faster (e.g. ignoring or fast-pathing the resolution of `Base._sysimage_modules`). I need to investigate this a bit more.

PR #2698, opened 2023-10-31
Turn ExpressionExplorer.jl into its own package

fonsp:
🥳 https://github.com/JuliaPluto/ExpressionExplorer.jl

# TODO
- [x] ExpressionExplorer.jl 1.0.0 release (needs https://github.com/JuliaPluto/ExpressionExplorer.jl/issues/1)
- [x] What should happen to the extra functions? `is_implicit_using transform_dot_notation collect_implicit_usings compute_usings_imports! compute_usings_imports external_package_names is_toplevel_expr get_rootassignee can_be_function_wrapped`

PR #2717, opened 2023-11-16
Factor topology into package: PlutoDependencyExplorer.jl

fonsp:
We already did #2698 which means that people can find the references and definitions of an `Expr` in another package more easily. This PR factors out a second package that layers on top of ExpressionExplorer: given a list of `Expr` (i.e. a notebook of cells), in which order should the expressions be evaluated?

Not sure about the name PlutoReactiveCore

TODO:
- [x] Factor out the `Cell`. PlutoReactiveCore only uses it as a key (like `Dict{Cell,ReactiveNode}`), without accessing any of its fields. (This was the result of previous refactors.) To solve the dependency, it could be replaced with a `UUID` everywhere. But this will means lots of conversions everywhere in our code that's less charming. We could also use an `AbstractCell` type in PlutoReactiveCore, with a `Cell <: AbstractCell` in Pluto.
- [x] Factor our the `PlutoRuner.@bind` dependency used in the static macro expand of ExpressionExplorerExtras. Easiest is probably to write a copy of the macro inside PlutoReactiveCore, since it's only used for analysis.
- [x] Make sure that it works inside the Pluto codebase as a `module`, and put into a separate package afterwards.


fonsp:
@quildtide you would like this!

fonsp:
@pankgeorg any name suggestions?

src/analysis/PlutoReactiveCore/ExpressionExplorer.jl
```diff
@@ -78,17 +91,39 @@ function macro_has_special_heuristic_inside(; symstate::SymbolsState, expr::Expr
...
+const found_plutorunner = Ref{Union{Nothing,Module}}(nothing)
+
+function get_plutorunner()
+    fpr = found_plutorunner[]
+    if fpr === nothing
+        # lets try really hard to find it!
+        if haskey(Base.loaded_modules, plutorunner_id)
+            found_plutorunner[] = Base.loaded_modules[plutorunner_id]
+        elseif haskey(Base.loaded_modules, pluto_id)
+            found_plutorunner[] = Base.loaded_modules[pluto_id].PlutoRunner
```
Pangoraw, line 116: I don't think we want this one first. If Pluto is loaded in the process then Pluto.PlutoRunner should be preferred (supposing that PlutoReactiveCore is most likely being called by Pluto).
```suggestion
        if haskey(Base.loaded_modules, pluto_id)
            found_plutorunner[] = Base.loaded_modules[pluto_id].PlutoRunner
        elseif haskey(Base.loaded_modules, plutorunner_id)
            found_plutorunner[] = Base.loaded_modules[plutorunner_id]
```

Pangoraw (commented):
Nice! I think the added complexity is justifiable if we can use it directly in other packages such as PlutoUI for `@div` or maybe we can get away with a simpler version in PlutoUI without the extra complexity?

pankgeorg:
> Not sure about the name PlutoReactiveCore

It feels that this isn't the reactive core; the reactive core _uses_ the topology that this package outlines IIUC.

I'd call this `ReactivityExplorer` or `DependencyExplorer` or `OrderExplorer`, following `ExpressionExplorer` or something

fonsp:
yolooo

Issue #2719, opened 2023-11-18
Show multiple authors in frontmatter preview


lukavdplas:
This is a tiny feature request for the frontmatter-based preview!

When you have multiple authors, the preview will show only the first author in the preview:

![screenshot of Pluto frontmatter GUI. The form shows two authors, Pat and Mat, but the preview only lists Pat](https://github.com/fonsp/Pluto.jl/assets/43678097/fb949f29-c67a-462a-81ac-c09070e29d36)

(Notebook used for this screenshot: https://gist.github.com/lukavdplas/2f06ccf359994dd5c9593afcf1040904)

To be honest, I'm sure if the preview is used for anything other than the featured notebooks, but it matters for those: it wrongly gives the suggestion that a collaborative work was created by a single author.

**Suggested solution:**

Listing all authors can be an issue in the limited space (keep in mind I've used comically short names in the screenshot). But just adding "et al." or "and others" would be fine in my opinion - just to make it clear that this is not the only person involved.

![edited version of the previous screenshot. The preview shows the author as "Pat et al."](https://github.com/fonsp/Pluto.jl/assets/43678097/59a1a0e8-3751-4173-aeaf-0fd17387b067)

PR #2723, opened 2023-11-20
Multiple authors in feature card

lukavdplas:
Show "and others" in the featured notebook card when it has more than one author listed (close https://github.com/fonsp/Pluto.jl/issues/2719)

For instance, if the first notebook would have more authors, it would look like this:

screenshot showing preview of "getting started" notebook, with authors listed as "pluto.jl and others"


fonsp:
Wooohooo thanks!!

PR #2760, opened 2023-12-18
⌨️ Keyboard shortcut Alt-Arrow to move cells

fonsp:
You can now use `Alt` + `Up`/`Down` (or `Option` + `Up`/`Down` on Mac) to move cells in the notebook.

This was already the shortcut to move lines within a cell (just like VS Code). I found a nice way to make the shortcut work naturally for both actions:
- If you selected one or more cells (with the click-drag action), use the shortcut to move cells.
- If you are editing a single-line cell, use the shortcut to move the cell.
- If you are editing a multiline cell, use the shortcut to move (selected) lines.
- If you are editing a multiline cell and you selected all code (e.g. with Cmd + A), use the shortcut to move the cell.

All methods will auto-scroll (instant) to keep the cell in view. 

https://github.com/fonsp/Pluto.jl/assets/6933510/96e120bb-80e2-4b72-92bc-861b76fe2656

### DIFF
--- a/frontend/common/KeyboardShortcuts.js
+++ b/frontend/common/KeyboardShortcuts.js
@@ -1,6 +1,9 @@
 export let is_mac_keyboard = /Mac/.test(navigator.platform)
 
-export let ctrl_or_cmd_name = is_mac_keyboard ? "Cmd" : "Ctrl"
+export let control_name = is_mac_keyboard ? "⌃" : "Ctrl"
+export let ctrl_or_cmd_name = is_mac_keyboard ? "⌘" : "Ctrl"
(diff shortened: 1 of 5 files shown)


fonsp:
I had one weird issue when moving the cell from the CM (when everything is selected): moving up works fine, but moving down causes the CM to lose focus, so I had to manually refocus the cell after moving. It looks janky and `cm_forced_focus`ed cells dont fully work. Annoying!

Not sure why it happens, the preact `key` is working properly: the cell input is not being regenerated, it just loses focus.

fonsp:
Part of #65

PR #2767, opened 2024-01-03
Enable Malt.jl by default on all systems

fonsp:
This PR enables #2240 on all systems!

This completes the gradual roll-out of the new feature, everything went great!


PallHaraldsson:
Why is it that "Linux and MacOS will work best" (in README, added there 3 years ago)? Is it because you couldn't interrupt on Windows but now can (because of Malt)? Is Malt at least as good on all platforms? And if no other Windows Pluto issues, drop that outdated caveat?

fonsp:
@PallHaraldsson That's totally true, thanks for the comment! We can remove it now

PallHaraldsson:
The something could now be one one line:
```
elseif something(session.options.evaluation.workspace_use_distributed_stdlib, false)
        Malt.DistributedStdlibWorker
```

maybe more readable, I wasn't familiar with it, and also is it for sure needed, i.e. can you (for sure) expect `nothing` returned?

PR #2774, opened 2024-01-15
Add heap size hint

kirill-kondrashov:
This MR intends to integrate the [heap size hint](https://julialang.org/blog/2023/04/julia-1.9-highlights/#memory_usage_hint_for_the_gc_with_--heap-size-hint): 

> that enables users to set a limit on memory usage, after which the garbage collector (GC) will work more aggressively to clean up unused memory.

For the Julia processes launched in Pluto.


pankgeorg:
One thing we've found to work better than heap-size-hint, is actually invoking the `GC.gc()` function on a background task, when memory pressure on a system is important and is creating problems.

kir19890817:
> One thing we've found to work better than heap-size-hint, is actually invoking the `GC.gc()` function on a background task, when memory pressure on a system is important and is creating problems.

I see; it sounds like a good thing to check. Let me have a look.

fonsp:
Can you make this PR again, but:
- Avoid formatting unrelated lines of code
- Only add the option to `CompilerOptions`, not also to `ServerOptions`
- Make sure that the tests pass

I would recommend searching for `math_mode` and doing the exact same for `heap_size_hint`

fonsp:
@pankgeorg That's a nice suggestion but let's track that in a new issue and PR! This is just about a command line flag that was added in Julia 1.9

pankgeorg:
> @pankgeorg That's a nice suggestion but let's track that in a new issue and PR! This is just about a command line flag that was added in Julia 1.9

True, this is completely unrelated to the PR, it's mostly an offtopic response that addresses the intent of the PR, not the content of it. I'll open a new issue, and probably also a PR for memory management.

DartWEIDE:
> Can you make this PR again, but:
> 
>     * Avoid formatting unrelated lines of code
> 
>     * Only add the option to `CompilerOptions`, not also to `ServerOptions`
> 
>     * Make sure that the tests pass
> 
> 
> I would recommend searching for `math_mode` and doing the exact same for `heap_size_hint`

Sure, thanks for suggestion!

kir19890817:
> Can you make this PR again, but:
> 
>     * Avoid formatting unrelated lines of code
> 
>     * Only add the option to `CompilerOptions`, not also to `ServerOptions`
> 
>     * Make sure that the tests pass
> 
> 
> I would recommend searching for `math_mode` and doing the exact same for `heap_size_hint`

Sure, thanks for comment!

kir19890817:
> > @pankgeorg That's a nice suggestion but let's track that in a new issue and PR! This is just about a command line flag that was added in Julia 1.9
> 
> True, this is completely unrelated to the PR, it's mostly an offtopic response that addresses the intent of the PR, not the content of it. I'll open a new issue, and probably also a PR for memory management.

Hey, @pankgeorg thanks for the suggestion again!

Could you please elaborate a bit on what you mean by background task here? As far as I can track (with my poor understanding of the codebase as a disclaimer), for instance, in case of running the notebooks for exporting, in [PlutoSliderServer.jl](https://github.com/JuliaPluto/PlutoSliderServer.jl/blob/main/src/PlutoSliderServer.jl#L352) in the function `run_directory` which is called by `export_directory`, the new 'process' is created:
```julia
new = process(s; server_session, settings, output_dir, start_dir, progress)
```
`process` is implemented in [Actions.jl](https://github.com/JuliaPluto/PlutoSliderServer.jl/blob/main/src/Actions.jl#L46). It triggers [Pluto.jl/src/webserver/SessionActions.jl](https://github.com/fonsp/Pluto.jl/blob/main/src/webserver/SessionActions.jl#L44) `open`. Because it seems to be a separate process, there's no way to trigger `GC.gc()` somewhere 'in the background', outside of `Pluto.jl`.

fonsp:
Thanks @kirill-kondrashov ! Let us know when the PR is ready

pankgeorg:
> > > @pankgeorg That's a nice suggestion but let's track that in a new issue and PR! This is just about a command line flag that was added in Julia 1.9
> > 
> > 
> > True, this is completely unrelated to the PR, it's mostly an offtopic response that addresses the intent of the PR, not the content of it. I'll open a new issue, and probably also a PR for memory management.
> 
> Hey, @pankgeorg thanks for the suggestion again!
> 
> Could you please elaborate a bit on what you mean by background task here? As far as I can track (with my poor understanding of the codebase as a disclaimer), for instance, in case of running the notebooks for exporting, in [PlutoSliderServer.jl](https://github.com/JuliaPluto/PlutoSliderServer.jl/blob/main/src/PlutoSliderServer.jl?rgh-link-date=2024-01-17T16%3A28%3A43Z#L352) in the function `run_directory` which is called by `export_directory`, the new 'process' is created:
> 
> ```julia
> new = process(s; server_session, settings, output_dir, start_dir, progress)
> ```
> 
> `process` is implemented in [Actions.jl](https://github.com/JuliaPluto/PlutoSliderServer.jl/blob/main/src/Actions.jl?rgh-link-date=2024-01-17T16%3A28%3A43Z#L46). It triggers [Pluto.jl/src/webserver/SessionActions.jl](https://github.com/fonsp/Pluto.jl/blob/main/src/webserver/SessionActions.jl?rgh-link-date=2024-01-17T16%3A28%3A43Z#L44) `open`. Because it seems to be a separate process, there's no way to trigger `GC.gc()` somewhere 'in the background', outside of `Pluto.jl`.

Indeed. Pluto does inject _some_ code to the host process though, so it could also add a `Timer(() -> GC.gc(), 3600.; interval=3600.)`, or something a little bit more complex. Pluto itself runs a webserver too, so `Pluto.run()` could also have a function that periodically does a full sweep.

The tricky part (which julia could/should do, but currently doesn't) is to track
1. whether it's a good time to run the GC _now_ (good time = not too stressed)
2. whether there has been enough allocations to warrant a GC sweep.

I think the GC doesn't keep statistics of allocations around, so knowing [2] isn't very easy, at least not without consulting the kernel.

On the other hand, none of this should be Pluto's job in the first place right? This effort should most likely be another package.

kirill-kondrashov:
> Thanks @kirill-kondrashov ! Let us know when the PR is ready

Seems to pass the tests now

kirill-kondrashov:
> > > > @pankgeorg That's a nice suggestion but let's track that in a new issue and PR! This is just about a command line flag that was added in Julia 1.9
> > > 
> > > 
> > > True, this is completely unrelated to the PR, it's mostly an offtopic response that addresses the intent of the PR, not the content of it. I'll open a new issue, and probably also a PR for memory management.
> > 
> > 
> > Hey, @pankgeorg thanks for the suggestion again!
> > Could you please elaborate a bit on what you mean by background task here? As far as I can track (with my poor understanding of the codebase as a disclaimer), for instance, in case of running the notebooks for exporting, in [PlutoSliderServer.jl](https://github.com/JuliaPluto/PlutoSliderServer.jl/blob/main/src/PlutoSliderServer.jl?rgh-link-date=2024-01-17T16%3A28%3A43Z#L352) in the function `run_directory` which is called by `export_directory`, the new 'process' is created:
> > ```julia
> > new = process(s; server_session, settings, output_dir, start_dir, progress)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `process` is implemented in [Actions.jl](https://github.com/JuliaPluto/PlutoSliderServer.jl/blob/main/src/Actions.jl?rgh-link-date=2024-01-17T16%3A28%3A43Z#L46). It triggers [Pluto.jl/src/webserver/SessionActions.jl](https://github.com/fonsp/Pluto.jl/blob/main/src/webserver/SessionActions.jl?rgh-link-date=2024-01-17T16%3A28%3A43Z#L44) `open`. Because it seems to be a separate process, there's no way to trigger `GC.gc()` somewhere 'in the background', outside of `Pluto.jl`.
> 
> Indeed. Pluto does inject _some_ code to the host process though, so it could also add a `Timer(() -> GC.gc(), 3600.; interval=3600.)`, or something a little bit more complex. Pluto itself runs a webserver too, so `Pluto.run()` could also have a function that periodically does a full sweep.
> 
> The tricky part (which julia could/should do, but currently doesn't) is to track
> 
>     1. whether it's a good time to run the GC _now_ (good time = not too stressed)
> 
>     2. whether there has been enough allocations to warrant a GC sweep.
> 
> 
> I think the GC doesn't keep statistics of allocations around, so knowing [2] isn't very easy, at least not without consulting the kernel.
> 
> On the other hand, none of this should be Pluto's job in the first place right? This effort should most likely be another package.

Makes sense! I think Pluto, indeed, shouldn't be the first place; PlutoSliderServer seemed to me as a suitable place to insert that procedure.

PR #2778, opened 2024-01-16
Move precompile statements to PlutoRunner

Pangoraw:
not ideal to add a dependency to PlutoRunner but it seems to speed up startup.


fonsp:
Thank you!!

PR #2786, opened 2024-01-21
fix tests on nightly

Pangoraw:


fonsp:
It looks like we have a new failure on nightly with method deletion, can you take a look?

https://github.com/fonsp/Pluto.jl/actions/runs/7697545251/job/20974785353?pr=2797#step:5:1271

PR #2787, opened 2024-01-22
◐ Spin Pkg terminal spinners with CSS for performance

### LINKED ISSUE #2739: Fancyprint pkg logs: not throttled on slow connection
#2500 is causing a slowdown when using Pluto remotely on slow connections

fonsp:
#2500 introduced two performance issues:
1. High network pressure and latency #2739
2. The terminal outputs are stored in the state, and are included in every codemirror as a compartment value (because the terminal text is displayed in the Pkg popup). This was causing lots of expensive re-renders. According the the profile, most time was spent in the CM6 compartment/extension state management.

The solution in this PR is to replace all 4 progress characters with `◐` on the server side (before diffing). This fixes both issues because the updates are now much less frequent.

To still get a spinning animation, we search for the `◐` character, give it a class and make it spin with a CSS animation. The result is almost identical!

# Before
https://github.com/fonsp/Pluto.jl/assets/6933510/55d2052b-fd74-455d-be66-47065a714130

Scherm­afbeelding 2024-01-22 om 11 07 29

# After

https://github.com/fonsp/Pluto.jl/assets/6933510/c8ab4466-4c95-46f1-a5d2-3c9c769b9f3e

Scherm­afbeelding 2024-01-22 om 11 04 15

<details>
<summary>You can see that there are much fewer messages from the server:
</summary>

https://github.com/fonsp/Pluto.jl/assets/6933510/05df3724-833c-46c1-b4d5-0ed3038dd6dc

</details>

fix #2739

### DIFF
--- a/frontend/components/PkgTerminalView.js
+++ b/frontend/components/PkgTerminalView.js
@@ -1,12 +1,14 @@
 import AnsiUp from "../imports/AnsiUp.js"
 import { html, Component, useState, useEffect, useRef, useLayoutEffect } from "../imports/Preact.js"
 
+const make_spinner_spin = (original_html) => original_html.replaceAll("◐", `<span class="make-me-spin">◐</span>`)
+
 const TerminalViewAnsiUp = ({ value }) => {
     const node_ref = useRef(/** @type {HTMLElement?} */ (null))
 
     useEffect(() => {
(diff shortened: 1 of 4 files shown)


fonsp:
I'm really proud of this PR 😙

PR #2789, opened 2024-01-23
Make `#id` URLs work by setting the location.hash again after all cells are rendered

fonsp:
Now you will be able to share a URL like:

https://plutojl.org/en/docs/abstractplutodingetjes/#c3e83475-6ab0-4aa8-9a57-c2f25772cc75

Before this PR, this does not work because cells are rendered with a delay, and the UA tries to scroll to the element quickly after the page got loaded.

---

This PR also makes this URL work:

https://plutojl.org/en/docs/abstractplutodingetjes/#possible_values

By giving all `Docs.binding` displays the `id` of the variable name.


fonsp:
Unfortunately this does not work with Chrome's "Copy link to this selection" feature, because the special chrome hash (`#:~:text=.-,Add%20a%20comment,-Comment`) is hidden from the `window.location.hash`, so there's nothing we can do about that.

PR #2792, opened 2024-01-26
fix more button not working

### LINKED ISSUE #2791: The `more` button for showing more data of an output does not work anymore
Pressing the `more` button does not seem to work anymore

https://github.com/fonsp/Pluto.jl/assets/12846528/3c6b5373-03cd-43cd-8c16-2b769e7a37fd

Warning of the error showed up on the REPL which started the Pluto server:
image

disberd:
Should fix #2791


fonsp:
Thanks!!

Issue #2668, opened 2023-10-19
Copy-pasteable code blocks


gdalle:
It would be nice if the code blocks put in Pluto's Markdown cells were copy-pasteable, like those from Documenter. I use Pluto a lot for teaching, and sometimes I want students to copy-paste code that should not run in my notebook

gdalle:
:party

PR #2799, opened 2024-02-02
feat: add copy button for markdown code blocks

chayandatta:
fixes: #2668

image
image

Thanks @pankgeorg for all the help


fonsp:
Heyyy that's cool! Could you reupload the screenshots? The URLs are from a private repo

fonsp:
I would like to also use this functionality in more places!
- inside the Live Docs panel
- in code blocks that are used inside of a `PlutoUI.ExperimentalLayout.hbox`

Do you think it is possible to move this functionality to the RawHTMLContainer component? You could take a look at where we call `highlight` (which uses highlight.js to syntax highlight code blocks). Maybe this can be done in the same place?

Als try to extract this functionality into a new function, just like `highlight`.

chayandatta:
> You could take a look at where we call highlight (which uses highlight.js to syntax highlight code blocks). Maybe this can be done in the same place?

As far I can understand, where `hightlight()` is being called, It can be accessed only when the notebook is running, not on static mode.

and `generateCopyCodeButton` generates an element of Type 'ReactElement' and it is unable to prepend`(container.prepend())` with the element `container` because it is of it's type HTMLElement.

fonsp:
Thanks for taking a look again! 

This is not exactly what I meant. 🙈 Instead of using regex to find the code blocks, and then a preact element for the button, consider using `container.querySelectorAll` to find the code blocks, and `document.createElement` to create the button.

In CellOutput.js, add this block:

```js
                // Apply syntax highlighting
                try {
					... highlight ...
				} catch ... {}

				// NEW:
				// Find code blocks and add a copy button:
                try {
                    if (container.firstElementChild?.matches("div.markdown")) {
                        container.querySelectorAll("pre > code").forEach((code_element) => {
                            const pre = code_element.parentElement
                            add_code_copy_button(pre)
                        })
                    }
                } catch (err) {
                    console.warn("Adding markdown code copy button failed", err)
                }
```

Then the `add_code_copy_button` uses `document.createElement` to create the button, and `appendChild` to add it to the page.

fonsp:
> As far I can understand, where hightlight() is being called, It can be accessed only when the notebook is running, not on static mode.

It's true that `highlight` is not called when the notebook is in "safe mode". Let's think about that later!

frontend/components/CellOutput.js
```diff
@@ -683,3 +695,44 @@ export let highlight = (code_element, language) => {
...
+        const txt = range.startContainer.textContent || ''
+        navigator.clipboard
+            .writeText(txt)
+            .then(() => {
+                console.log("Code copied to clipboard:\n" + txt)
+            })
+            .catch((error) => {
+                console.error("Error copying code to clipboard:", error)
+            })
+    }
```
pankgeorg, line 721: I would move the copy function outside this function, to avoid creating a closure for every cell output.

check https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

```suggestion
/** Copies the contents of an HTML Element into the clipboard */
const copyToClipboard = (e) => {
  const pre = e.target.parentNode.parentNode // or something similar, to find the `pre`. See `.closest()`
  const range = document.createRange()
  range.selectNode(pre)
  const txt = range.startContainer.textContent || ''
  navigator.clipboard
    .writeText(txt)
    .then(() => {
      console.log("Code copied to clipboard:\n" + txt)
    })
    .catch((error) => {
       console.error("Error copying code to clipboard:", error)
    })
}

/**
 * Generates a copy button for Markdown code blocks and copies them into the clipboard.
 * @param {HTMLElement | null} pre - The <pre> element containing the code block.
 */
export const generateCopyCodeButton = (pre) => {
    if (!pre) {
        console.error('Error: pre is null.');
        return;
    }
```

frontend/components/CellOutput.js
```diff
@@ -683,3 +695,44 @@ export let highlight = (code_element, language) => {
...
+
+    // create copy button
+    const copyCodeButton = document.createElement("button")
+    copyCodeButton.className = "markdown-code-block-copy-code-button"
+    copyCodeButton.addEventListener("click", copyToClipboard)
+
+    // Create copy button image
+    const copyImg = document.createElement("img")
+    copyImg.src = "https://unpkg.com/ionicons@7.1.0/dist/svg/copy-outline.svg"
```
pankgeorg, line 730:
```suggestion
    copyImg.src = new URL("https://unpkg.com/ionicons@7.1.0/dist/svg/copy-outline.svg", import.meta.url)
```

I think this will help the bundler prepare the offline export, so may be a good addition

chayandatta:
https://github.com/fonsp/Pluto.jl/assets/32599474/eef8c171-5d84-4a3f-9545-d7cf0cd5f1c1

fonsp:
Thanks!

image

The button was positioned relative to the last line of code (with `pre.appendChild(button)` it ends up at the end of the child list, and `top: -20px` moves it up.) I fixed that by prepending the button to the start of the children of `<pre>`.

Now it's positioned as expected:

image

I also tested this on Firefox and Safari and it works great!

I removed the 

```js
.then(() => {
            console.log("Code copied to clipboard:\n" + txt)
        })
        .catch((error) => {
            console.error("Error copying code to clipboard:", error)
        })
```

Because the `then` won't be so useful anymore after this is merged, more useful during development, and the `catch` is actually done automatically! If you have a promise in an event handler that rejects, but there is no `catch`, Chrome will automatically log it to the console.

Thanks again!

PR #2802, opened 2024-02-06
[WIP] feat: offline-only export

pankgeorg:
[WIP]: this is an attempt to bundle all the assets required for a static export to work completely offline, with all the assets bundled in the HTML.

Parcel is supposed to do this, but there are quite a few caveats that make this not work as-is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ test/frontend/artifacts/*
 .DS_Store
 
 frontend-dist
+frontend-dist-static
 node_modules
 .parcel-cache
 frontend/dist
--- a/frontend-bundler/.parcelrc-static
+++ b/frontend-bundler/.parcelrc-static
@@ -7,5 +7,12 @@
     "reporters": [
         "...",
         "parcel-reporter-bundle-manifest"
-    ]
+    ],
+    "transformers": {
+        "*.svg": [
+            "...",
+            "@parcel/transformer-inline"
+        ],
+        ".html": ["..."]
+    }
 }
--- a/frontend-bundler/package-lock.json
+++ b/frontend-bundler/package-lock.json
@@ -12,249 +12,89 @@
                 "parcel-resolver-like-a-browser"
             ],
             "dependencies": {
-                "@parcel/config-default": "^2.0.0",
-                "mkdirp": "^1.0.4",
-                "parcel": "^2.0.0",
+                "@parcel/config-default": "^2.11.0",
+                "mkdirp": "^3.0.1",
+                "parcel": "^2.11.0",
                 "parcel-reporter-bundle-manifest": "^1.0.0",
                 "parcel-resolver-like-a-browser": "file:../frontend-bundler/parcel-resolver-like-a-browser"
             },
             "devDependencies": {
-                "@parcel/optimizer-data-url": "^2.0.0",
-                "@parcel/transformer-inline-string": "^2.0.0",
+                "@parcel/optimizer-blob-url": "^2.11.0",
+                "@parcel/optimizer-data-url": "^2.11.0",
+                "@parcel/transformer-inline": "^2.11.0",
+                "@parcel/transformer-inline-string": "^2.11.0",
                 "@plutojl/posthtml-crossorigin": "^1.0.0",
-                "@types/node": "^16.11.6",
+                "@types/node": "^20.11.16",
                 "posthtml": "^0.16.6",
                 "posthtml-sri": "^1.2.0"
             }
         },
         "node_modules/@babel/code-frame": {
-            "version": "7.22.13",
-            "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.22.13.tgz",
-            "integrity": "sha512-XktuhWlJ5g+3TJXc5upd9Ks1HutSArik6jf2eAjYFyIOf4ej3RN+184cZbzDvbPnuTJIUhPKKJE3cIsYTiAT3w==",
+            "version": "7.23.5",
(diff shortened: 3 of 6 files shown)


fonsp:
Interesting! Curious if the bundler issues can be fixed easily!

Did you look into external tools that could do this? (i.e. make an offline-ready snapshot version of this html file.) I tried the "Save complete page" feature from Chrome but that did not work. :(

And did you consider solutions that involve a custom CDN? Maybe you could self-host? You need to statically serve the `frontend-dist` folder of every Pluto version. Or maybe `cdn.jsdelivr.net` could be allow-listed?

The solution could also be a Julia script: `julia viewnotebook.jl path_to_my_nb.html` which starts a static file server and shows the notebook.

pankgeorg:
> Interesting! Curious if the bundler issues can be fixed easily!

me too! some issues seem like noone cares to fix

> Did you look into external tools that could do this? (i.e. make an offline-ready snapshot version of this html file.) I tried the "Save complete page" feature from Chrome but that did not work. :(

I've tried parcel with various configurations, [monolith](https://github.com/Y2Z/monolith) and [inliner](https://github.com/remy/inliner). They all get stuck somewhere in the dynamically loaded images, in loading the modules despite them being text/javascript or some combination

> And did you consider solutions that involve a custom CDN? Maybe you could self-host? You need to statically serve the `frontend-dist` folder of every Pluto version. Or maybe `cdn.jsdelivr.net` could be allow-listed?

No this is not an option for our use case, unfortunately. 

> The solution could also be a Julia script: `julia viewnotebook.jl path_to_my_nb.html` which starts a static file server and shows the notebook.

This might be good to do either way. I'll make a small package that does this, it can be a nice workaround for now.

pankgeorg:
And, of course, there is the [service worker approach](https://learn.microsoft.com/en-us/microsoft-edge/progressive-web-apps-chromium/how-to/service-workers)

fonsp:
One problem with the SW is that it only works the second time a page is loaded, but you can work around it by triggering a refresh on the first load. You should check that this will work on `file://` though.

pankgeorg:
> One problem with the SW is that it only works the second time a page is loaded, but you can work around it by triggering a refresh on the first load.

Yes, that was my thought too

>You should check that this will work on `file://` though.

[MDN says it should work](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) 

> Locally-delivered resources such as those with http://127.0.0.1 URLs, http://localhost and http://*.localhost URLs (e.g. http://dev.whatever.localhost/), and file:// URLs are also considered to have been delivered securely.

But will know more when I have a MWE

fonsp:
Also got mathjax and the fonts :)

fonsp:
With the JS changes and inlining ionicon svgs, the exports work offline, with mathjax, only the fonts don't load. 

But inlining the svgs bring the bundled editor.css from 80kB to 120kB and that's acceptable! So we can do this in the main bundle, and we don't need a separate bundle for offline support.

I'm thinking of tweaking our font stacks a bit to make the fallback fonts look more like Alegreya and Vollkorn, and just keeping it like that! Then the offline html will be just 1.5MB with slightly different looking fonts when viewing without internet. But we don't need a second bundle :)

fonsp:
https://github.com/fonsp/Pluto.jl/pull/2896 to make the fallback fonts look nicer :) which is something i wanted to do for a long time!

sjkelly:
Is there any chance this can be merged as-is? This seems like it is useful for procedurally generating notebook pages to HTML even without a UI:

```
using Pluto

s = Pluto.ServerSession()

nb_file = "plotly_static.jl"

nb = Pluto.SessionActions.open(s, nb_file; run_async=false)

# Generate the HTML file
html_content = Pluto.generate_html(nb)

# Write the HTML content to a file
open("notebook.html", "w") do io
    write(io, html_content)
end
```

fonsp:
Hey @sjkelly !

I also like this PR! Btw the code you wrote already works on the current Pluto release :) but the generated HTML does not work offline.

I want to enable this feature by default for all exports, so that our HTML files will always work without relying on the jsdelivr.com CDN like we currently do. 

What's left for this PR is #2898. Without that, the generated HTML files will not use Pluto's standard typefaces but always use one of the default system fonts. And that's not nice, because we want a what-you-see-is-what-you-get experience.

fonsp:
I'm going to merge the cleanup & npm updates & some boilerplate from this PR right now since that's already useful!

fonsp:
We still need https://github.com/fonsp/Pluto.jl/issues/2898 for this PR

pankgeorg:
> We still need #2898 for this PR

There is an additional issue here now; I had to update parcel and it now removes double quotes from attributes if they are unambiguous so our intelligent regexes need an update :)

fonsp:
> We still need #2898 for this PR

Fixed! 

> There is an additional issue here now; I had to update parcel and it now removes double quotes from attributes if they are unambiguous so our intelligent regexes need an update :)

Also fixed!

fonsp:
This PR is now almost ready! I updated the TODO list in the PR description

fonsp:
The output file size is now 7MB.

fonsp:
I noticed that `frontend-dist-offline` and `frontend-dist` almost completely overlap. The only files in `offline` that dont exist in `dist` are: `["editor.ae6e0116.js", "editor.bf2b6dcc.js", "editor.da1033a9.css"]`.

This means that there is a lot of potential file savings... But maybe package downloads happen in zip and it doesnt matter? It would be 6MB file savings

fonsp:
I think I will merge this in an almost-ready state, but not enable it by default yet when clicking on the Export HTML button in the Pluto UI. That lets us test the CDN and release pipeline first.

PR #2807, opened 2024-02-07
Fix method table stuff in Julia nightly

### LINKED ISSUE #2806: Nightly tests: overload of Base function not recovered
On Julia nightly, it looks like we are unable to restore an overloaded Base function:

https://github.com/fonsp/Pluto.jl/actions/runs/7761348184/job/21169580988#step:5:1320

We should check whether this is the case in general, or if something else is up with this particular test.

If so, oh ohh... maybe we need to get in touch with the Julia devs before this gets released and we lose the ability to reset base overloads? That would be a big bummer.

fonsp:
Fix #2806

### DIFF
--- a/src/runner/PlutoRunner/src/PlutoRunner.jl
+++ b/src/runner/PlutoRunner/src/PlutoRunner.jl
@@ -807,7 +807,7 @@ function delete_toplevel_methods(f::Function, cell_id::UUID)::Bool
     # we define `Base.isodd(n::Integer) = rand(Bool)`, which overrides the existing method `Base.isodd(n::Integer)`
     # calling `Base.delete_method` on this method won't bring back the old method, because our new method still exists in the method table, and it has a world age which is newer than the original. (our method has a deleted_world value set, which disables it)
     #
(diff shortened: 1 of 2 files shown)


src/runner/PlutoRunner/src/PlutoRunner.jl
```diff
@@ -785,6 +785,13 @@ function delete_method_doc(m::Method)
     end
 end
 
+
+if VERSION < v"1.7.0-0"
```
pankgeorg, line 789:
```suggestion
@static if VERSION < v"1.7.0-0"
```

I don't know what this does but I've seen in in places

fonsp: Thanks! I think it doesn't matter when the `if` is at toplevel inside of a module, because it will only run once either way. It's useful in cases like this:

https://github.com/fonsp/Pluto.jl/blob/6dfbefe20800c3963f016e4f9af7663e41d57029/src/packages/PkgCompat.jl#L77

so that the `if` will run only once when you import Pluto, not every time that you call the function

PR #2813, opened 2024-02-15
🧾 Stack traces: code preview, URL, truncated, design

fonsp:
image

In the REPL:

image

# New design
New design makes the stack trace look more inviting! It also adds more visual variety. There is also a title and some text to explain what it is.

# Code previews
The most relevant frames are (often) those from your own code. In this PR, you get a small preview of the code from your notebook that made the call. You can click to jump to the line.

This is implemented without CodeMirror, but just with a `<pre>` and highlight.js to keep things simple :)

# URLs
You can now click on file locations from Julia Base, and it will send you to the source code on github :) Adding this functionality for stdlibs and remote packages would be nice! Working on this with @adrhill

# Truncated stack trace
We search for the first frame that comes from this notebook, and all frames that came before it are hidden behind a *More...* button. This makes the stack trace less intimidating, and it is hopefully a good heuristic for "which frames are more useful?". It would be nice to add more heuristics. 

# Remove "top-level-scope"
Pretty technical, just removed the whole phrase, showing only the file:line.

# TODO
- [x] Dark mode tweaks
- [ ] More experimentation!
- [x] Make the error message stand out more, especially when it is a single line.

# More pictures

image

image

image

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -936,7 +936,6 @@ const InputContextMenu = ({ on_delete, cell_id, run_cell, skip_as_script, runnin
 
     const prevously_focused_element_ref = useRef(/** @type {Element?} */ (null))
     const setOpen = (val) => {
-        console.error("setOpen", val)
         if (val) {
             prevously_focused_element_ref.current = document.activeElement
         }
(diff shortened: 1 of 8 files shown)


adrhill:
Yay! 😃

Issue #2810, opened 2024-02-13
Precompilation problem on 1.11 with complete_remote_package


j-fu:
See  https://gist.github.com/j-fu/1fd2a044636cf0a2737b999db043cc22
Pluto doesn't precompile on nightly, seen this in nightly ci for mac,linux, windows.

May be a problem with Julia, though.

balenamiaa:
I think this is because [complete_remote_package](https://github.com/JuliaLang/Pkg.jl/blob/d267986c2d2100ad324cc67417c82e1f94114da5/ext/REPLExt/completions.jl#L55) is defined in an package extension of REPL now. It's used in [_registered_package_completions](https://github.com/fonsp/Pluto.jl/blob/c424619126afdbea3dd07f1cf27052aa01f52126/src/packages/PkgCompat.jl#L282). Replacing `Pkg.REPLMode` with `REPLExt` where 
```
using REPL
const REPLExt = Base.get_extension(Pkg, :REPLExt)
```
fixes it for me, but this seems like a hack.

j-fu:
Yes, this indeed seems like a hack.
Did some research: in 1.10, `complete_remote_package` is in the REPLMode submodule, and not exported. Formally, then it  is not part of the public API.

Not sure if it make sense to code a work-around in Pluto now.

IMHO in 1.11 this could be made part of public API using the new `public` keyword. According to the logic of extensions, once it is public, `Pkg.jl` somwehre needs to declare 

```julia
public function complete_remote_package end
```
Then the extension would extend this function once REPL is loaded. 
IMHO this would warrant an issue or a PR with Pkg.jl .

j-fu:
> const REPLExt = Base.get_extension(Pkg, :REPLExt)

Well, I just tried this out, the error as described in the gist persists.
The error occurs when calling `RegistryInstances.reachable_registries()` during precompilation: 

https://github.com/fonsp/Pluto.jl/blob/9f9ab2e1e858e978bcfe53f9ab2c28f2e547dfd9/src/packages/PkgCompat.jl#L170

I can call `RegistryInstances.reachable_registries()` from the REPL without problems. So there seems to be more to the picture, and we have at least two problems:
- fix the situation with `complete_remote_package`
- figure out what is going on with `RegistryInstances.reachable_registries()`

PR #2815, opened 2024-02-18
call _get_registries() in __init__ of PkgCompat.jl

j-fu:
See https://docs.julialang.org/en/v1/manual/modules/#Module-initialization-and-precompilation

Probably solved part of #2810


fonsp:
Thanks! Can you check the PR files for unwanted changes?

j-fu:
I think this is a longer story, as precompilation would have to be refactored, it calls RegistryInstances.reachable_registries() as well.

In the moment it is switched off in order to see what else could happen. I can try to get a more comprehensive refactoring.

I did run with a patched Pkg in order to see if it works. As a result I get what is in the screenshot, so it seems that there is still another issue:

tryagain

j-fu:
Ok what do we have now on nightly ?
- the "Connection reset" problem goes away whe using dev'd HTTP.jl
- In the notebook we get the following (for each notebook cell):
no-resize

j-fu:
For the `complete_remote_package` problem I used the band-aid proposed in 
https://github.com/fonsp/Pluto.jl/issues/2810#issuecomment-1943750071

I think it may make sense to make a PR to Pkg which marks internal things used by Pluto  as `public` in Julia 1.11 sense, making these public interface.

j-fu:
Ah, nicer stactrace now :)

no-resize2

j-fu:
Now it works with dev'd HTTP.jl - the difference seems to be just one commit which also appears to be about the new Memory type: https://github.com/JuliaWeb/HTTP.jl/commit/040cf0ac773670bbd1ea66e91715583a02537213

Not sure though if returning a view is a good idea. Will need through testing.

As for the Pkg stuff: I would go with @balenamiaa's hack implemented here and fix the Pkg stuff in a different PR when the public interface issues have been resolved.

src/runner/PlutoRunner/src/PlutoRunner.jl
```diff
@@ -1350,7 +1350,7 @@ end
 function show_richest_withreturned(context::IOContext, @nospecialize(args))
     buffer = IOBuffer(; sizehint=0)
     val = show_richest(IOContext(buffer, context), args)
-    return (resize!(buffer.data, buffer.size), val)
```
fonsp, line 1353: This function should return the final `Vector{UInt8}`, not a view, because it will be sent to the Julia server and then to JS.

You could wrap it in `collect`, but I was hoping to avoid the allocations. The performance is important because this is part of the display and messaging pipeline. E.g. when you move a slider that controls a Plots.jl plot, this function gets called on the plot SVG data.

j-fu: `buffer.data` is of now of type `Memory{UInt8}`, which has no `resize!` method. See the "Memory PR" commit.
https://github.com/JuliaLang/julia/commit/909bceae8a60969dfceba4924142e5efee459d47

Not sure what do do here.  `return resize (Vector(buffer.data),buffer.size),val)` seems to be possible, but likewise probably would allocate. 
Didn't look up all the discussions on the Memory stuff, so I don't know if lacking `resize!` is due to concept or due to omission.

fonsp: Fixed in https://github.com/fonsp/Pluto.jl/pull/2828 👍

j-fu:
Guess  it remains to wait now for a new release of HTTP.jl ...

j-fu:
Still need https://github.com/JuliaWeb/HTTP.jl/commit/040cf0ac773670bbd1ea66e91715583a02537213  to be released

j-fu:
worked locally now under 1.6, 1.11.0-alpha1 and nigthly.

fonsp:
Thanks! Let's work on the method table problems in another PR...

fonsp:
The method table issue is here https://github.com/fonsp/Pluto.jl/issues/2840

fonsp:
Thanks again @j-fu ! The other issue was only on Julia 1.12 (nightly), I now changed our CI to run the 1.11 preview and 1.12 nightly separately: https://github.com/fonsp/Pluto.jl/commit/fabde1ff780badd91ac05fb59924679a33ccac2c

The new issue was caused by a change in Julia specifically meant to improve Pluto support! Super cool :) https://github.com/JuliaLang/julia/pull/53415

PR #2821, opened 2024-02-20
fix: make sure mathjax will be bundled

pankgeorg:
The `not-the-src-yet` attribute was, of course, not being bundled. we use the `new URL()` trick of parcel to make sure this will load, keeping the same flow.


fonsp:
The tex file appears in the bundle twice :o

https://github.com/fonsp/Pluto.jl/actions/runs/7972036214/job/21762929218?pr=2821#step:8:181

fonsp:
It seems like MathJax v4 is almost released! And there are already beta releases. It might make sense to make the switch to v4 first before, otherwise we spend quite some effort debugging this, while the v4 will be bundled and included very differently.

fonsp:
Related work in https://github.com/fonsp/Pluto.jl/pull/2802 and https://github.com/fonsp/Pluto.jl/pull/3091

Issue #2417, opened 2022-12-20
Confusing and inconsistent array-of-dicts display


aplavin:
These three arrays of dicts are displayed in totally different ways:
![image](https://user-images.githubusercontent.com/687995/208744203-c1d3b0da-0d0d-4291-a708-f4020f538221.png)
Structurally they are the same, and REPL shows them in very similar ways.

I would expect all of them to display as the first one, with the Pluto  explorer. The second seems outright wrong, and the third just confusing and inconvenient.

More generally, I often find the table viewer too "eager" to engage. Maybe, there should be an override toggle (display as table or with interactive explorer) in each cell output? It would be useful both ways: sometimes tables don't get recognized.

<details><summary>Notebook file</summary>

```julia
### A Pluto.jl notebook ###
# v0.19.18

using Markdown
using InteractiveUtils

# ╔═╡ cd19d0b9-d038-40a3-8b5f-e5ac50af3045
using StructArrays

# ╔═╡ 7abef450-8091-11ed-0d95-5b0ca33aadc4
[Dict(1 => 2), Dict(3 => 4)]

# ╔═╡ f73f0bae-3fd6-48d0-8e8a-5f41d0c538a0
[Dict(1 => 2), Dict(3 => 4)] |> StructArray

# ╔═╡ a95d68d0-b6e0-4845-a5cc-b6afa5c33176
[Dict("a" => 2), Dict("b" => 4)]

# ╔═╡ 00000000-0000-0000-0000-000000000001
PLUTO_PROJECT_TOML_CONTENTS = """
[deps]
StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"

[compat]
StructArrays = "~0.6.14"
"""

# ╔═╡ 00000000-0000-0000-0000-000000000002
PLUTO_MANIFEST_TOML_CONTENTS = """
# This file is machine-generated - editing it directly is not advised

julia_version = "1.8.3"
manifest_format = "2.0"
project_hash = "8e30e14200747e20846286370d32183fac038aac"

[[deps.Adapt]]
deps = ["LinearAlgebra"]
git-tree-sha1 = "195c5505521008abea5aee4f96930717958eac6f"
uuid = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
version = "3.4.0"

[[deps.Artifacts]]
uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"

[[deps.Base64]]
[...]

PR #2824, opened 2024-02-22
Set IOContext for redirected stdout/stderr

### LINKED ISSUE #2823: IOContext not properly set for `stdout`, disables ANSI colors
Since #2148 the `stdout` capture window supports ANSI terminal colors, and `:color => true` is set in `default_stdout_iocontext` as seen here:

https://github.com/fonsp/Pluto.jl/blob/216451c3c839d1c3c6c50a0b69ba1b8297cceb43/src/runner/PlutoRunner/src/PlutoRunner.jl#L1007-L1012

However, when querying `stdout` in a Pluto cell, `:color` is set to `false`, so well-behaved Julia code will not print colors. MWE: `printstyled("hello"; color=:red)`. A workaround is to wrap `stdout` in your own `IOContext` and set `:color => true`, but this should not be necessary.

image
(Please take this screenshot in place of a video. Nothing dynamic here.)

Notebook file: https://gist.github.com/danielwe/0663e49fbfb2219b2230d3fe2092c097

Using Pluto 0.19.39 on Julia 1.10.1

danielwe:
Fixes #2823 

No tests yet

### DIFF
--- a/src/runner/PlutoRunner/src/PlutoRunner.jl
+++ b/src/runner/PlutoRunner/src/PlutoRunner.jl
@@ -2792,8 +2792,8 @@ function with_io_to_logs(f::Function; enabled::Bool=true, loglevel::Logging.LogL
     # Redirect both the `stdout` and `stderr` streams to a single `Pipe` object.
     pipe = Pipe()
     Base.link_pipe!(pipe; reader_supports_async = true, writer_supports_async = true)
-    pe_stdout = pipe.in
-    pe_stderr = pipe.in
+    pe_stdout = IOContext(pipe.in, default_stdout_iocontext)
+    pe_stderr = IOContext(pipe.in, default_stdout_iocontext)
     redirect_stdout(pe_stdout)
     redirect_stderr(pe_stderr)
 
@@ -2826,7 +2826,7 @@ function with_io_to_logs(f::Function; enabled::Bool=true, loglevel::Logging.LogL
     end
 
     # To make the `display` function work.
-    redirect_display = TextDisplay(IOContext(pe_stdout, default_stdout_iocontext))
+    redirect_display = TextDisplay(pe_stdout)
     pushdisplay(redirect_display)
 
     # Run the function `f`, capturing all output that it might have generated.
--- a/src/runner/PlutoRunner/src/PlutoRunner.jl
+++ b/src/runner/PlutoRunner/src/PlutoRunner.jl
@@ -1007,10 +1007,17 @@ const default_iocontext = IOContext(devnull,
     :pluto_with_js_link => (io, callback, on_cancellation) -> core_with_js_link(io, callback, on_cancellation),
 )
 
+# `stdout` mimics a TTY, the only relevant property is :color
 const default_stdout_iocontext = IOContext(devnull, 
-    :color => true, 
-    :limit => true, 
-    :displaysize => (18, 75), 
+    :color => true,
+    :is_pluto => false,
+)
+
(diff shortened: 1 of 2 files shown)


danielwe:
Tests added, loosely following the pattern in `test/Logging.jl`. The idea is to verify that every IOContext property set in `PlutoRunner.default_stdout_iocontext` is visible on `stdout` and `stderr` inside a notebook. (I included `stderr` because `PlutoRunner` sends both streams to the same pipe.) Let me know if this should be done differently.

fonsp:
Thanks! Can you take a look at https://github.com/fonsp/Pluto.jl/pull/2727 and try to find a solution that works for both use cases?

danielwe:
How about like this?

fonsp:
Nice! This seems to match the REPL:

```
julia> IOContext(stdout).dict
Base.ImmutableDict{Symbol, Any} with 1 entry:
  :color => true

julia> struct ShowMyIO end

julia> Base.show(io::IO, m::MIME"text/plain", s::ShowMyIO) = @info "s" IOContext(io).dict

julia> ShowMyIO()
┌ Info: s
│   (IOContext(io)).dict =
│    Base.ImmutableDict{Symbol, Any} with 4 entries:
│      :module                => Main
│      :limit                 => true
│      :last_shown_line_infos => Tuple{String, Int64}[]
└      :color                 => true

```

fonsp:
Could you write the tests in a more end-to-end style? Right now the test is very exact for the question: "did the IO context get set properly"? What I would prefer to test is whether the two issues are fixed:
- Does `printstyled` work? 
- Does `show(bigarray)` and `show(stdout, bigarray)` and `show(stdout, "text/plain", bigarray)` show the full array contents? (e.g. check if the middle element occurs in the string)
- Does `display(bigarray)` limit the display?

It would also be nice if the tests are added to the existing notebook in test/Logging.jl, this makes the tests run faster because no new process needs to be started.

fonsp:
PS Thanks again! Nice implementation :)

fonsp:
@danielwe Hey! Do you have time to take a look at the tests?

danielwe:
Thanks for the feedback! I'm rather strapped for time at the moment. I can probably look at it next week, but if someone wants to adopt this to get it across the finish line sooner, feel free!

fonsp:
anyone should feel free to finish this! just fork the fork :)

ReubenJ:
I've created some end-to-end style tests to cover the cases you mentioned, @fonsp. What's the best way to incorporate these? Just open a new PR?

danielwe:
I just gave you write access to my fork, @ReubenJ, so you can continue using this PR if you wish. Thanks so much for taking this on! When I said "next week" I must have meant next year...

danielwe:
Just push your commits to the `stdoutiocontext` branch in my fork

fonsp:
Nice! Could you test all three forms? show(bigarray) and show(stdout, bigarray) and show(stdout, "text/plain", bigarray)

ReubenJ:
Looks like just one of the Throttled tests is failing. Should be all set otherwise!

fonsp:
Thanks @ReubenJ, amazing!!

Looking forward to your next contribution! Maybe https://github.com/fonsp/Pluto.jl/issues/2417 ? :)

ReubenJ:
Sure thing. I won't have time until a couple of weeks from now, but I'd be happy to then.

danielwe:
Fantastic! :rocket:

PR #2832, opened 2024-02-29
feat: avoid adding packages that exist in the sysimage

pankgeorg:
Experimental. When someone is using a sysimage, `using SysImageModule` triggers a `Pkg.add`. This avoids that. I'm still testing if this is exactly what I have in mind; we probably need to do a special treatment that is similar to stdlibs too.

### DIFF
--- a/src/packages/Packages.jl
+++ b/src/packages/Packages.jl
@@ -106,7 +106,7 @@ function sync_nbpkg_core(
         new_packages = String.(external_package_names(new_topology)) # search all cells for imports and usings
         
         removed = setdiff(old_packages, new_packages)
-        added = setdiff(new_packages, old_packages)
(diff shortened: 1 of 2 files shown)


fonsp:
Don't you want to disable the package manager? Like this but not just for new notebooks? https://github.com/fonsp/Pluto.jl/pull/1631

pankgeorg:
> Don't you want to disable the package manager? Like this but not just for new notebooks? #1631

I think I don't, no. The Package Manager is amazing. There used to be some limitations (which I don't completely understand) that disallowed us from adding _new_ packages and/or doing a correct resolve in the context of a sysimage. But that doesn't seem to be the case any more (1.10.2). We just need to stop the Package Manager from going crazy with packages in the sysimage, and let it do what it does best for the rest. I think this PR does that, but I haven't tested out thoroughly yet.

PR #2862, opened 2024-03-28
Frontend test: instantiate and precompile before

fonsp:


fonsp:
This might modify the fixtures because of the bidirectional syncing...

Issue #1788, opened 2021-12-27
Allow to provide environment at notebook start


j-fu:
I think about something like `Pluto.run(notebook=..., project=... )` . This would start the notebook with the environment provided under `project` instead of the notebook environment.   A similar option to `Pluto.SessionActions.open` and  corresponding methods of PlutoSliderServer.jl would be nice.

The main use case is the use of Pluto notebooks as package tutorials which at once are used in the documentation and in CI. 
- Disable PlutoPkg when working in the tree of a package under development. This allows to  develop notebook and package in parallel and to take advantage of updates in the package via Revise.jl .  
- Eventually run the notebook during unit tests for a dev'ed  package(though this could be achieved by running the notebook as a script) 
- Batch generation of  html pages from Pluto notebooks using a not yet registered version of the package during  documentation development and deployment. 

Outside of these contexts, and without providing an environment, the notebook should run in "reproducible mode" using  PlutoPkg with a fixed version of the package.  
I find it important that switching between these two modes is possible without changes in the notebook
(like  e.g. "markdowning" a cell with `Pkg.activate`.  

In the moment I achieve this behavior in a quite [hackish way](https://github.com/j-fu/VoronoiFVM.jl/blob/25dc8d20606091c4cc45fd470082786085834a70/pluto-examples/nbproto.jl#L7).
The the initial  proposal in #1281 seems to have a similar effect. 

A slightly related discussion is found  [here](https://github.com/fonsp/Pluto.jl/discussions/1345).

j-fu:
See https://discourse.julialang.org/t/julia-1-9-0-beta2-precompilation-taking-almost-twice-as-long/92498/34

j-fu:
Maybe it is just sufficient to have a switch which disables PlutoPkg and runs the notebook in the environment of the caller.

PR #2868, opened 2024-03-29
More embedded Pkg API

fonsp:
Lots of new Pluto API to work with Pkg environments embedded in notebooks

# `activate_notebook_environment`
New method `activate_notebook_environment(fun::Function, notebook_path::String)` with a function argument that you can use inside scripts instead of REPL use. Example:

```julia
Pluto.activate_notebook_environment("notebook.jl") do
    Pkg.add("Example")
end

# Now the file "notebook.jl" was updated!
```

The method `activate_notebook_environment(path::String)` has a new `show_help::Bool` kwarg to disable the help banner.

# `will_use_pluto_pkg `
New function `will_use_pluto_pkg(path::String)` or `will_use_pluto_pkg(notebook::Notebook)` to check if a notebook will use Pluto's package manager. `false` means that the notebook contains `Pkg.activate` or another deactivator.

# More
Added tests for all public API

Fixed `testnb` (not public)


fonsp:
@j-fu This should help with your update scripts!

j-fu:
Cool! Once you are at this - do you see a possibility for #1788  ?

Something like `Pluto.run(notebook="notebook.jl", pkgactivate=pwd())`

PR #2887, opened 2024-04-08
Use <link rel=prefetch> for the statefile

fonsp:
When viewing a static notebook online (like https://featured.plutojl.org/language/structure%20and%20language ), the statefile is downloaded **by the editor** and displayed. But this can be optimized! We can download the statefile in parallel to the editor!

# Before
image

You can see the waterfall effect:  the request for the statefile happens after the editor has loaded.

# Doubts
This article https://web.dev/learn/performance/resource-hints makes me a bit doubtful, it looks like you can easily make the browser download the resource twice... 

I am also not able to debug this very well, I can't find the prefetch/preload in the browser dev tools. Maybe just merge and see if featured.plutojl.org is faster?

Maybe the type of prefetch should be a setting in PlutoSliderServer?

### DIFF
--- a/src/notebook/Export.jl
+++ b/src/notebook/Export.jl
@@ -76,9 +76,9 @@ inserted_html(original_contents::AbstractString;
     """
 )
 
-function prefetch_html(statefile_js::AbstractString)
+function preload_statefile_html(statefile_js::AbstractString)
     if length(statefile_js) < 300 && startswith(statefile_js, '"') && endswith(statefile_js, '"') && !startswith(statefile_js, "\"data:")
-        """\n<link rel="prefetch" href=$(statefile_js)>\n"""
+        """\n<link rel="preload" href=$(statefile_js)>\n"""
     else
         ""
     end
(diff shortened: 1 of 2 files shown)


fonsp:
Before this PR, I did Chrome clear all cache, then Chrome dev tools > Performance > measure reload with Fast 3G network throttle on https://featured.plutojl.org/basic/images

[imagesp1.json](https://github.com/fonsp/Pluto.jl/files/14915390/imagesp1.json)
[imagesp2.json](https://github.com/fonsp/Pluto.jl/files/14915392/imagesp2.json)

image

fonsp:
Aha! I was using `preload` wrong...

`preload` is a better choice than `prefetch` because it is designed for resources loaded in the same page, while `prefetch` is more for resources on the next page after navigation.

I also set the crossorigin settings wrong, but now I followed https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload and it seems to work!

The statefile is being requested even before the editor! And the `fetch` request does not show up in the Network panel as a second request, so that means the preload is working!

image

Issue #2910, opened 2024-05-06
Julia 1.11 beta MemoryRef bounds violation


fonsp:
Pluto does not work on Julia 1.11 because of an error in our Pkg log capture system, but I can't figure out what it is. It seems to be caused by using a `LoggingExtras.ConsoleLogger` with an `IOListener` as its io argument.

When you launch a notebook, you get this error:

```

Instantiating...


Pkg error!

BoundsError: attempt to access MemoryRef{UInt8} at index [1]
Stacktrace:
  [1] GenericMemoryRef
    @ ./boot.jl:522 [inlined]
  [2] unsafe_copyto!
    @ ./genericmemory.jl:108 [inlined]
  [3] copyto!
    @ ./genericmemory.jl:143 [inlined]
  [4] _resize!(io::IOBuffer, sz::Int64)
    @ Base ./iobuffer.jl:309
  [5] ensureroom
    @ ./iobuffer.jl:391 [inlined]
  [6] unsafe_write(to::IOBuffer, p::Ptr{UInt8}, nb::UInt64)
    @ Base ./iobuffer.jl:508
  [7] unsafe_write
    @ ./io.jl:452 [inlined]
  [8] unsafe_write
    @ ./io.jl:803 [inlined]
  [9] write(s::IOContext{IOBuffer}, A::Vector{UInt8})
    @ Base ./io.jl:837
 [10] handle_message(logger::Logging.ConsoleLogger, level::Base.CoreLogging.LogLevel, message::Any, _module::Any, group::Any, id::Any, filepath::Any, line::Any; kwargs...)
    @ Logging /Applications/Julia-1.11.app/Contents/Resources/julia/share/julia/stdlib/v1.11/Logging/src/ConsoleLogger.jl:193
 [11] handle_message(::LoggingExtras.TeeLogger{Tuple{Logging.ConsoleLogger, Logging.ConsoleLogger}}, ::Base.CoreLogging.LogLevel, ::Vararg{Any}; kwargs::@Kwargs{exception::BoundsError})
    @ LoggingExtras ~/.julia/packages/LoggingExtras/VLO3o/src/CompositionalLoggers/tee.jl:24
 [12] invokelatest(::Any, ::Any, ::Vararg{Any}; kwargs::@Kwargs{exception::BoundsError})
    @ Base ./essentials.jl:1032
 [13] macro expansion
    @ ./logging.jl:376 [inlined]
[...]

PR #2915, opened 2024-05-09
Julia 1.11 compat: Use Base.BufferStream for capturing Pkg IO

fonsp:
implement #2910

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -42,10 +42,10 @@ jobs:
             - uses: actions/checkout@v4
 
             # Makes the `julia` command available
-            - uses: julia-actions/setup-julia@v1
+            - uses: julia-actions/setup-julia@v2
               with:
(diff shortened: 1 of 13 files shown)


fonsp:
This PR seems to fix the issue! Except it stalls forever when doing Pkg operations, unless I add lots of logs to debug this, and then it magically works. So I need to figure out which log fixed it and why? (Maybe I need a `yield` or something?)

fonsp:
pankgeorg suggested `flush(some_io)`

fonsp:
The problem was that I was doing `readavailable` on a stream that was already fully read, so it was waiting forever for the next byte. The logs "fixed" it because the logs were also printed in this stream, adding bytes.

fonsp:
This fixed it! There is one more Pkg issue that I will maybe do in a separate PR

fonsp:
last failure is https://github.com/julia-actions/setup-julia/issues/240

PR #2922, opened 2024-05-16
⌨️ Keyboard shortcut for folding cell

### LINKED ISSUE #2917: Shortcut to "Show/hide code"
Hi all,

I use Pluto a lot and I teach two courses with it. In this context, I have to use Markdown extensively. I have a stupid tendinitis that makes using the mouse in repetitive movements quite painful. For people who have this kind of disability, the use of keyboard shortcuts may be precious.

It would be highly beneficial to access the keyboard to **_hide the written and executed Markdown_** instead of using the mouse. I searched and found two cases where this issue has been discussed: [in this thread](https://github.com/fonsp/Pluto.jl/issues/65#issue-595782206) and [here](https://github.com/lucio-cornejo/custom-pluto). @lucio-cornejo's proposal looks very complicated, and this issue was not considered (apparently) in the former thread. In that thread, is the shortcut named as "Fold/unfold cell" intended to do what I am searching for here?  Has it been implemented?

Thanks in advance.

fonsp:
Fix #2917

The implemented keyboard shortcut is ~**`Ctrl/Cmd + K`**~

MacOS: `⌥⌘[` to fold, `⌥⌘]` to unfold
Windows/Linx: `Ctrl+Shift+[` to fold, `Ctrl+Shift+]` to unfold

Loosely matching VS Code

You can do this while editing a cell, or with multiple cells selected

I also made #2920 to make it more clear that folding the cell does not hide the code yet because you are still focused on the codemirror.

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -630,7 +630,7 @@ export const CellInput = ({
         }
         const keyMapFold = (/** @type {EditorView} */ cm, new_value) => {
             set_cm_forced_focus(true)
-            pluto_actions.fold_remote_cells([cell_id], new_value).catch(console.warn)
+            pluto_actions.fold_remote_cells([cell_id], new_value)
             return true
         }
 
@@ -1043,7 +1043,6 @@ const InputContextMenu = ({ on_delete, cell_id, run_cell, skip_as_script, runnin
         if (e.key === "Escape") {
             setOpen(false)
         }
-        e.stopPropagation()
     })
 
     return html`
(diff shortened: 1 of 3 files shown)


fonsp:
Nooo `Cmd+K` is not so good, i misunderstood the vs code thing

But the shortcuts for "folding" "unfolding" code sections might actually be a good fit. I don't think we ever want to implement code folding (ie within cells) in Pluto because cells should be fairly small.

image

Ctrl+Shift+[ ] on windows linux

PR #2927, opened 2024-05-24
Add featured notebooks from plutojl.org/en/docs

fonsp:
This adds some notebooks from plutojl.org/en/docs as extra featured notebooks

image


fonsp:
cc @lukavdplas

PR #2932, opened 2024-06-03
Eris components

bmandracchia:
Notebook customization with Eris components and Jive Sidebar


pankgeorg:
Hello @bmandracchia! It's exciting to see activity based on Pluto!! 

An (off-topic) remark on the name: Eris, apart from the [dwarf planet](https://en.wikipedia.org/wiki/Eris_(dwarf_planet)), is also the godess of strife and discord. Of course, Pluto is also dark; the god of the underworld and the likes. Also Eris is not related to [Iris](https://en.wikipedia.org/wiki/Iris_(mythology)) (🌈deity)

Looking forward to seeing your progress, and I hope we find nice things we can upstream from your efforts!!

Best,
Panagiotis

bmandracchia:
Hello Panagiotis,
Sorry, this commit was meant to customize Pluto in order to use it within another framework we are developing. I don´t think there is anything of use for you, but it would be great if in the future that would be the case!

Yes, I know Eris is the goddess of discord (here we literally call her Discordia) but I wanted to follow the planet naming convention. However, given this commit... it seems now she's goddess of mistakes, too! :)

PR #2949, opened 2024-07-05
enable paths autocompletions

### LINKED ISSUE #2935: No path completions
Pluto 0.19.42, Julia 1.10.4.

[](https://github.com/fonsp/Pluto.jl/assets/9824244/83e96495-6b07-488d-a923-902d2d353790)

Pangoraw:
fixes #2935


fonsp:
Nice! I want path completions to only show when the autocomplete is explicit (TAB was pressed), not automatically on type.

Pangoraw:
ah that makes sense, i made it work only for explicit completions and when the query contains '/' since there are a lot of false positives otherwise.

fonsp:
Really nice, thanks!!

PR #2956, opened 2024-07-13
fix compiler options test windows Julia 1.11

fonsp:
ugh


fonsp:
That fixed it yay

CI still failing because of https://github.com/JunoLab/FuzzyCompletions.jl/pull/20

PR #2966, opened 2024-07-20
Add support for custom name in `@progress` (ProgressLogging.jl)

disberd:
This PR adds support for the visualization of custom names of progress bars created with `@progress` from ProgressLogging.

I also added a small sync icon to the logs of ProgressLogging as there was no icon at the moment.

## Before this PR
image

## After the PR:
image

### DIFF
--- a/frontend/components/Logs.js
+++ b/frontend/components/Logs.js
@@ -88,7 +88,13 @@ export const Logs = ({ logs, line_heights, set_cm_highlighted_line, sanitize_htm
     `
 }
 
-const Progress = ({ progress }) => {
+const Progress = ({ name, progress }) => {
+    console.log("Progress Name: ", name)
+    return html`<pluto-progress-name class=${name === "" ? "no_name" : null}>${name}</pluto-progress-name>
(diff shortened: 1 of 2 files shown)


disberd:
@fonsp
We could also defaults to the name `Progress` in case no name was provided as done in the REPL when using TerminalLoggers.jl
image

I fine with either option, as the important part is to have the name displayed when it's actually provided in my opinion

fonsp:
Nice! I didn't know about names but thats a very useful feature!

I'm not sure about the sync icon, I feel like it's not necessary because a progress bar is familiar. Or do you think this needs to be communicated better?

disberd:
I just thought it felt strange not to have an icon when every other log had one, but I don't have a very strong opinion on this so we can also just ditch it

fonsp (approved):
Okay! Let's remove the icon.

The rest looks good! Two tweaks you can make:
- remove the console log
- use the CSS `:empty` pseudo-class instead of calculating it yourself

disberd:
Thanks, did not know about the `:empty` selector, definitely better.

I removed the icon and applied the suggested changes, with this visual result:
image

fonsp:
Thanks so much! Sorry, my holiday got inbetween 🙈

PR #2973, opened 2024-08-05
Speedup large notebooks: fix cache miss in `notebook_to_js`

### LINKED ISSUE #2958: Substantial slowdown (>x10) opening large notebooks starting from version 0.19.37
My larger notebooks have been opening much slower in more recent versions of Pluto. A large notebook that loads in 200-300 seconds on version 0.19.36 takes around 2500 seconds on 0.19.43. The slowdown seems to start at version 0.19.37 and doesn't affect all notebooks to such an extent, but I couldn't figure out so far what difference is relevant. I've been using 0.19.32 for the last months to avoid this issue.

I could reproduce some similar result with this notebook containing nothing but comments:
https://gist.github.com/yha/4ed9406d2c8b8fd7c50aadccdb5c2982

For this notebook, the initial "Running code" takes 15 seconds on 0.19.36, and 810 seconds on 0.19.43 (The effect is larger with comments than with just empty cells)
image

In both cases, Pluto is running on:
```
julia> versioninfo()
Julia Version 1.10.4
Commit 48d4fd48430 (2024-06-04 10:41 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 48 × Intel(R) Xeon(R) Gold 6248R CPU @ 3.00GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, cascadelake)
Threads: 1 default, 0 interactive, 1 GC (on 48 virtual cores)
```
and the notebook is accessed remotely from Firefox on Windows.

fonsp:
Fix #2958 

We accidentally removed a caching system in https://github.com/fonsp/Pluto.jl/commit/2183b0aaae03bd98cc393e82fffdf8bc0331500c#diff-def3c256c1eef2beff41b35619e1d28209bf7cbb56c223d408f6cf53ed98582dL139-R137 in 0.19.37 that led to a big slowdown in `notebook_to_js` for large notebooks.

This brings the `notebook_to_js` time for a big notebook from `250ms` to `1.85ms`. 

I also changed the `notebook_to_js` to always use the cached topological order, even if a newer one is possible. Just because this field `cell_execution_order` is not even used by us, it's just there for future experiments.

### DIFF
--- a/src/notebook/Notebook.jl
+++ b/src/notebook/Notebook.jl
@@ -31,7 +31,7 @@ Base.@kwdef mutable struct Notebook
     path::String
     notebook_id::UUID=uuid1()
     topology::NotebookTopology
-    _cached_topological_order::Union{Nothing,TopologicalOrder}=nothing
+    _cached_topological_order::TopologicalOrder
     _cached_cell_dependencies_source::Union{Nothing,NotebookTopology}=nothing
 
     # buffer will contain all unfetched updates - must be big enough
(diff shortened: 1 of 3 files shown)


fonsp:
After:

image

Issue #2943, opened 2024-06-25
Update CodeMirror Julia support


savq:
These are the pending tasks to update the CodeMiror Julia thingies:

### Dependencies

Merge PRs and release new versions.

- [x] JuliaPluto/lezer-julia#7
- [x] JuliaPluto/lang-julia#3
- [ ] JuliaPluto/codemirror-pluto-setup#15

JuliaPluto/codemirror-pluto-setup#15 can be merged together with this PR! For now, we can use its latest commit sha to test Pluto with the new versions

### Here

Refactor `frontend/components/CellInput/` modules that depend on lezer stuff.

- [ ] `block_matcher_plugin.js`
  - Works surprisingly well...
- [ ] `pkg_bubble_plugin.js`
  - This can probably be simplified with `ImportPath`
- [ ] `debug_syntax_plugin.js`
- [ ] `scopestate_statefield.js`
- [ ] `lezer_template.js`
- [ ] `mixedParsers.js`
- [ ] new color theme

PR #2980, opened 2024-08-06
Upgrade codemirror with new parser

fonsp:
@savq did incredible work on the Julia parser! https://github.com/JuliaPluto/lezer-julia/pull/7

Fix https://github.com/fonsp/Pluto.jl/issues/2943


fonsp:
@savq mabye you could help with updates to lezer_template and scopestate_statefield? I think the rest is easy enough for me :)

fonsp:
This PR got stuck because of the pattern matching code by dralletje. I gave it a try but this is way to hard to update, I'm declaring it unmaintainable!

Since we barely use the pattern matching and scopestate system, I will look into removing it and rewriting functionality into simpler (less precise) tree traversal code.

fonsp:
Ready! Merging soon :)

PR #2987, opened 2024-08-11
Tidy up frontend

savq:
- Create `fonts/` and `themes/` directories
- Organize style tags
  - Use a single set of style tags for julia, python, and javascript.
- Check mismatches between light/dark themes
  - Some variables are not defined in both files, so I added comments where they're missing.
  - Some variables are probably wrong (copy-pasted were they shouldn't have), but I left them as-is.
  - (Hopefully we can use [`light-dark()`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark) in the near future)
- Rename codemirror color variables

---

I made this PR to make it easier to update the syntax highlighting colors later. It doesn't change any colors.


fonsp:
Awesome, thanks!

PR #2989, opened 2024-08-11
Fix state out of sync issue caused by race condition

### LINKED ISSUE #2904: Cell running forever
Sometimes, in a large notebook, running a cell with many dependencies will leave the cell running forever:

image

But refreshing the browser fixes the issue.

In some cases, the cell also shows up in the Status tab, but not always.

This was Pluto 0.19.41, the cell had classlist `show_input activate_animation running ` and it was `queued=false, running=true` in the statefile.

fonsp:
Fix #2904  🌟 yay!!!

The problem was........

https://github.com/fonsp/Pluto.jl/blob/45be7c15a063b65d0be8f3cb15b3fdf6e5ed83fc/src/webserver/Dynamic.jl#L434

`skip_send` meant that the patch was not being sent 😑

# `skip_send`?
`skip_send` was added in https://github.com/fonsp/Pluto.jl/pull/2273, but right now it looks like that was a mistake. The comment says that it is needed to fix #1892, but #1892 is about bonds, and this change is in the `:run_multiple_cells` function that is not triggered during a bond run.

## #1892?
The original fix from #1892 was also no longer working. I fixed this in a separate PR: https://github.com/fonsp/Pluto.jl/pull/2992 :) But that's probably not related.

# `queued = true`?
So why is this code here? This is from https://github.com/fonsp/Pluto.jl/pull/710/files#diff-d0b76cc753bebaa7bbde63848adae9617f1e69dd9cd9986b98b7a3c5e39fc02bR360-R372

When a cell is triggered with Shift+Enter, its state is set to `queued = true` immediately on the frontend. And this code does the same in the backend.

# Solution
Instead of calling `send_notebook_changes!` with `skip_send=true` to go through the client state update logic, we now just update the client state dictionary directly.


fonsp:
`run_threaded` does not make a difference, also with `julia --threads=auto`

fonsp:
TODO:
move outbox back
change big comment to not talk about 1891

Issue #2978, opened 2024-08-05
Performance: throttled status update loop impacts performance on large notebooks


fonsp:
In large notebooks, the throttled update loop from the status tree, i.e.:

https://github.com/fonsp/Pluto.jl/blob/8b7a65d7e191eb404d6ecd5c479b40d7039fb1d4/src/webserver/SessionActions.jl#L189-L192

Can slow down the notebook. When the notebook is large, and you are moving sliders, this can halve the update speed.

Ideas:
- During reactive run, share the throttled update with the new one.
- Throttle rate should adapt to the time it takes to update clients.
- It also looks like there is a bug in the `throttle` code: the throttled code runs in batches of 2...

PR #2992, opened 2024-08-12
Fix intermediate bond values again

### LINKED ISSUE #1891: Too many intermediate bond values
Our bond values iterator is *lossy*, meaning that we don't send intermediate bond values **while the notebook is running**. This is important, because:
https://github.com/fonsp/Pluto.jl/blob/d8bf85935778f0a831ddd7058ad2519639bfc772/sample/Interactivity.jl#L206

But recently, it has not been working very well 👀

# Example 1

https://user-images.githubusercontent.com/6933510/152383314-197d51da-f6c1-4e9c-9363-9025b00bd88a.mov

You can see that it "keeps going for too long", because the server is catching up with previous bond values.

# Example 2

Here is another example, a button that sends 60 values in one second when you click. The first cell should only show `0` and `59`, no intermediate values.

https://user-images.githubusercontent.com/6933510/152383485-d8b79069-7f1d-4d37-8167-29cbbaa61f10.mov

<details>
<summary>Code</summary>

```julia
# ╔═╡ 3bee5e56-302d-4b50-86c3-42b9229304f1
@bind laksjdflkjasdf Slider(1:10000)

# ╔═╡ 5b2865a0-edd9-4a92-b349-5b9a682da49e
sleep(1); laksjdflkjasdf

# ╔═╡ 5247aa82-5099-49c6-b967-2a3cec0cd5e9
sleep(1); zzz

# ╔═╡ a3a63f07-9c71-4af1-a65b-2eb1a7567d02
@bind zzz @htl("""
<div>
	<button>Click me</button>
	<span></span>
	

	<script id="zzz">
	const div = currentScript.parentElement
	
	const button = div.firstElementChild
	const span = div.querySelector("span")

	div.value = "click to start!"

	if(this == null) {
		button.addEventListener("click", (e) => {
			let counts = 0
			let f = () => {
				div.value = counts
				div.dispatchEvent(new CustomEvent("input"))
		
   				counts++
				span.innerText = counts
				if(counts < 60){
				   requestAnimationFrame(f)
				}
			}
			f()
			e.stopPropagation()
		})
	}

	return div
	</script>
</div>
""")
```

</details>

# Cause

See #1892

fonsp:
This will fix #1891 again which got reopened by #2399

This PR actually contains two fixes! So now it's superfixed!!!
1. The heuristic from #1892 was updated to ignore patches related to `status_tree`.
2. I continued https://github.com/fonsp/Pluto.jl/issues/2978 and made sure that the status tree sync loop is also throttled at the start of the reactive run. #1891 was happening because of the initial status updates.

This time the fix also includes an end-to-end test with puppeteer :)

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -805,7 +805,9 @@ patch: ${JSON.stringify(
                         const set_waiting = () => {
                             let from_update = message?.response?.update_went_well != null
                             let is_just_acknowledgement = from_update && message.patches.length === 0
-                            // console.log("Received patches!", message.patches, message.response, is_just_acknowledgement)
+                            let is_relevant_for_bonds = message.patches.some(({ path }) => path.length === 0 || path[0] !== "status_tree")
+
+                            // console.debug("Received patches!", is_just_acknowledgement, is_relevant_for_bonds, message.patches, message.response)
 
                             if (!is_just_acknowledgement) {
                                 this.waiting_for_bond_to_trigger_execution = false
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -812,7 +812,7 @@ patch: ${JSON.stringify(
                             }
                         }
                         apply_promise.finally(set_waiting).then(() => {
-                            this.send_queued_bond_changes()
+                            this.maybe_send_queued_bond_changes()
                         })
 
                         break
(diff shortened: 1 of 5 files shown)


fonsp:
oops some last changes in https://github.com/fonsp/Pluto.jl/commit/fb2f187f0d5ec9ea3eb2fe10490e57624dcce15e

PR #2993, opened 2024-08-12
Upgrade immer to 10.1.1

### LINKED ISSUE #2988: Update immer.js
v10 is out and we use v8. Should be 33% faster

fonsp:
fix #2988


fonsp:
I reverted this in https://github.com/fonsp/Pluto.jl/commit/7354eb8656cb147b8621841d26a4989d1c2b7257

Issue #3012, opened 2024-08-25
Error in fallback path for catching logging errors


danielwe:
GPUCompiler.jl implements slightly tweaked loggers for safe use in generated functions:

https://github.com/JuliaGPU/GPUCompiler.jl/blob/27d4b190243f9a4440e8c0bfda876af89dd1140e/src/utils.jl#L55-L73

Pluto.jl fails to display these logs, instead erroring with `ArgumentError: Base.TTY(RawFD(4294967295) invalid status, 0 bytes waiting) is not initialized`.

MWE notebook (this does not require a GPU, GPUCompiler is only used via Enzyme): https://gist.github.com/danielwe/c30299b9c11bb0cc5ce182913b640da3

In the REPL this emits a warning:

```julia-repl
julia> include("pluto_gpucompiler_error.jl");
┌ Warning: Using fallback BLAS replacements for (["drot_64_"]), performance may be degraded
└ @ Enzyme.Compiler ~/.julia/packages/GPUCompiler/Y4hSX/src/utils.jl:59
dx = [-0.8660254037844386]
```

In Pluto it errors:

[Screencast from 08-25-2024 11:33:19 AM.webm](https://github.com/user-attachments/assets/b8a1ac7d-842e-4511-b9a1-1a1bb107fcc9)

wsmoses:
cc @vchuravy didn't michel or someone else hit this?

danielwe:
The displayed error comes from printing to `stderr` in the catch block of `handle_message`, here: https://github.com/fonsp/Pluto.jl/blob/aaa835ba8ba3318f46231af354fb58dd29cc91c7/src/runner/PlutoRunner/src/io/logging.jl#L116

So I guess there's actually two issues here, one from failing to display the log, sending us to the catch block, and one from failing to print to `stderr` in the catch block

danielwe:
As an experiment, I inserted `error()` inside the try block in `handle_message`, such that every logging attempt ends up in the catch block. The result is that any attempt at logging inside a notebook, like `@info "hello"`, produces the same error as above: `ArgumentError: Base.TTY(RawFD(4294967295) invalid status, 0 bytes waiting) is not initialized`.

In other words, the catch block is broken. The `const original_stderr = stderr` apparently isn't a valid, writable IO when `handle_message` is called.

Then there's the more fundamental issue of why GPUCompiler logs end up in the catch block in the first place. But fixing the catch block would make this a lot easier to debug.

[...]

PR #3013, opened 2024-08-27
Fix fallback for errors in logging

danielwe:
This fixes the bug identified in #3012, in which the catch block that's supposed to save you in case an error is thrown during logging ends up throwing an error of its own.

* `original_stderr` is now assigned at runtime through the module's `__init__` function, not at compile time as a hardcoded global constant. This was the source of the bug.
* I rewrote the catch block's print/showerror as an `@error` log, which stands out more with the red frame and also looks more consistent with the other information logged to the REPL by Pluto.
* Capturing stderr output for testing is not entirely straightforward. There are tradeoffs between using a local and remote worker. Please review the comments in the testing code and let me know which approach is preferred (or if there's a better design I haven't thought of).

Having looked more deeply at #3012, I think this PR is a sufficient fix on the Pluto side. The reason GPUCompiler's log messages error in the first place is that the purportedly safe logging macros are implemented incorrectly and aren't actually safe. I'll put together a PR to GPUCompiler for that.

### DIFF
--- a/src/runner/PlutoRunner/src/PlutoRunner.jl
+++ b/src/runner/PlutoRunner/src/PlutoRunner.jl
@@ -76,4 +76,6 @@ include("./io/logging.jl")
 include("./io/stdout.jl")
 include("./precompile.jl")
 
+__init__() = redirect_original_stderr(stderr)
+
 end
--- a/src/runner/PlutoRunner/src/PlutoRunner.jl
+++ b/src/runner/PlutoRunner/src/PlutoRunner.jl
@@ -76,6 +76,8 @@ include("./io/logging.jl")
 include("./io/stdout.jl")
 include("./precompile.jl")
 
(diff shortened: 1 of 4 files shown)


danielwe:
Apparently, capturing stderr from the remote worker doesn't work on either Windows or julia 1.6; the latter because `redirect_stderr(::Pipe)` isn't supported in that version, the former for unknown reasons. I guess a local worker is the way to go for testing then. The latest commit implements that.

fonsp:
Thanks for the research and the nice solution! Looks good!

(I changed the PR a bit to move the `redirect_original_stderr` functionality from the PlutoRunner source to the test.)

fonsp (approved):

danielwe:
The changes look good to me. Anything else you need before merging?

Issue #2737, opened 2023-12-06
Live Docs: show package defining a function


fonsp:
When reading docs, it would be nice to show from which package a function/method originates. 

Current:

<img width="420" alt="image" src="https://github.com/fonsp/Pluto.jl/assets/6933510/16a65df8-cea3-489e-a31a-e1526cd34e54">

fonsp:
I did some experiments here, and the big question in this issue is on the Julia side:

Currently, we use `Docs.doc(binding)` to get documentation, but this just returns a `Markdown.MD`, but we are not sure how to get more info, such as:
- For how many methods did you show docs? E.g. docs for `sqrt` gives two entries for two methods.
- For each method:
  - In which package is it defined?
  - Can we get the `Method` object? (And link to the source code?)

cstjean:
Why not evaluate the symbol (in a try/catch), and show its `methods(f)` if it is a function?

Incidentally, because I have `JULIA_EDITOR` configured, I can take advantage of

![{4982748C-D3F4-41E7-B180-AF28308EF54D}](https://github.com/user-attachments/assets/1f1b4865-3eba-44a5-900e-a41ee29ade12)

or type `edit(some_function)`. It is super useful for debugging in the REPL, and there's no reason why Pluto couldn't link its stacktraces in the same way. I can figure out the relevant methods if that helps.

fonsp:
Yes amazing! For stacktraces we are have this already implemented: https://github.com/fonsp/Pluto.jl/pull/2813 and we are working on supporting URLs for more methods in https://github.com/JuliaPluto/MethodURL.jl (see https://github.com/JuliaLang/julia/issues/47709)

fonsp:
Evaluating in trycatch and getting `methods` is a first step, the question is:

`Docs.doc` gives all docs for a search string as one single `Markdown.MD` object. And this object often contains docs for multiple methods. 

What we want is to get something like:
```julia
[
	Dict(
		:method => "sqrt(::Matrix)",
		:method_url => "https://...",
		:docs => "<p>..."
	),
	Dict(
		:method => "sqrt(::Real)",
		...
	)
]
```

We can give this list to the frontend and show documentation for multiple methods in a nicer way, and include URLs to source code. I.e. the goal is to get more structured information than what `Docs.doc` currently provides, because that structured data allows the frontend to be more helpful.

cstjean:
>  For stacktraces we are have this already implemented

[...]

PR #3038, opened 2024-10-02
Stack trace URL improvements

fonsp:
New features:
- Clickable URLs for functions from packages (thanks to @adrhill !!)
- Show package name for external functions
- Icon to show that the link will open in a new tab
- Some CSS tweaks

image

image


fonsp:
The new URL code does not work on Julia 1.6, and I don't see an easy way to make it work :(

fonsp:
It might be worth trying this code https://github.com/JuliaLang/julia/issues/47709#issuecomment-2388629772 ?

cstjean:
The clickable URL feature is what I'm talking about in https://github.com/fonsp/Pluto.jl/issues/2737#issuecomment-2450879776. I would selfishly prefer JULIA_EDITOR over an URL because
 - It works even if offline
 - It works for local packages
 - It's always pointing to the right line, even for old versions.
 - My editor is comfier than github.
I know you favor simplicity and what's-best-for-students, so I'm not going to hold my hopes up. This PR as is looks great!

fonsp:
We can't launch an editor from Pluto because it's not a terminal app.

fonsp:
I think launching an editor doesnt fit our target audience

cstjean:
> I think launching an editor doesnt fit our target audience

Yes, that's what I thought you would say 🙂. Oh well! I agree that for students the URL works out better.

You could in theory favor JULIA_EDITOR if it's defined (since it's undefined by default) and fallback to the URL, but it's ugly logic and complexity.

PR #3039, opened 2024-10-02
Stack frame tweaks

fonsp:
imrpvoements from #3038 without the URL bits

- show package name
- collapse super long type signatures
- fade out frames that are probably not important
- button to read docs
- tweaks colors and spacing

image

image


cstjean:
That's great, thank you for the continued improvements!

PR #3041, opened 2024-10-03
Error message title

fonsp:
New title above the error: "Error message"

If we can deduce which package threw the error, we say "Error message from Example"

image

image

image


cstjean:
❤️ 

Showing the very common `no method matching` error in HTML (with linked methods) might be another improvement.

fonsp:
Yes! I think that's the #1 error to make improvements to :)

PR #3057, opened 2024-10-14
docs: use JuliaMono

Pangoraw:
documenter has since switch to JuliaMono for code blocks.

https://github.com/JuliaDocs/Documenter.jl/blob/a0257f939c2be20c7a68ecc02c4ada8210f29495/assets/html/scss/documenter-dark.scss#L18
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="docs-julia-mono")
julia> using Pluto
```

### DIFF
--- a/frontend/editor.css
+++ b/frontend/editor.css
@@ -2551,7 +2551,8 @@ body > pluto-helpbox > header > button:is(.helpbox-close, .helpbox-popout) {
 .helpbox-docs code,
 .helpbox-docs .cm-line {
     /* from https://docs.julialang.org/en/v1/assets/themes/documenter-light.css */
-    font-family: "Roboto Mono", "SFMono-Regular", "Menlo", "Consolas", "Liberation Mono", "DejaVu Sans Mono", monospace;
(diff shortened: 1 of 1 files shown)


frontend/editor.css
```diff
@@ -2551,7 +2551,8 @@ body > pluto-helpbox > header > button:is(.helpbox-close, .helpbox-popout) {
 .helpbox-docs code,
 .helpbox-docs .cm-line {
     /* from https://docs.julialang.org/en/v1/assets/themes/documenter-light.css */
-    font-family: "Roboto Mono", "SFMono-Regular", "Menlo", "Consolas", "Liberation Mono", "DejaVu Sans Mono", monospace;
+    font-family: "JuliaMono", "SFMono-Regular", "Menlo", "Consolas", "Liberation Mono", "DejaVu Sans Mono", monospace;
```
fonsp, line 2554:
```suggestion
    font-family: var(--julia-mono-font-stack);
```

It's a different stack from documenter but this one has fallback fonts that look more like juliamono

Issue #2970, opened 2024-07-29
Adding One-Shot cells


oschulz:
Following up on a conversation with @fonsp at JuliaCon I'm creating this issue based on an [older discussion](https://github.com/fonsp/Pluto.jl/discussions?discussions_q=one-shot):

Using Pluto for use cases the include long(er) running computations is currently a bit tricky:

(In the following, I use "above" and "below" in the sense in dependency-order, not notebook layout cell-order.)

If some Pluto cells have run times on the order of minutes (e.g. statistical inference, machine learning training, etc.), and those (typicall few) cells depends on several input/configuration/parameter values "above", then the cell will be triggered again and again while users fiddle with those "inputs". This can make their system really slow, drain their battery, or (when run on large systems that use accounting) come with resource costs in terms of compute credits or possibly actual money (e.g. when the cells spawns workers on on-demand cloud-resources).

We've had the ability to disable cells temporarily for a while now, which can come in handy. But it doesn't really address the "a few very expensive cells" scenario - one has to manually disable and re-enable the right cells again and again, and while they are disabled everything "below" is grayed out.

Could we add "single-shot" cells to Pluto? They would behave similar to cells that immediately disable themselves after they've been run. But in contrast to actually disabled cells, the content "below" would not be grayed out immediately - not until is becomes invalid - i.e. until the "one-shot" cell should be run again because the values "above" have changed. When that happens, the "one-shot" cell could be marked red, get a big "play" button overlay, or something like that, and everything "below" would be grayed-out - clearly showing that what's "below" is not consistent with what's "above" anymore.

I think this would be in keeping with Pluto philosophy - this is not about adding Jupyter-like cells, it's more like the "single-trigger" mode on an oscilloscope (we already have "continuous trigger" and "trigger off", so to speak).

@Pangoraw came up with a [partial workaround based @use_memo](https://github.com/fonsp/Pluto.jl/discussions/2574#discussioncomment-6035273) and @SimonLafran [recently posted](https://github.com/fonsp/Pluto.jl/discussions/2574#discussioncomment-9653292) a macro-based approach to this (thanks!!), but it can't mark dependent cells as outdated. Having this as a Pluto built-in feature would be great, in my opionion.

cstjean:
https://github.com/fonsp/Pluto.jl/issues/1079 would be another solution to this for us. The problem isn't necessarily that $LONG_COMPUTATION is rerun, but rather that it is blocking the evaluation of everything else. 

[...]

PR #3068, opened 2024-10-21
Add stale cells (aka single-shot cells)

SimonLafran:
As discussed in #2970 and previous discussions, having a way to temporarily block the execution of long-running cells would be beneficial.

This implementation goes for a very flexible solution: `Main.PlutoRunner.Stale`. When returned from a cell, this struct passes down its formatting to its single member `out`. However, the cell also gets marked as stale. This means that none of the cells that depend on it will be run. The stale cell, however, will still be run when its dependencies run, to see if its no longer stale.

On the frontend stale outputs are displayed on a light or dark orange background while stale cells and cells that depend on them get orange traffic lights.

When a cell is in one of those two states a warning sign with an explanatory popup is displayed below the "fold code" button.

(see picture below)

Picture of frontend rendering

## Small details
- We might want to change the frontend formatting a bit. Maybe dim the cells like when they are disabled.
- Implementation-wise, I decided not to touch the topology when running the cells, because of the dynamic nature of the stale state. I basically only check if a cell depends on stale ones before running it. We could pro-actively exclude all dependents of a cell that returns Stale.

## Future
We could add utility macros to PlutoHooks/PlutoLinks, something like `single-shot`. This would show how to use the marker.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/SimonLafran/Pluto.jl", rev="stale-cells")
julia> using Pluto
```

### DIFF
--- a/frontend/components/Cell.js
+++ b/frontend/components/Cell.js
@@ -105,7 +105,19 @@ const on_jump = (hasBarrier, pluto_actions, cell_id) => () => {
  * */
 export const Cell = ({
     cell_input: { cell_id, code, code_folded, metadata },
-    cell_result: { queued, running, runtime, errored, output, logs, published_object_keys, depends_on_disabled_cells, depends_on_skipped_cells },
+    cell_result: {
+        queued,
+        running,
+        runtime,
+        errored,
+        output,
+        logs,
+        published_object_keys,
+        depends_on_disabled_cells,
+        depends_on_skipped_cells,
+        stale,
+        depends_on_stale_cells,
+    },
     cell_dependencies,
     cell_input_local,
     notebook_id,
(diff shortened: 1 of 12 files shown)


fonsp:
Hey @SimonLafran! Let's have a call? Send me an email on fons@plutojl.org

fonsp:
Notes from our call:

# Integrated approach 

The single shot feature can be integrated into the existing disabled-cells feature!

Disabled cells will get a button to "run once" (and stay disabled).

For this, we would also improve the existing disabled feature:
- Display: the current 50% opacity display is an accessibility issue. We need an alternative with full visibility like implemented in this PR. It could be an icy frozen 🥶 look
- Dirty: when a disabled cell becomes dirty (one of its dependencies re-runs), this should be visible. So you know that it needs a re-run.

## Implementation 
The "run" command from the frontend can contain a list of single-shot ids. The disabled status of those cells will be ignored during the run.
If simon sees opportunities to improve our implementation of disabled cells in the meantime then great!!

## Conflict
It's possible that a cell depends on two disabled cells, and you single-shot one of them. We need to decide what to do here.

fps:
I am really interested in this feature. :)

Issue #2781, opened 2024-01-17
Periodic Memory Management


pankgeorg:
Memory management on a Garbage Collected language is something people usually don't think about. Julia is a _little_ bit different, in the sense that there isn't much memory monitoring going on the process, until the process starts feeling some memory pressure.

When the process does start to feel that memory pressure, it will aggressively run the GC _when an allocation happens_. This is not the time you'd want to run the GC, when a web request is inflight. Until julia is better at this (and _if_, because this approach is pretty good for a lot of use cases that `julia` addresses right now, e.g. doing math) we need to manage the GC a bit more manually.

fonsp:
Nice! Do you mean on the notebook process (https://github.com/fonsp/Pluto.jl/issues/1850) or the Pluto server? cc @BioTurboNick

fonsp:
I think a useful first step is a MWE where memory goes up and a reproducible way to measure it

pankgeorg:
> I think a useful first step is a MWE where memory goes up and a reproducible way to measure it

Good point. The first step is adding observability, which is a whole new issue I'm going to open. Since OpenTelemetry seems to have a [memory leak](https://github.com/oolong-dev/OpenTelemetry.jl/issues/95#issue-1983567696) I wouldn't use that, but we can figure something simpler out.

Issue #2691, opened 2023-10-30
Preserve cell order in Safe preview


fonsp:
> *This is a good first issue, and requires no JavaScript knowledge!* 

We recently launched **Safe preview** (#2563, take a look), but it has one missing feature!

When you open a notebook without running it, and you make a small change (eg fix a typo in a cell), Pluto will save the file, but the order of cells in the file can be drastically different, and you get a big git diff. Try this with one of your more complex notebooks to see what I mean, and take a look at which cells get moved and why.

> This is also an issue currently in some more niche areas:
> - Pluto.activate_notebook_environment and Pluto.update_notebook_environment will modify the embedded pkg info (yay) but also mess up the cell order (boo).
> - If you open a notebook and shut it down before it finished, you have the same problem.

# The cause
Pluto saves cells in the `.jl` file in an order that allows the cells to be executed as a standalone script. It essentially takes the visual cell order, and then we do a [stable sort](https://en.wikipedia.org/wiki/Category:Stable_sorts) based on cell dependencies.

The reason why the issue happens is that Pluto often needs to evaluate your code to fully understand cell dependencies. The main examples are 
- `using Plots` defines the symbol `scatter`
- a macro can create or remove variable definitions and references, e.g. `@gensym x` looks like it references `x`, but it actually defines `x`

You cannot derive this info statically, so we run code.

When you open a notebook without running it, we only have information that can be derived statically. When we save your notebook, we do the best we can using that information, but this can lead to a very different file order than what was previously derived (based on complete information), leading to a large diff.

# Solution idea
My idea for solving this is: when we load a notebook file, also remember the original order in which cells appear in the file. Whenever you save a notebook, check if we have enough information to fully derive the order. If so, remove the memorized order and do as usual. If not, try to stick to the original file order as much as possible (i.e. use the original file order as a way to disambiguate the unknown cell deps).

# Relevant code is here:
Pluto's reactivity algorithm. The core is this function that sorts a list of cells into search order
https://github.com/fonsp/Pluto.jl/blob/v0.19.29-src/src/analysis/topological_order.jl#L10-L137

The output of this function is:
https://github.com/fonsp/Pluto.jl/blob/v0.19.29-src/src/analysis/TopologicalOrder.jl

One of the inputs of this function is:
https://github.com/fonsp/Pluto.jl/blob/v0.19.29-src/src/analysis/Topology.jl#L28-L36
[...]

PR #3072, opened 2024-10-23
Make JuliaFormatter ignore bind macro in notebook file

### LINKED ISSUE #2623: Make Pluto boilerplate formatted according to JuliaFormatter
This is a very minor thing which has annoyed me for some time. It would be nice if the boilerplate added by Pluto was formatted in such a way that running JuliaFormatter on the notebook file didn't reformat anything. 

The only issue at the moment is [this line](https://github.com/fonsp/Pluto.jl/blob/1d2c3038032108fff3a9c565eaa84609f1e75b4e/src/runner/PlutoRunner.jl#L2116C1-L2123C7), i.e
```julia
macro bind(def, element)
    quote
        local iv = try Base.loaded_modules[Base.PkgId(Base.UUID("6e696c72-6542-2067-7265-42206c756150"), "AbstractPlutoDingetjes")].Bonds.initial_value catch; b -> missing; end
        local el = $(esc(element))
        global $(esc(def)) = Core.applicable(Base.get, el) ? Base.get(el) : iv(el)
        el
    end
end
```
which gets formatted to
```julia
macro bind(def, element)
    quote
        local iv = try
            Base.loaded_modules[Base.PkgId(
                Base.UUID("6e696c72-6542-2067-7265-42206c756150"),
                "AbstractPlutoDingetjes",
            )].Bonds.initial_value
        catch
            b -> missing
        end
        local el = $(esc(element))
        global $(esc(def)) = Core.applicable(Base.get, el) ? Base.get(el) : iv(el)
        el
    end
end
```

To reproduce you can create a new notebook and add
```julia
begin 
	using PlutoUI
	@bind x Slider(1:10)
end
```
Save the notebook and format it with `JuliaFormatter.format()`. It should return false, meaning that the file was not already correctly formatted. Opening the notebook again, closing it and then formatting it again gives false. 

The reason this is annoying for me is that if you format a Pluto notebook with JuliaFormatter and then open it with Pluto the above code gets "unformatted" again. It of course doesn't affect the result of the code but if the notebook is committed to a git repository you get a modified file.

christiangnrd:
This is regularly the only change when I format my code/notebooks, and it gets reset whenever the notebook is saved. By telling JuliaFormatter to ignore it, it reduces noise in commits (if accidentally committed) and means I don't have to worry about formatting code I didn't write nor will I ever see.

I believe this is a good solution to #2623 as it solves the issue of files being constantly reformatted for those who use JuliaFormatter, and it does not affect those who do not. (Other than 2 extra lines in the notebook)

Closes #2623

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/christiangnrd/Pluto.jl", rev="patch-2")
julia> using Pluto
```

### DIFF
--- a/src/notebook/saving and loading.jl
+++ b/src/notebook/saving and loading.jl
@@ -50,10 +50,8 @@ function save_notebook(io::IO, notebook::Notebook)
     # Super Advanced Code Analysis™ to add the @bind macro to the saved file if it's used somewhere.
     if any(!must_be_commented_in_file(c) && occursin("@bind", c.code) for c in notebook.cells)
         println(io, "")
-        println(io, "#! format: off") # Turn off formatting. See https://domluna.github.io/JuliaFormatter.jl/stable/#Turn-off/on-formatting
         println(io, "# This Pluto notebook uses @bind for interactivity. When running this notebook outside of Pluto, the following 'mock version' of @bind gives bound variables a default value (instead of an error).")
         println(io, PlutoRunner.fake_bind)
-        println(io, "#! format: on") # Turn formatting back on.
     end
     println(io)
 
@@ -93,6 +91,7 @@ function save_notebook(io::IO, notebook::Notebook)
         end
     end
 
+    println(io, "#! format: off") # Turn off formatting for the rest of the document.
 
     using_plutopkg = notebook.nbpkg_ctx !== nothing
 
--- a/src/notebook/saving and loading.jl
+++ b/src/notebook/saving and loading.jl
@@ -32,7 +32,7 @@ Have a look at our [JuliaCon 2020 presentation](https://youtu.be/IAF8DjrQSSk?t=1
 function save_notebook(io::IO, notebook::Notebook)
     println(io, _notebook_header)
     println(io, "# ", PLUTO_VERSION_STR)
-    
+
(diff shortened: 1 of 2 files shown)


fonsp:
Nice! Can you add a comment to the code that links to https://domluna.github.io/JuliaFormatter.jl/stable/#Turn-off/on-formatting ?

christiangnrd:
@fonsp Done!

fonsp:
Hey! Actually can you put the comments inside the fake macro? That makes it more clear that it's part of the macro and not the whole file

image

christiangnrd:
@fonsp Done and tested.

fonsp:
Awesome! Did you also test that JuliaFormatter works for the rest of the file? i.e. that `format: on` also works

christiangnrd:
I did!

fonsp:
Awesome! Maybe you're interested in a bigger project?

Like https://github.com/fonsp/Pluto.jl/issues/2691 🌟

Or research for https://github.com/fonsp/Pluto.jl/issues/2781 would be really useful!

These are some Julia-only projects, but if you are also interested in using JavaScript/React then let me know!

PR #3106, opened 2024-11-25
Make fake_bind compatible to runic.

j-fu:
[Runic](https://github.com/fredrikekre/Runic.jl) is a new "runic" formatter  for Julia code.
Due to peculiarities of implementation, format toggling is better handeled by wrapping around whole functions or macros, see  

See https://github.com/fredrikekre/Runic.jl/issues/107#issuecomment-2498098108

So in order to make Pluto notebooks invariant under runic formatting I propose to modify `fake_bind` accordingly. 

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/j-fu/Pluto.jl", rev="runic_fake_bind")
julia> using Pluto
```

### DIFF
--- a/src/runner/PlutoRunner/src/bonds.jl
+++ b/src/runner/PlutoRunner/src/bonds.jl
@@ -133,16 +133,16 @@ end
 
 """
 Will be inserted in saved notebooks that use the @bind macro, make sure that they still contain legal syntax when executed as a vanilla Julia script. Overloading `Base.get` for custom UI objects gives bound variables a sensible value.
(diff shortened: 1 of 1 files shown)


fredrikekre:
(Simply adding `return` in front of the quote would be an alternative fix if you don't want to move the comments to the outside of the macro.)

fonsp:
Let's go with the `return` then! See https://github.com/fonsp/Pluto.jl/pull/3072 for previous discussion

pjaap:
@j-fu ping

fonsp:
Thanks! Can you check if my last commit is correct? Or was the extra newline intentional?

j-fu:
LGTM

PR #3134, opened 2025-01-14
New API: Pluto.is_single_expression

fonsp:
To check if a `String` contains a single expression.

If this function returns `false`, then Pluto will show a "multiple expressions in one cell" error in the editor.

See tests in the PR files
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="is_single_expression")
julia> using Pluto
```

### DIFF
--- a/src/analysis/Parse.jl
+++ b/src/analysis/Parse.jl
@@ -119,6 +119,24 @@ end
 preprocess_expr(val::Any) = val
 
 
+"""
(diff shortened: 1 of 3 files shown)


contra-bit:
Thanks, every nice. One step closer to never opening julia  code in jupyter again, thanks to the automatic conversions :)

PR #3145, opened 2025-02-13
🐭 Use scratch space for notebook package environment directory

### LINKED ISSUE #2641: Feature: Detect if temporary folder has been deleted and require notebook restart
If a notebook is left running for a really long time, the temporary folder storing its environment can get cleaned up by the system.

This leaves the notebook in a broken state, but it will manifest in unusual ways, such as by UndefVarErrors due to missing packages that should have been loaded.

It would be helpful if Pluto could detect if the environment folder has been deleted and require the notebook to be restarted.

fonsp:
The directory where Project.toml + Manifest.toml is (secretly) stored for a running notebook is current a `mktempdir()` folder. This PR uses Scratch.jl for that, to fix #2641 
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="scratch-for-notebook-env")
julia> using Pluto
```


fonsp:
This can still lead to issues when running a notebook for more than 7 days...

fonsp:
This needs an async loop to "refresh" all current scratch spaces...

PR #3156, opened 2025-03-17
Generate html in presentation mode

blegat:
I've been trying to export notebooks to pdf on presentation modes with PlutoPDF.
With this option, it works. It does print a few blank pages but these can easily be removed in post-processing.
This PR feels a bit hacky, expecially with the `delay` that is there because `present()` need to be run after all js libraries have been loaded.
Ideas for a better solution are most welcomed.
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/blegat/Pluto.jl", rev="present")
julia> using Pluto
```

### DIFF
--- a/src/notebook/Export.jl
+++ b/src/notebook/Export.jl
@@ -99,27 +99,51 @@ This function takes the `editor.html` file from Pluto's source code, and uses st
 See [PlutoSliderServer.jl](https://github.com/JuliaPluto/PlutoSliderServer.jl) if you are interested in exporting notebooks programatically.
 """
 function generate_html(;
-        version::Union{Nothing,VersionNumber,AbstractString}=nothing, 
+        version::Union{Nothing,VersionNumber,AbstractString}=nothing,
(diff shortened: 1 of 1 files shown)


pankgeorg:
Hello Benoît! Thanks for the contribution! I'll take a look for the presentation mode, it does feel a bit hacky, but we're not against that per-se!

Best,
P

fonsp:
You can probably use the `header_html` or `preamble_html_js` kwarg to achieve this:

```julia
generate_html(;
	...
	header_html = """
	<script>
	// your script
	</script>
	""",
)
```

Actually `preamble_html_js` might be even better, you won't need to await page load. Note the extra quote:

```julia
preamble_html_js = """`
	<script>
	// your script
	</script>
`"""
```

fonsp:
closing in favour of https://github.com/JuliaPluto/PlutoPDF.jl/pull/29

PR #3157, opened 2025-03-19
Add support for TestEnv.activate()

### LINKED ISSUE #2775: Support TestEnv.activate to activate old package management
Hi there, I just run into the wish to activate the test environment of my current project - inside Pluto! (aka using Pluto as the IDE).

Pluto already supports a couple of different triggers to activate the old package management, but not yet [`TestEnv.activate`](https://github.com/JuliaTesting/TestEnv.jl)

https://github.com/fonsp/Pluto.jl/blob/be51e4bf8ceaf14bbc5977892a9ac74afefb54c0/src/packages/Packages.jl#L20C1-L35

jonas-schulze:
Closes #2775 (if it had not been closed already)

How do I test this properly? Ideally, I would have wanted to check for the output of the `haskey(...)` cell to be `true`. I was inspired by the "DrWatson cell" test (does that test even need the `🍭`?), but that doesn't actually run the notebook, IIUC.
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/jonas-schulze/Pluto.jl", rev="testenv")
julia> using Pluto
```

### DIFF
--- a/frontend/components/CellInput/pkg_bubble_plugin.js
+++ b/frontend/components/CellInput/pkg_bubble_plugin.js
@@ -18,6 +18,7 @@ export const pkg_disablers = [
     "Pkg.API.develop",
     "Pkg.add",
     "Pkg.API.add",
+    "TestEnv.activate",
     // https://juliadynamics.github.io/DrWatson.jl/dev/project/#DrWatson.quickactivate
     "quickactivate",
     "@quickactivate",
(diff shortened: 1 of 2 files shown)


fonsp:
Thanks! You can just remove the test, we already test for other disablers.

jonas-schulze:
Done.

fonsp (approved):

fonsp:
Thanks! Merging in a bit

PR #3169, opened 2025-04-04
😵‍💫 Make the pluto-editor component embeddable inside a pluto editor

fonsp:
Embedding Pluto inside Pluto was already possible (also without iframe since https://github.com/fonsp/Pluto.jl/pull/1976). But this PR fixes some CSS issues where the two editors influence layout/styling of each other. E.g. the `disable_ui` feature is no longer global, but specific to each `pluto-editor` component.
 
https://github.com/user-attachments/assets/d6f93724-ee12-4771-8064-5a9e899a8b6d

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="embeddable-in-itself")
julia> using Pluto
```

# TODO
- [ ] Disable scroll on load?
- [ ] Auto scroll on output height change – check for first parent that is scrolled, and scroll it more


fonsp:
Fix missing traffic light

PR #3177, opened 2025-04-07
PlutoPkg: include former stdlibs in logic

fonsp:
Include future and former stdlibs in the logic used by PlutoPkg. 

This mostly affects UI ("this is a stdlib") and whether we automatically set a compat entry.

TODO
- [ ] Use GracefulPkg.jl for the stdlib list? Can be a separate PR


fonsp:
Implemented in https://github.com/fonsp/Pluto.jl/pull/3212

PR #3186, opened 2025-04-10
Just ignore the deleted stdlib warnings

fonsp:
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="update_nbpkg_cache!-resilience")
julia> using Pluto
```


fonsp:
Looks like it works 👍

Issue #3062, opened 2024-10-15
"Motivational words for errors" are being displayed for all errors


AshlinHarris:
#3040 mentions that motivational words wil be displayed at a low frequency, but I'm seeing them for 100% of error messages.

Could you please add a toggle switch to disable all motivational words? My team prepares Pluto notebooks for users who are not programmers or Julia users, so if an error occurs, they could be confused by the humor or the link to the Julia Discourse

pankgeorg:
There is always the low tech solution of adding this cell:

```
html"""<style>.dont-panic{ display: none }</style>"""
```

ederag:
Overall, the new error presentation feels much clearer, thanks !

But here is the story of my encounter with "Keep calm, you got this!":
I was calm, as the errors were expected and easy to fix.
But seeing the computer telling me how to behave ("keep calm"),
and saying something I knew already ("you got this")
did not cheer me up (which was the intent I suppose),
but rather broke my concentration
(where is it coming from, why on some errors only, ... ?)

And after some use I still find those "Motivational words for errors" at best distracting (_not_ entertaining :wink:)
(especially as they appear randomly, changing, and prominent).


By the way, I quickly searched for this issue last week, to no avail,
so to help people searching for the message itself,
here is the list of messages (found in [frontend/components/ErrorMessage.js](https://github.com/fonsp/Pluto.jl/blob/d2dfa8cec3c3d0d6e1c35acd1c4ae6b3fce58c5a/frontend/components/ErrorMessage.js#L549-L566)):

    "Don't panic!",
    "Keep calm, you got this!",
    "You got this!",
    "Silly computer!",
    "Silly computer!",
    "beep boop CRASH 🤖",
    "computer bad, you GREAT!",
    "Probably not your fault!",
    "Try asking on Julia Discourse!",
    "uhmmmmmm??!",
    "Maybe time for a break? ☕️",
    "Everything is going to be okay!",
    "Computers are hard!",
    "C'est la vie !",
    "¯\\_(ツ)_/¯",
    "Oh no! 🙀",
    "this suckz 💣",
    "Be patient :)"

Final note, about the "link to Julia Discourse":
[the few lines below](https://github.com/fonsp/Pluto.jl/blob/d2dfa8cec3c3d0d6e1c35acd1c4ae6b3fce58c5a/frontend/components/ErrorMessage.js#L577-L579) explain why suddenly a discourse page appears sometimes.
[...]

PR #3197, opened 2025-04-23
feat: add option to completely disable motivational quotes

pankgeorg:
~`PLUTO_PROFESSIONAL=true julia -e "import Pluto; Pluto.run()"`~

`julia -e "import Pluto; Pluto.run(;dismiss_motivational_quotes=true)"` should do it

Fixes https://github.com/fonsp/Pluto.jl/issues/3062

Folks, sorry for our delayed response on this and thanks for bearing up with us ❤️ 
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="pg/motivational")
julia> using Pluto
```


tamasgal:
Aw, I was about to provide a PR but was still unsure if this should be provided as a "server option".

Many thanks!

PR #3201, opened 2025-04-24
💫 Use AI to automatically fix syntax errors

fonsp:
https://github.com/user-attachments/assets/9abd41e1-0d54-4fc8-aef2-9e74c0799434

It also includes a prompt to explain what will happen (not shown on the video)

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="ai-syntax-error-fix")
julia> using Pluto
```


digital-carver:
It would be great if there was a way to either:
* go back and forth between the pre-AI-fix version of the code and the fixed code, or
* get a preview of the AI suggestion and be able to either accept it or reject it

Right now, it's hard to see what the AI is changing, especially because there's an unpredictable amount of delay between pressing the button and the change happening. Implementing either of the above would give us a way to compare the two versions, and especially for students/learners, be able to learn from it. 

(I can make an issue for this if you think that'll be helpful, let me know if so!)

fonsp:
Hi @digital-carver ,

That's a nice idea! Feel free to repost in a new issue.

What did you think of the feature besides this?

ctessum:
This looks really interesting! One thing, though, is that I use Pluto for a class that I teach, both for in-class exercises and homeworks (with internet connectivity enabled) and also for exams (with internet connectivity disabled). So I would worry with this feature that students would become reliant on the AI during their homeworks and then not be able to fix any bugs during the exams.

For this reason, would it make sense to also have an option where the AI could just give hints about what is wrong, but still require the user to fix the code themselves?

fonsp:
Hi @ctessum ! Thanks for your feedback!

Just FYI, right now this is only for syntax errors (code does not parse as legal Julia, like misspelling `function`), not for any other type of error.

For syntax errors, Julia already gives a couple of hints of what is wrong, right? In the video, the hints you see are from Julia, and with AI disabled you can already see those hints. But your suggestion is to make the hints better using AI, right? (Maybe use more words, less technical?)

---

I also like your general suggestion (and from @digital-carver as well) about the educational opportunity here: to show what was wrong and what got fixed, and to not just fix it magically.

ctessum:
Yep, that would be the idea. I spend quite a bit of time in the class working with the students to interpret their error messages, and I guess AI could do that too.

Or, another possibility could be to do a "suggested changes" type of thing that shows the student the change that should be made (maybe with highlighting or color coding), but doesn't actually make the change. Just something that elicits more of a "hey I can do this" type of response and less of a "wow that was magic" type of response.

But I think this is really cool too!

PR #3202, opened 2025-04-24
feat: support `~/.plutorc` for options

pankgeorg:
This PR greatly simplifies the duplication in `Pluto.Configuration.from_flat_kwargs` and adds a layer of reading options from a file, in a seamless way.
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="pg/pluto-persistent-options")
julia> using Pluto
```

### DIFF
--- a/src/Configuration.jl
+++ b/src/Configuration.jl
@@ -295,9 +295,9 @@ pluto_file_settings() = try
     end
 catch e;
     if isassigned(pwd_ref)
-        @warn "EXception " (e, catch_backtrace())
+        @warn "Exception " (e, catch_backtrace())
         @warn """Invalid ~/.plutorc
-        
+
         Start with this template:
 
         [Pluto.server]
@@ -312,7 +312,7 @@ catch e;
         [Pluto.evaluation]
         run_notebook_on_load = true
         capture_stdout = true
-        
+
         [Pluto.security]
         require_secret_for_open_links = true
         require_secret_for_access = true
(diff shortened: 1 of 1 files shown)


fonsp:
Hey @pankgeorg thanks for looking into this! Unfortunately this is some pretty rigid code...

For the duplication: this was done to fix a performance issue: https://github.com/fonsp/Pluto.jl/pull/2005 If we remove it, we also need to check these benchmarks again. (Might have changed for new Julia versions.) For me it's not worth the time, as LLMs can work with this file really well and we don't need to change it often.

For `~/.plutorc`: the two issues here are:
- We already have a config file format (for PlutoSliderServer) that can configure Pluto: PlutoDeployment.toml, so I want to avoid adding a new one.
- After we implemented Configurations.jl and PlutoDeployment.toml in Pluto, Julia came out with Preferences.jl and LocalPreferences.toml. 

So if we change something to Pluto's config system, it would be moving to Preferences.jl and LocalPreferences.toml (maybe Configurations.jl can help with this) and supporting Julia 1.12 module arguments (https://github.com/Roger-luo/Configurations.jl/issues/100). This would probably be a breaking change, one that will take a lot of time across the Pluto ecosystem, and I'm not sure if it is worth the trade-off.

src/Configuration.jl
```diff
@@ -280,108 +281,66 @@ Collection of all settings that configure a Pluto session.
...
-        math_mode,
-        heap_size_hint,
-        startup_file,
-        history_file,
-        threads,
-    )
-    return Options(; server, security, evaluation, compiler)
+@option struct PlutoOptions
+    Pluto:: Options = Options
+end
```
pankgeorg, line 286: This struct is specifically introduced so we stay compatible with PlutoDeploymentSettings 😅

pankgeorg (commented):
I've tried to keep the `.plutorc` valid as a `PlutoDeploymentSettings.toml`. I haven't checked performance yet. But I think I can match the other PR, with some tweaks!

As for Preferences.jl, while it is julia friendly, it's not too user friendly; people expect to set an `~/.(software-name)rc`. Not open julia and start setting properties with an API 🐝😅

pankgeorg:
Performance wise it's not good enough at this point

main:
```
julia> import Pluto
Precompiling Pluto...
  1 dependency successfully precompiled in 26 seconds. 68 already precompiled.

  0.000244 seconds (82 allocations: 3.250 KiB)
```

this PR: 
```
julia> import Pluto
Precompiling Pluto...
  1 dependency successfully precompiled in 29 seconds. 68 already precompiled.

[ Info: Using options from ~/.plutorc
  1.228102 seconds (2.13 M allocations: 107.029 MiB, 13.95% gc time, 99.74% compilation time)

[ Info: Using options from ~/.plutorc
  0.000946 seconds (664 allocations: 36.703 KiB)
```

I'll report improvements over this soon!

PR #3203, opened 2025-04-25
Dont use inline flag on Julia 1.12 during tests

fonsp:
Workaround for https://github.com/JuliaLang/julia/issues/58229

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="inline-1.12")
julia> using Pluto
```


fonsp:
Precompilation is still not working, I will continue in another PR

PR #3204, opened 2025-04-25
Julia 1.12: Support `Base.@__doc__`

fonsp:
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="Support-Base__doc__-macro")
julia> using Pluto
```


fonsp:
Can't run CI for this because of FuzzyCompletions

PR #3212, opened 2025-04-29
Use past-present-future list of stdlibs

fonsp:
Since the stdlib list keeps changing, we need to have a list of past-present-or-future stdlibs. I added this in GracefulPkg
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="gracefulpkg-stdlibs")
julia> using Pluto
```


fonsp:
Tests will fail because of other 1.12 issues. Relevant tests pass locally

PR #3239, opened 2025-05-20
Performance: don't render loading bar animation when idle

pankgeorg:
@fonsp 🦜

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="pg/co2-saver")
julia> using Pluto
```


fonsp:
Nice! I brought back the old animation behaviour :) by adding a second, slower delayed truth used for removing the element completely.

fonsp:
Sick!! Well found 💛

PR #3243, opened 2025-05-23
Automatic `id` tags for Markdown headers

fonsp:
I was thinking of a couple ways to do this, but the simplest one turned out the best! I tried lowercasing, replacing ` ` with `-`, URI encoding, deaccenting, etc. But just using the header content as ID works great! You can just copy a header text, and use it as ID. And both id and anchor will use the same unicode->URI transformation, sick!

https://github.com/user-attachments/assets/7f57875d-2ca2-421e-bd54-1d93c9a6db56

### DIFF
--- a/src/runner/PlutoRunner/src/PlutoRunner.jl
+++ b/src/runner/PlutoRunner/src/PlutoRunner.jl
@@ -64,6 +64,7 @@ include("./evaluation/deleting globals.jl")
 
 
 include("./display/LaTeX.jl")
+include("./display/HeaderID.jl")
 include("./display/format_output.jl")
 include("./display/IOContext.jl")
 include("./display/syntax error.jl")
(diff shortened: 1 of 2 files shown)


fonsp:
The spaces in the URL are fine (I like the simplicity!) but they can be a problem when sharing URLs online with auto-link detection:

https://plutojl.org/en/docs/javascript-api/#ObservableHQ stdlib

Whereas if you write it with `%20`, it's the same link and it auto-detects:

https://plutojl.org/en/docs/javascript-api/#ObservableHQ%20stdlib

But this will only be a problem depending on how give the URLs to people. Let's add a 🔗 click-to-copy-URL to headers with `%20` for spaces to fix this :)

fonsp:
I ended up replacing spaces with dash in #3264 hihi

Issue #3259, opened 2025-06-07
precompilation fails on Julia v1.12: "FieldError: type Method has no field `deleted_world`"


nsajko:
```julia-repl
julia> using Pluto
Info Given Pluto was explicitly requested, output will be shown live 
ERROR: LoadError: FieldError: type Method has no field `deleted_world`, available fields: `name`, `module`, `file`, `line`, `dispatch_status`, `primary_world`, `sig`, `specializations`, `speckeyset`, `slot_syms`, `external_mt`, `source`, `debuginfo`, `unspecialized`, `generator`, `roots`, `root_blocks`, `nroots_sysimg`, `ccallable`, `invokes`, `recursion_relation`, `nargs`, `called`, `nospecialize`, `nkw`, `isva`, `is_for_opaque_closure`, `nospecializeinfer`, `did_scan_source`, `constprop`, `max_varargs`, `purity`
Stacktrace:
  [1] getproperty(x::Method, f::Symbol)
    @ Base ./Base_compiler.jl:54
  [2] top-level scope
    @ ~/.julia/packages/Pluto/endwR/src/runner/PlutoRunner/src/evaluation/deleting globals.jl:203
  [3] include(mapexpr::Function, mod::Module, _path::String)
    @ Base ./Base.jl:304
  [4] top-level scope
    @ ~/.julia/packages/Pluto/endwR/src/runner/PlutoRunner/src/PlutoRunner.jl:63
  [5] include(mapexpr::Function, mod::Module, _path::String)
    @ Base ./Base.jl:304
  [6] top-level scope
    @ ~/.julia/packages/Pluto/endwR/src/Pluto.jl:52
  [7] include(mod::Module, _path::String)
    @ Base ./Base.jl:303
  [8] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
    @ Base ./loading.jl:2991
  [9] top-level scope
    @ stdin:5
 [10] eval(m::Module, e::Any)
    @ Core ./boot.jl:489
 [11] include_string(mapexpr::typeof(identity), mod::Module, code::String, filename::String)
    @ Base ./loading.jl:2837
 [12] include_string
    @ ./loading.jl:2847 [inlined]
 [13] exec_options(opts::Base.JLOptions)
    @ Base ./client.jl:319
[...]

PR #3260, opened 2025-06-08
Leave deactivated cells in old_runnable

### LINKED ISSUE #3089: Deleting disabled cell does not update indirectly disabled cells
One cell defines `x` and is disabled, and another cell contains `x + 1` (indirectly disabled). When I remove the first cell, the second cell remains indirectly disabled.

ederag:
Fixes #3089

### DIFF
--- a/src/evaluation/Run.jl
+++ b/src/evaluation/Run.jl
@@ -82,7 +82,7 @@ function run_reactive_core!(
     # by setting the reactive node and expression caches of deleted cells to "empty", we are essentially pretending that those cells still exist, but now have empty code. this makes our algorithm simpler.
     new_topology = PlutoDependencyExplorer.exclude_roots(new_topology, removed_cells)
 
-    # find (indirectly) deactivated cells and update their status
+    # find (directly and indirectly) deactivated cells and update their status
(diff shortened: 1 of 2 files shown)


ederag:
The failures are similar to #3259, so probably unrelated to this PR.

_Side note_:
This got me puzzled: `indirectly_deactivated` reads more restrictive than `deactivated`,
but IIUC the meaning in this function is really "all disabled cells, directly or not".

So would you agree to rename `indirectly_deactivated` to `deactivated` ?
(and maybe - didn't check yet - replace `deactivated` with `directly_deactivated` when relevant ?)
If agreed, then I'll file another PR.

ederag:
Thanks, then I'll add test for functions here first.

ederag:
Thanks for  68aa5639f029a0cb46fcc9d12bdda36617d5d098;
I took `workspace_use_distributed` for `workspace_use_distributed_stdlib` (which defaults to [`false`](https://plutojl.org/en/docs/configuration/)).

The new tests would indeed error if the `Run.jl` code was reverted for instance (checked).

I'll wait for this PR to be merged before filing the one with the renames (hopefully this week).

fonsp:
Very nice! I just want to tweak the test a bit, to make the `update_run` calls exactly as the frontend would make them.

fonsp:
Awesome, thanks so much!!

PR #3268, opened 2025-06-16
🙈 Button to "forget" stale recent notebooks from main menu

### LINKED ISSUE #877: Remove stale 'recent' notebooks
## definitions

`stale` means notebooks that no longer exist in user's filesystem (deleted or moved)
`recent` are the notebooks that appear in the 'Welcome' page. This list is populated when you open a notebook.

## useful information / pointers

The `recent` notebooks list is saved in the [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) object. The list is updated on the `edit` page ([here](https://github.com/fonsp/Pluto.jl/blob/master/frontend/components/Editor.js#L911)) and is read on the `Welcome` page, [here](https://github.com/fonsp/Pluto.jl/blob/master/frontend/components/Welcome.js#L329)

## suggested implementation outline

The browser cannot check if *local files* exist, so the test should take place in the backend.

Communication with the Pluto backend happens through Websockets from the `/edit` page.
Also, Pluto has some tricks to open a notebook remotely (when you click the '>' button on the recent notebook list) and understands that the file doesn't exist if you [try to start the notebook](https://github.com/fonsp/Pluto.jl/blob/master/frontend/components/Welcome.js#L264) it and that [fails](https://github.com/fonsp/Pluto.jl/blob/master/frontend/components/Welcome.js#L264).

- A basic implementation could _remove_ the local storage entry when such a call (e.g. when it fails to open the notebook)
- A more advanced implementation would use Editor's "[this.client.send](https://github.com/fonsp/Pluto.jl/blob/master/frontend/components/Editor.js#L297)" to send a _new_ request, through the websockets mechanism. The request would be handled like [this one](https://github.com/fonsp/Pluto.jl/blob/master/src/webserver/Dynamic.jl#L26) but instead of the `:connect` symbol it would have your own (`:clean_recent`?). Then a function on the backend would receive the list you'll have sent, check if the [files exists](https://docs.julialang.org/en/v1/base/file/#Base.Filesystem.isfile), and return the new list. Then the frontend (which patiently `await`s all this time) will update the localStorage accordinly. This can run once each time the `/edit` is opened.

If you're interested in doing that and have more questions, feel free to ask here or in [zulip topic on this](https://julialang.zulipchat.com/#narrow/stream/243342-pluto.2Ejl/topic/Remove.20links.20on.20landing.20page)

fonsp:
Fix #877

https://github.com/user-attachments/assets/f5e25e05-7eec-4311-8f5d-4a478050f1d5

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="recent-notebooks-forget-button")
julia> using Pluto
```

### DIFF
--- a/frontend/components/welcome/Recent.js
+++ b/frontend/components/welcome/Recent.js
@@ -175,6 +175,14 @@ export const Recent = ({ client, connected, remote_notebooks, CustomRecent, on_s
         document.body.classList.toggle("nosessions", !(combined_notebooks == null || combined_notebooks.length > 0))
     }, [combined_notebooks])
 
+    const on_clear_click = (/** @type {CombinedNotebook} */ nb) => {
+        // Remove from localStorage
+        remove_notebook_from_storage(nb.path)
+
+        // Remove from component state
+        set_combined_notebooks((prevstate) => prevstate?.filter((n) => n.path !== nb.path) ?? null)
+    }
+
     /// RENDER
 
     const all_paths = combined_notebooks?.map((nb) => nb.path)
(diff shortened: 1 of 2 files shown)


phlaster:
Thanks!

PR #3277, opened 2025-06-23
Update min HTTP to 1.10.7

pankgeorg:
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="pankgeorg-patch-1")
julia> using Pluto
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,7 @@ name = "Pluto"
 uuid = "c3e4b0f8-55cb-11ea-2926-15256bba5781"
 license = "MIT"
 authors = ["Fons van der Plas <fons@plutojl.org>"]
-version = "0.20.13"
+version = "0.20.14"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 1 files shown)


fonsp:
This PR was about https://github.com/JuliaWeb/HTTP.jl/security/advisories/GHSA-4g68-4pxg-mw93

We looked at our codebase carefully (because we have cookie-based authentication, and access to the Pluto server would allow remote code execution). We are confident that the CVE does not affect Pluto. 

The issue is with URIs.jl URI parsing (`String -> URI`). IIUC, the exploit relies on data from such a `URI` object entering:
- ⁠a response header
	- but we don't have this
- ⁠an html response
	- this is possible [here](https://github.com/fonsp/Pluto.jl/blob/cf3ebefabbc83be2ca93d958eb2a3a609cee5a31/src/webserver/Router.jl#L58), but it's html escaped (and you need the secret)


I also believe that our [secret-based Auth](https://github.com/fonsp/Pluto.jl/blob/cf3ebefabbc83be2ca93d958eb2a3a609cee5a31/src/webserver/Authentication.jl#L8) cannot be exploited using this CVE: the header required for privileged access is `Secret: --secret-here--`, and not `Access: true` or something similar. And modifying response headers does not give privileged access.

---

Our conclusion is that it's good to merge this PR (for good practice). But we don't need to "escalate", which would mean to publish a warning on Julia forums, and to publish a Pluto release after this PR, and using our "Recommended release" system to issue a warning to all users on an old version to upgrade.

PR #3279, opened 2025-06-24
🧾 Options to store notebook in visual order or executable order

### LINKED ISSUE #3179: Remove `Cell order` section in `.jl` file
I currently need to merge a lot of branches containing a Pluto notebook and it's a huge pain wasting several hours.

One of the advertising features is "Pluto notebook files are designed to work well with version management.".

I would not subscribe to this statement. Yes, they do work with version management, but not particularly well at the moment.

There are two reasons:
1. The order of the cells seems to change a lot over time. When merging two versions of a notebook which are 6 months apart, more a less the whole file shows up as one large diff although one version only has a handful of cells added/changed and the rest is some seemingly random resorting done by Pluto over time. But it is nearly impossible with the usual tools to identify this "real" diff.
2. Whenever you change something during merge, you need to be really cautious to always keep the cell itself in sync with the reference to the cell at the bottom of the document in the `Cell order` section.

It would be so much simpler if the cells were stored in the `.jl` file in the "real" order of how they appear in the document. Then, the `cell order` section could be completely removed if my understanding is correct.
The dependency between the cells can be calculated when reevaluating the document.
Whether a cell is collapsed or expanded can be stored in the cell itself.

fonsp:
_Work in progress_

This PR allows you to configure how notebook files are stored. Files can be stored:
1. As they are now – in executable order, with a section `Cell order` at the bottom.
2. In a **new format – in visual order, without the `Cell order` section**

This new format is:
- more useful with git (smaller diffs)
- more useful with AI codegen (easier for the AI to write new cells in the correct place), together with `auto_reload_from_file=true`
- more useful when editing Pluto notebook files outside of Pluto by hand (although this is not a use case I want to spend to much time on)

# Backward compatibility
The new format can already be opened in old Pluto versions, but unfortunately the cells will appear in random order... (because they are stored in an unordered `Dict`).

# Configuration

This adds two new configurations:

`store_in_executable_order_new::Bool = true` For newly created files, should the notebook file store cells in executable order, so that the notebook file can run as a standalone Julia file? If false, the visual order will be used as the file order, and the Cell order section is ommited.

`store_in_executable_order_existing::Union{Nothing,Bool} = nothing` After opening an existing notebook, should the notebook file store cells in executable order, so that the notebook file can run as a standalone Julia file? If false, the visual order will be used as the file order, and the Cell order section is ommited. If nothing, the notebook will be saved in the format it was opened in.

# Use
The important ideas with this change are:
- You don't need to know about or understand this system to open, use and share Pluto files. If you never look at the literal contents of the `.jl` Pluto files, you will not see a difference.
- By default, existing notebook files should not change file format.

Fix #3179 

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="no-executable-order")
julia> using Pluto
```

### DIFF
--- a/src/notebook/saving and loading.jl
+++ b/src/notebook/saving and loading.jl
@@ -208,7 +208,17 @@ function _read_notebook_collected_cells!(@nospecialize(io::IO))
         if cell_id_str == "Cell order:"
             break
         else
-            cell_id = UUID(cell_id_str)
+            cell_id_parsed = tryparse(UUID, cell_id_str)
+            cell_id = if cell_id_parsed isa UUID
+                if haskey(collected_cells, cell_id_parsed)
+                    @warn "Cell ID appears multiple times in the file. Generating a new one."
+                    uuid1()
+                else
+                    cell_id_parsed
+                end
+            else
+                uuid1()
+            end
 
             metadata_toml_lines = String[]
             initial_code_line = ""


fonsp:
🙈 But now it's not storing the code visible/hidden status... Need to rethink this

PatrickHaecker:
> 🙈 But now it's not storing the code visible/hidden status... Need to rethink this

I think the code visible/hidden status needs to be shifted to the cell itself with the new format, i.e. in the `# ╔═╡` of the cell itself instead of in the `# ╠═` at the bottom. I am not sure whether that kill compatibility with opening it from older Pluto versions, though.

fonsp:
Yes! Actually I want to store it as cell metadata (just like "disabled")

fonsp:
Updated parser factored out into #3285

PR #3285, opened 2025-07-01
More stable notebook file parsing when cell order is missing

fonsp:
Split out from #3279, to make Pluto more future-proof for when we release #3179.

This PR makes Pluto better at opening notebook files that are missing the `Cell Order` section at the bottom of the file.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="more-stable-notebook-cell-order-parsing")
julia> using Pluto
```

### DIFF
--- a/sample/Basic.jl
+++ b/sample/Basic.jl
@@ -1,5 +1,5 @@
 ### A Pluto.jl notebook ###
-# v0.20.14
+# v0.14.0
 
 using Markdown
 using InteractiveUtils
@@ -21,14 +21,14 @@ converges to
 ```
 """
 
-# ╔═╡ b2d792c2-7f73-11ea-0c65-a5042701e9f3
-sqrt(sum(seq) * 6.0)
-
 # ╔═╡ b2d79330-7f73-11ea-0d1c-a9aad1efaae1
 n = 1:100000
 
 # ╔═╡ b2d79376-7f73-11ea-2dce-cb9c449eece6
 seq = n .^ -2
+
+# ╔═╡ b2d792c2-7f73-11ea-0c65-a5042701e9f3
+sqrt(sum(seq) * 6.0)
 
 # ╔═╡ Cell order:
 # ╟─b2d786ec-7f73-11ea-1a0c-f38d7b6bbc1e
--- a/src/notebook/saving and loading.jl
+++ b/src/notebook/saving and loading.jl
@@ -34,8 +34,6 @@ function save_notebook(io::IO, notebook::Notebook)
     println(io, _notebook_header)
     println(io, "# ", PLUTO_VERSION_STR)
     
-    exe_order = notebook.store_in_executable_order
-
     # Notebook metadata
     let nb_metadata_toml = strip(sprint(TOML.print, get_metadata_no_default(notebook)))
         if !isempty(nb_metadata_toml)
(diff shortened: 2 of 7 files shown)


sample/Basic.jl
KeithWM, line 1: I was going to write something else, but now I'm confused: does the new file even work when parsed as a regular .jl file? Shouldn't the definition of `seq` happen _before_ computing the sum?

src/notebook/saving and loading.jl
```diff
@@ -122,15 +122,16 @@ function save_notebook(io::IO, notebook::Notebook)
...
-
-    println(io, _cell_id_delimiter, "Cell order:")
-    for c in notebook.cells
-        delim = c.code_folded ? _order_delimiter_folded : _order_delimiter
-        println(io, delim, string(c.cell_id))
-    end
-    if write_package
-        println(io, _order_delimiter_folded, string(_ptoml_cell_id))
-        println(io, _order_delimiter_folded, string(_mtoml_cell_id))
+    begin
```
KeithWM, line 125: What is the point of this begin-end block?

src/notebook/saving and loading.jl
```diff
@@ -199,12 +200,23 @@ end
...
 function _read_notebook_collected_cells!(@nospecialize(io::IO))
     collected_cells = Dict{UUID,Cell}()
+    collected_cells_order = UUID[]
     while !eof(io)
         cell_id_str = String(readline(io))
         if cell_id_str == "Cell order:"
             break
         else
-            cell_id = UUID(cell_id_str)
+            cell_id_parsed = tryparse(UUID, cell_id_str)
```
KeithWM, line 209: I would recommend replacing lines 209:219 by a single function call `cell_id = unique_parse(collected_cells, cell_id_str)` (with a better name).

Maybe it's because I'm unfamiliar with the syntax, but the nested if else on the rhs of an assignment is a bit scary to me. As an added bonus, you lose the `cell_id_parsed` variable from the outer scope _and_ shorten the function `_read_notebook_collected_cells!`.

src/notebook/saving and loading.jl
```diff
@@ -300,13 +313,13 @@ function _read_notebook_nbpkg_ctx(cell_order::Vector{UUID}, collected_cells::Dic
     return nbpkg_ctx
 end
 
-function _read_notebook_appeared_order!(cell_order::Vector{UUID}, collected_cells::Dict{Base.UUID, Cell})
+function _notebook_appeared_order(cell_order::Vector{UUID}, collected_cells_order::Vector{UUID})
     setdiff!(
         union!(
             # don't include cells that only appear in the order, but no code was given
-            intersect!(cell_order, keys(collected_cells)),
+            intersect(cell_order, collected_cells_order),
```
KeithWM, line 320: If you do `union!` (bang) on the output of `intersect` (no bang) do you actually modify a variable in scope?

src/notebook/saving and loading.jl
```diff
@@ -316,15 +329,16 @@ end
...
     cell_order = _read_notebook_cell_order!(io, collected_cells)
     nbpkg_ctx = skip_nbpkg ? nothing : _read_notebook_nbpkg_ctx(cell_order, collected_cells)
-    appeared_order = _read_notebook_appeared_order!(cell_order, collected_cells)
 
+    appeared_order = _notebook_appeared_order(cell_order, collected_cells_order)
     appeared_cells_dict = filter(collected_cells) do (k, v)
         k ∈ appeared_order
     end
     topology = _initial_topology(appeared_cells_dict, appeared_order)
+    was_stored_in_executable_order = !isempty(cell_order)
```
KeithWM, line 341: This variable appears to be unused.

KeithWM (changes requested):
I did my best to do some review because I do really appreciate the effort, but I'm not sure how valuable my comments are.

sample/Basic.jl
fonsp: I'll revert changes to this file to avoid confusion, but nothing is changing to file writes. Not sure when I changed this file by accident

fonsp:
Thanks!

src/notebook/saving and loading.jl
KeithWM: I tested this, and of course it makes sense and saves you an allocation :-).

KeithWM (approved):

PR #3303, opened 2025-07-31
feat(api): javascript api for custom ui building

pankgeorg:
These are a bunch of utilities extracted from Editor.jl (thank you claude) that allow for easy embedding of Pluto in other web applications.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="pg/frontend-api")
julia> using Pluto
```


fonsp:
We could factor out the inline definition of `real_action` in Editor.js into a separate file. We can structure it as a function:

```js
create_actions_for(client, {get_state, set_state}) -> PlutoActions
```

and Editor.js would call it with:

```js
create_actions_for(client, {get_state: () => this.state.notebook(), set_state: this.setStatePromise})
```

fonsp:
```juila
import MinimalPlutoDisplay as MPD


function cool_plot(x)
const diplay = MPD.create_display(
let
import PlutoPlotly
PlutoPlotly.plot($x)
end)

end


```


And it would:
- Not start a new process
- Use the Pluto Julia-JS communication
- Use Pluto's display (same script execution, styling maybe, etc)

fonsp:
I still believe we can turn this into a much smaller script.

Can you try starting from scratch by adding `ui.ts` and adding this line to package json

```json
        "build-rainbow": "cd ../frontend && parcel build --no-source-maps --public-url . --dist-dir ../frontend-dist-rainbow --config ../frontend-bundler/.parcelrc-offline ui.ts",
```

That already creates a standalone JS lib, but it needs some tweaking with the `target` so Node can import it.

PR #3307, opened 2025-08-04
Refactor to allow translating UI

fonsp:
This PR lets us translate Pluto into different natural languages! It is now really easy for people to contribute translations for the Pluto interface, so text like "Add cell", "Restart process" and "Multiple definitions for x" can be displayed in another language!

 Here is an example of the main menu, in the [Dutch language](https://nl.wikipedia.org/wiki/Nederlands_in_Nederland) (I started on a translation file):

Scherm­afbeelding 2025-08-04 om 09 49 42

At the bottom of the notebook, there is a new menu where you can select the language. This dropdown also shows the translation percentage, and a link with instructions to contribute:

image

In the future (when there are good translations), the language will be auto-detected, based on the system preference.

# Implementation

This uses i18next, and json translation files. I wrote some custom functions to make it work in preact :)

# Contributing

Everyone is very welcome to contribute to making Pluto available in more languages! You do not need to write a complete translation (about 300 words/phrases), you can also translate a couple of phrases, and your contribution will be much appreciated!

Check out the new documentation here: https://github.com/fonsp/Pluto.jl/tree/main/frontend/lang#readme

# TODO
- [ ] update https://github.com/JuliaPluto/pluto-developer-instructions


fonsp:
For the translation process:
- We could ask translators to edit the JSON files directly. Then you can also test more easily: you can `Pkg.dev` Pluto, edit the json files and refresh to see what it looks like. We can write JS/Julia scripts to update the JSON files against each other.
- Crowdin looks kind of nice, but it does not attribute authors in git commits: https://github.com/crowdin/github-action/issues/35
- Weblate might be better then. It's not clear from their website if it's free or not for FOSS projects.

PR #3325, opened 2025-08-17
🎈🇩🇪 German language support 

kellertuer:
In german there is a formal “Sie”-level of conversation and an informal/personal “Du”-Level. The Du is much more common these days even in some stores.
So for the software I also went for that. I used an AI (Claude to be precise) for the rough translation, but no worries, German is my mother tongue, so that was just to reduce a bit of typing. I then carenfully checked all strings again that they chose the right tone/form.
If one wants to be super exact the is a `de/de` (German/German) form, there might be subtle differences in de/at (Austrian) and de/ch (Swiss) German, but they should also just be subtle.

## A few comments and questions

* I kept order and line numbers the same for easier maintenance
* i dadpted the addition for languages from the Polish PR - should we order languages alphabetically in those places? Makes them easier to find when we have a few more.
* One thing I did not translate is “Notebook”. Formally that would be “Notizbuch” but for Pluto or  Jupyter Notebooks I have never heard anyone say Notizbuch; everyone says Notebook.
* The same for “Code” where also the AI kept that. The original word would be “Quelltext” but that is a bit like the 80s called...
* For some terminology like Stack-Trace there is no reasonable tranmslation and everyone would use Stack-Trace or souund from the 50s.
* both `t_stack_frame_this_cell` and `t_stack_frame_other_cell` are “This cell” in the original, is that on purpose? I would have called the second differently? 
* German has a few different capitalization rules in full sentences, I tried to adapt that carefully
* Though “⚠️ Danger” means “Gefahr” in the context of loading unsafe notebooks and such, I went for “⚠️ Achtung!”
* I feel AI and LLM are a bit mixed, but I kept their original usages.
* I am a bit unsure about `instantiate1` to 3, what do they mean? (lines 311-313)
* 314/315 run contains the verb evaluate evaluate contains Running - is that on purpose or to confuse the German (me) ;) ?
* For the lines 318-322 I am unsure wehther that reports ongoing actions (e.g. displayed _while_ connecting to backend) or when that is finished (_after having connected_) - in german those would translate differently

Finally I was a bit unsure how to test this - locally in pkg dev mode?

PR #3329, opened 2025-08-18
🎈🇳🇴 Norwegian (Bokmål) language support

kellertuer:
Here is a first approach to a Norwegian translation.

First note that Norwegian has two official languages: [Bokmål](https://no.wikipedia.org/wiki/Bokmål) and [Nynorsk](https://no.wikipedia.org/wiki/Nynorsk).
Nynorsk is a minority, but should be respected and supported. Sadly, I only learned bokmål until now; still, I will prepare this PR already with Nynorsk in mind.

## Comments and Notes

* I again (similar to #3325 ) called the “notatbok” just the English “notebook”
* this is a draft for now and just an AI based start I till have to carefully work though

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/kellertuer/Pluto.jl", rev="kellertuer/norsk")
julia> using Pluto
```

### DIFF
--- a/frontend/lang/norsk_bokmål.json
+++ b/frontend/lang/norsk_bokmål.json
@@ -1,6 +1,6 @@
 {
 
-    "t_language_name": "Norsk (bokmål)",
+    "t_language_name": "Norsk bokmål",
 
     "t_my_work": "Mitt arbeid",
     "t_newnotebook": "Opprett en <strong>ny notebook</strong>",


fonsp:
Super exciting! Looking forward :)

fonsp:
Use this for `lang_imports.js`

```js
// Separate file because Prettier cannot handle it.

// These imports require quite modern browsers (https://caniuse.com/mdn-javascript_statements_import_import_attributes) but it will be bundled so it doesn't matter for Pluto releases.
// @ts-ignore
export { default as deutsch } from "../lang/deutsch.json" with { type: "json" }
// @ts-ignore
export { default as ellinika } from "../lang/ellinika.json" with { type: "json" }
// @ts-ignore
export { default as english } from "../lang/english.json" with { type: "json" }
// @ts-ignore
export { default as nederlands } from "../lang/nederlands.json" with { type: "json" }
// @ts-ignore
export { default as norsk_bokmål } from "../lang/norsk_bokmål.json" with { type: "json" }

```

kellertuer:
Ah yes, I forgot to save the file before writing the commit ;)

kellertuer:
Thanks for checking the translation @elorannug !
This should be good to go. I hope I merged main correctly.

fonsp:
Awesome!! Thank you!

I fixed the import in my last commit so you can try it in Pluto now. Did you already try to use Pluto with the new language? It might be nice to use it a bit to see how the language works in action :)

kellertuer:
Looks good to me, maybe besides my perfectionism that it states Norse is only 99% (and deutsch even only 98%). But both are maybe also something for a new PR checking what was added since both PRs were started.

fonsp:
Yes feel free to make a new PR! You can use the update script `julia update_languages.jl ` to check for this, it will automatically add any missing keys to all language files.

I also added a change recently where anything above 98% shows as 100% to make it feel nicer :)

fonsp:
I will make a release next week! Currently busy with our class 😅

kellertuer:
Sure no problem. Maybe I find the time to check on both languages for the missing terms already this week.

PR #3334, opened 2025-08-21
German lang: use neutral gendered forms

fonsp:
Hey @kellertuer!

What do you think of this change? I want to use inclusive language :) This topic came up in the French localisation.

I wrote this myself using my mediocre german skills. I saw that there are also neutral forms without `:` like "Entwickelnde" but I like to show the inclusive language explicitly, since it fits Pluto's mission.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="german-neutral-gendered")
julia> using Pluto
```

### DIFF
--- a/frontend/lang/deutsch.json
+++ b/frontend/lang/deutsch.json
@@ -89,7 +89,7 @@
         "Hab Geduld :)"
     ],
     "t_package_could_not_load": "<p>Das Paket <strong>{{package}}.jl</strong> konnte nicht geladen werden, da es nicht initialisiert werden konnte.</p><p>Das ist nicht schön! Was du versuchen könntest:</p>",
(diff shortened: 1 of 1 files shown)


kellertuer:
Nice!

I saw Guillaume's note and already thought: Oh, I did not think of the inclusiveness in my translation. 
So, great that you directly checked that here as well.
I am a bit unsure wether `Autor*in` would be more common.
Similarly, both would still be a bit binary, but I do not see how to phrase that more inclusive
(For Student*in there is the more inclusive Studierende, but for author and developer this is nonexistent as far as I am aware)

So - let's keep your proposal and when I see a better phrasing, I do a new PR :)

PS: For the Norwegian, I found a native speaker who will take a look soon as well!

fonsp:
Awesome, thanks!

fonsp:
@merlijnkersten pointed out that `Autor:in` get pronounced correctly by screen reader software, but `Autor*in` not always.

PR #3339, opened 2025-08-27
Fix methods tables on Julia 1.12

### LINKED ISSUE #3314: Julia 1.12: MethodTable removed
Pluto is not working on Julia 1.12 because the MethodTable was removed? Or replaced by a global one... not sure.

This code is not working:

https://github.com/fonsp/Pluto.jl/blob/bcd248458bf88337cef2a4575e1f79ccacf6c59d/src/runner/PlutoRunner/src/evaluation/deleting%20globals.jl#L139

This might be useful, this was updated to work in Julia 1.12: https://github.com/nickrobinson251/Tricks.jl/blob/11a9e761cae50b1e0b31a5ef6d408f133cc46f72/src/Tricks.jl#L101-L107

But maybe we don't need this, since we only use it (in Julia 1.12) to visit all methods and there might be different API for that. Maybe related is https://github.com/JuliaLang/julia/pull/59158#issuecomment-3137646357

fonsp:
Fix #3314 

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="methods-tables-1.12")
julia> using Pluto
```

### DIFF
--- a/src/runner/PlutoRunner/src/evaluation/deleting globals.jl
+++ b/src/runner/PlutoRunner/src/evaluation/deleting globals.jl
@@ -136,16 +136,23 @@ Return whether the function has any methods left after deletion.
 function delete_toplevel_methods(f::Function, cell_id::UUID)::Bool
     # we can delete methods of functions!
     # instead of deleting all methods, we only delete methods that were defined in this notebook. This is necessary when the notebook code extends a function from remote code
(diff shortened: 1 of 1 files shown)


fonsp:
I would love a review of this PR, I'm not sure if I did this correctly.

PR #3341, opened 2025-09-01
Add Japanese translation

Itou-Kouki:
Japanese translation added.
I haven't translated the parts I don't understand yet, as I translated while using the it.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/Itou-Kouki/Pluto.jl", rev="Add_Japanese_translation")
julia> using Pluto
```


fonsp:
Awesome, thank you! Looks really good!

One request is that when you have not translated something yet, to leave it as the empty string `""`. Then Pluto will automatically use the English fallback text.

One advantage is that this will show the correct progress percentage in the language picker:

Scherm­afbeelding 2025-09-01 om 09 29 56

I made https://gist.github.com/fonsp/737236148b16b3eccb25d619d8cb9325 which is your current japanese.json file, but with all english strings replaced by the empty string `""`. (I did this by removing all lines with english strings from the file, and then running `julia update_languages.jl`.)

fonsp:
> I haven't translated the parts I don't understand yet, as I translated while using the it.

This is no problem at all! I'm happy that you are careful to understand each part before writing the Japanese text.

A partial translation (like the current PR) is very welcome! So you can let me know: we can merge this right now, and people can start using the partial localization. (Then you just start a new PR every time that you want to contribute something.) Or you can continue working on this PR.

Itou-Kouki:
Thank you good advice and reply for me.
Following your advice, removed all lines with english strings from the file.
I want to continue using it while making improvements wherever possible.

Itou-Kouki:
I have removed the full-width spaces that were thought to be causing the problem and add translations.

I Retryed test, can pass it.
~~But, the pull request can't pass test.~~
~~I serch about "exit code 139", the error means something Segmentation fault of MacOS.~~
~~I development on Windows, how can I fix the error?~~

fonsp:
Don't worry about the test! It's not related to this PR

fonsp:
I made a commit on `main` that should improve the stability of the failing check. But for this PR it's not important.

Itou-Kouki:
I understood the test not relared this pull request.
Is this a good place to wrap things up?

fonsp:
It looks really nice, thank you so much! I believe we have many users in Japan, they will really enjoy this!

If you ever want to continue, feel free to make another Pull Request, even if it is just for a single word.

fonsp:
Hey! Thanks again for your contribution! The latest Pluto releases came with some new features, would you be interested in updating the localization to add localization for these? This includes the new Settings menu, a confirm-before-running feature, and more. Let me know if you need help!

PR #3361, opened 2025-09-13
Added upstream field for localization

s-zymon:
There was no way to set translation for text interpolated by `{{upstreamLink}}`, i.e., "the upstream one".

The name of the field for this text is set as "t_cell_upstream". Not sure if it fully fits the convention.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/s-zymon/Pluto.jl", rev="missing_upstream_loc")
julia> using Pluto
```


fonsp:
Thank you!

Issue #2950, opened 2024-07-11
Support `cpu-target` in `CompilerOptions`


sjkelly:
This can be useful for ensuring repeatable results on a heterogeneous cluster. It appears we need to add a field here:

https://github.com/fonsp/Pluto.jl/blob/85f087ef5ea678f93c5338d2e2da149163e64007/src/Configuration.jl#L219-L268

fonsp:
Feel free to make a PR! Should not be too complex :)

PR #3362, opened 2025-09-15
feat: pass `cpu_type` to the malt worker process

pankgeorg:
fixes: #2950 

`using Pluto;Pluto.run(; cpu_target="generic")`

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/fonsp/Pluto.jl", rev="pg/2950-cpu-target")
julia> using Pluto
```


fonsp:
This stores it in the config but does it also set the command line flag? Is that automatic?

pankgeorg:
> This stores it in the config but does it also set the command line flag? Is that automatic?

Yes, this happens automagically [here](https://github.com/fonsp/Pluto.jl/blob/22f31f5961cf67258010ca54a69ddb5ae0080def/src/evaluation/WorkspaceManager.jl#L303)

Issue #3401, opened 2025-10-20
Revise does not work with Julia 1.12


j-fu:
When using Revise.jl in a Pluto notebook (with Pluto package manager disabled via `Pkg.activate(@__DIR__)`,  with Julia 1.12 (and current nightly) Revise does not react on changed code. It works however on 1.11.

MWE: https://github.com/j-fu/MWE_PlutoRevise : just start notebook.jl, edit `Subpackage.greet`, re-execute the corresponding cell in the notebook and see if the expected changes happen.

@TimHoly

fonsp:
Our revise code is here:

https://github.com/fonsp/Pluto.jl/blob/6c402c379b4f4492a25f0a6793d521afd5f579a1/src/runner/PlutoRunner/src/evaluation/deleting%20globals.jl#L224-L234

This function gets called before every reactive run (i.e. when you run code)

@timholy is this still correct? Any idea why this is not working on Julia 1.12? Thanks!

fonsp:
It looks like these variables are still defined:

<details>
<summary>Details</summary>


```
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.12.1 (2025-10-17)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org release
|__/                   |

julia> using Revise
 │ Package Revise not found, but a package named Revise is available from a
 │ registry. 
 │ Install package?
 │   (@v1.12) pkg> add Revise 
 └ (y/n/o) [y]: 
   Resolving package versions...
   Installed Revise ─ v3.11.0
    Updating `~/.julia/environments/v1.12/Project.toml`
  [c3e4b0f8] ~ Pluto v0.20.19 `~/Documents/Pluto.jl` ⇒ v0.20.20 `~/Documents/Pluto.jl`
  [295af30f] + Revise v3.11.0
    Updating `~/.julia/environments/v1.12/Manifest.toml`
  [da1fd8a2] + CodeTracking v2.0.1
  [807dbc54] + Compiler v0.1.1
  [aa1ae85d] + JuliaInterpreter v0.10.6
  [6f1432cf] + LoweredCodeUtils v3.4.4
[...]

PR #3410, opened 2025-11-06
call revise_if_possible via invokelatest

j-fu:
Solves #3401

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/j-fu/Pluto.jl", rev="invokelatest4revise")
julia> using Pluto
```


fonsp:
Awesome, thank you!

PR #3417, opened 2025-11-17
🇳🇴🇩🇪 Norwegian and German language updates

kellertuer:
Here is an update of Norwegian and German to increase their percentage.

* I think I spotted one old `Pluto.land` in the English file, I replaced it there and in my tranlsations with the placeholder other strings used, i.e. `{{plutoland}}`
* For the `t_frontmatter_language_placeholder` the English text does include an explanation in brackets. Should that be translated/kept? For now I just put the corresponding language codes in the two updated files here.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/kellertuer/Pluto.jl", rev="kellertuer/german-norwegian-2")
julia> using Pluto
```


fonsp:
Hi @kellertuer, thanks so much!!

By the way, I was recently at JuliaCon where Boshra and I gave a workshop. I noticed that a **lot** of people were using Pluto in different languages, English was not even that common! And I received many comments that people appreciate the German localization (also used in universities), and I also met one happy Norwegian user :)

# `t_plutoland_upload_delete`
Here, the original was actually correct. Normally we would turn `pluto.land` into a link using interpolation, but in this case, the text is used for a `title` attribute (tooltip), which needs to be plaintext.

But if it makes more sense for the localizers, we can also use the `{{plutoland}}` syntax, and we just interpolate `plutoland: "pluto.land"`. In that case, you need to change the place where `t_plutoland_upload_delete` is used, and add interpolation, just like how it is done for `t_plutoland_upload_description` (but without `html`, just a string). I'm happy either way.

# `t_frontmatter_language_placeholder`
About `t_frontmatter_language_placeholder`: Indeed we want to add the BCP-47 explainer. This is used in the [frontmatter editor](https://plutojl.org/en/docs/frontmatter/). The input field is free-form text, but we really want people to write a valid BCP-47 code, so I add an explainer in the placeholder as a quick solution.

Scherm­afbeelding 2025-11-17 om 16 10 45

kellertuer:
Oh that is great to hear! And even a Norwegian user!
And yes I also already saw your talk together with Bashra - a very nice talk as always!

For the two things, sure, I can just revert the English one then and will add the comments. I thought is was meant to be some metadata that is read and used, but thanks for your screenshot I see it is a note in a field, where of course the comment is very useful.

kellertuer:
( the failing CI is the current Cloudflare outage I believe)

From my side this PR should be good to go :)

fonsp:
Amazing, thanks so much!!

fonsp:
Btw, is `t_loading_something` correct?

    "t_loading_something": "Lade {{text}}...",

From dutch it would make more sense to say `{{text}} am laden...` but it's not my language of course 😅

Thanks again!

kellertuer:
Yes that is correct it is a kind of informative imperative and even close to “Loading X...”  – “Lade X ...”.
The longer form would be “Ich lade X ...” (but here shortened to the impersonal form) – and is the common thing to do in German here.

“X am laden...” is grammatically not wrong but sounds super strange in German (even worse it is close to “X am Laden” which would refer to something (X) at (or outside in front of) a supermarket ;)).
“Ich bin am Laden” would mean “I am at the store” (in the sense of in front of). So to avoid the “laden/Laden” (loading vs store) confusion in German – no one would say “am laden” here.

fonsp:
Alright, learned something new, thank you!! ☺️

fonsp:
Hey! Thanks again for your contribution! The latest Pluto releases came with some new features, would you be interested in updating the localization to add localization for these? This includes the new Settings menu, a confirm-before-running feature, and more.

kellertuer:
Sure, I can check both the Norwegian and the German translations.

Might take until the weekend though until I find time, especially getting into updating the files first and then checking the translations.

fonsp:
Alright, thank you! Looking forward :)

kellertuer:
I have started with the German version (cf. https://github.com/kellertuer/Pluto.jl/commit/783bcbc3c7e89fe94afb27b2061cc09fea427fd9); opening a PR currently 500ers (angry unicorn is back again!)

For the Norwegian version I probably need a bit more time.

PR #3422, opened 2025-11-22
🇩🇰 Add Danish language support

gwr-de:
While it may not yet be completely perfect, I have carefully checked this with ChatGPT 5.1 to be "modern, idiomatic Danish." I hope this to be a good start for having Danish enter the set of Pluto-supported languages. _Velbekomme_! 🇩🇰 

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/gwr69/Pluto.jl", rev="lang-dansk")
julia> using Pluto
```


fonsp:
Hey @gwr-de, mange tak! Some comments:

I changed the name from `dansk_da` to `dansk_dk`, based on some online results. This is correct, right?

I added the JS code to enable Danish in the language picker. That means that you can try Pluto with this PR, and see what the localization looks like in action! To do so, you just `Pkg.dev(..path to your clone of Pluto..)` and then run Pluto as usual. It will print a message "Looks like you are developing Pluto" in the terminal where you launched Pluto. Then you can pick Danish in the language picker.

Can you try it out and let me know what you think? Are there any changes (like layout) that you still want to make?

Thanks again!

gwr-de:
> Hey @gwr-de, mange tak!

_Det var så lidt._

> I changed the name from dansk_da to dansk_dk, based on some online results. This is correct, right?

Indeed, that's probably the next best thing to `da_DK.json` as `danish_da` is giving the same information twice...

I will try it out and report back.

EDIT

I would still recommend to use international standardized names for _languages_ (e.g., [IETF BCP 47](https://en.wikipedia.org/wiki/IETF_language_tag)) and (if needed/useful) _countries_ (e.g., [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)) to name files—not the language name in the language, given that some languages bring their own alphabets?

gwr-de:
~Thanks for renaming to `danish_dk.json`.~

Just as a suggestion for consistent naming: One might well argue that with regard to UI translations, there is really only one main variant of German (`de`) or even Dutch (`nl`), while country- or regional variants might become relevant for Portuguese (`pt-PR` and `pt-BR`) or English (`en-US` vs. `en-UK`) or even French (`fr-FR` vs. `fr-CA`).

Danish pretty squarely is like German here imho (e.g., there is Faroese `fo` _next_ to Danish) and there is no need to add the country, e.g., `da.json` could sit right next to `de.json` and `nl.json`.

fonsp:
If you want to make that change for danish then that's fine! 

For dutch this is not true (e.g. standard belgian dutch is different and there are more variations), for german the localizer decided to add the region for similar reasons.

gwr-de:
I did some testing using `Pkg.dev` and from my point of view this is now good to go.

fonsp:
Awesome, thanks so much!!

fonsp:
Hey! Thanks again for your contribution! The latest Pluto releases came with some new features, would you be interested in updating the localization to add localization for these? This includes the new Settings menu, a confirm-before-running feature, and more.

gwr-de:
I'll try to get to it this weekend.

gwr-de:
@fonsp The PR is in your inbox: https://github.com/JuliaPluto/Pluto.jl/pull/3576

PR #3218, opened 2025-04-30
📋 Guided Project.toml editor

fonsp:
Implement #3099, see the issue for more info

https://github.com/user-attachments/assets/eb498f8a-39ce-4a12-b657-386da1e544eb

But without https://github.com/JuliaLang/Pkg.jl/issues/4086 the usefulness is limited

PR #3424, opened 2025-11-26
Add `app` functionality

### LINKED ISSUE #100: What is the best way to support --module arguments in Julia 1.12?
Hey! I would like to add `julia -m Pluto` support in Julia 1.12 (see [Julia PR](https://github.com/JuliaLang/julia/pull/52103/files#diff-067d9845807c2637ff61073925364bc897c5ed88bc09ce51fafef0dc3e6c062bR10-R13)) to Pluto! I thought it would be nice if users could also provide some of Pluto's settings as args, like `julia -m Pluto --port=8000 --auto_reload_from_file=true`.

I know this functionality already exists in Comonicon.jl, what is the best way to go about this?

Thanks Roger!!

RomeoV:
This PR may relate to or close https://github.com/Roger-luo/Configurations.jl/issues/100.

## Summary
This PR implements a simple CLI interface to launch Pluto using Julia 1.11 ["apps"](https://pkgdocs.julialang.org/v1/apps/).

Using this PR allows a user to run 

``` sh
pluto mynotebook.jl
pluto --port=1235 --dont-launch-browser
# etc
```
Currently only a small subset of `Pluto.run(; ...)` flags are supported - in particular those which are documented under `?Pluto.run` (see PlutoApp.jl in this PR.)
For more complex use cases a user can just launch Pluto from a Julia REPL as before.

## Motivation
A minor issue that we keep running into is that students are confused exactly how Pluto relies to a "package" they are developing (typically bundling some research code etc).
They end up installing Pluto into their package (adding it as a dependency) and are confused why their package is not available from inside Pluto.
I think this is because the mental model that Pluto launches new Julia workers is not obvious.

With the app workflow I hope that users have a new mental model that you launch Pluto from "anywhere" as an app and then activate their own project from inside Pluto.

## Choice of CLI Parser library
I have chosen [Comonicon.jl](https://github.com/comonicon/Comonicon.jl). Alternatives include
- ad hoc parser
- [`ArgParse.jl`](https://github.com/carlobaldassi/ArgParse.jl)
- [`SimpleArgParse.jl`](https://github.com/admercs/SimpleArgParse.jl)
- [`Fire.jl`](https://github.com/ylxdzsw/Fire.jl)
- [`TrimableCLIParser.jl`](https://github.com/RomeoV/TrimmableCLIParser.jl)

I have chosen because it is relatively widely used, let's us write a quite concise interface, Pluto already uses [Configurations.jl](https://github.com/Roger-luo/Configurations.jl) from the same ecosystem (i.e., author), and I was the most familiar with it.

Happy to get feedback whether this feature is wanted and reasonable in this current form.

## Run the app code
You can either do something like

``` julia
julia --project -m Pluto.PlutoApp --help
```
or

``` julia
]app dev .
;pluto --help
```

For the prior, you may want to consider commenting out the precompile when playing around with the CLI interface (to accelerate your CLI interface feedback loop) by commenting out `include("./precompile.jl")` in `src/Pluto.jl`.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/RomeoV/Pluto.jl", rev="add-app")
julia> using Pluto
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -82,3 +82,6 @@ TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 
 [targets]
 test = ["DataFrames", "OffsetArrays", "Sockets", "Test", "TimerOutputs", "Memoize"]
+
+[apps]
+pluto = {submodule="PlutoApp"}
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,7 @@ version = "0.20.21"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+Comonicon = "863f3e99-da2a-4334-8734-de3dacbe5542"
 Configurations = "5218b696-f38b-4ac9-8b61-a12ec717816d"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
@@ -40,6 +41,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
 Base64 = "1"
+Comonicon = "1"
 Configurations = "0.15, 0.16, 0.17"
 Dates = "0, 1"
 Downloads = "1"
@@ -72,6 +74,9 @@ URIs = "1.3"
 UUIDs = "1"
 julia = "^1.10"
 
+[apps.pluto]
+submodule = "PlutoApp"
+
 [extras]
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Memoize = "c03570c3-d221-55d1-a50c-7939bbd78826"
@@ -82,6 +87,3 @@ TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 
 [targets]
 test = ["DataFrames", "OffsetArrays", "Sockets", "Test", "TimerOutputs", "Memoize"]
-
-[apps]
-pluto = {submodule="PlutoApp"}
--- a/src/Pluto.jl
+++ b/src/Pluto.jl
@@ -86,7 +86,9 @@ include("./webserver/Dynamic.jl")
 include("./webserver/REPLTools.jl")
 include("./webserver/WebServer.jl")
 
-include("PlutoApp.jl")
+@static if VERSION >= v"1.11"
+    include("PlutoApp.jl")
+end
 
 const reset_notebook_environment = PkgUtils.reset_notebook_environment
 const update_notebook_environment = PkgUtils.update_notebook_environment
--- a/src/Pluto.jl
+++ b/src/Pluto.jl
@@ -86,6 +86,8 @@ include("./webserver/Dynamic.jl")
 include("./webserver/REPLTools.jl")
 include("./webserver/WebServer.jl")
 
+include("PlutoApp.jl")
+
 const reset_notebook_environment = PkgUtils.reset_notebook_environment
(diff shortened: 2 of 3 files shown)


fonsp:
Ahh very cool! I definitely want Pluto to feel more like an "app" and I agree that this will improve the mental.

We are currently working again on https://github.com/JuliaPluto/PlutoDesktop, and the hope is to release this within the next year :) So that should cover it for most beginner users, but we will continue to publish Pluto as a Julia package.

For this PR, I would really like to investigate if it would be possible to automatically generate the complete list of keyword arguments. Since Comunicon and Configurations.jl and from the same awesome developer, I was wondering if this was already achieved for another project. The talk at Juliacon that we mentioned in the issue you linked never happened :( but I'm still curious to hear Roger's opinion about this :)

Would you mind investigating this a bit? Maybe you can find more about the combination in documentation, or you can ask Roger about it.

Thank you!

Btw, we recently launched #3218 which allows students to test their local package inside a Pluto notebook without registering it. I still need to write good documentation for this on the Pluto website (help welcome!), but I am curious if you already knew about this feature and what your experience is.

RomeoV:
Comonicon can directly work with `@option` annotated structs as per https://comonicon.org/stable/conventions/#Working-with-option-types-from-Configurations, for instance
```julia
using Test
using Comonicon
using Configurations

@option struct OptionA
    a::Int
    b::Int
end


"""
# Options

- `-c, --config <path/to/option/or/specific field>`: config.
"""
@main function run(;config::OptionA)
    @test config == OptionA(1, 1)
end
```

so this could lead to a very simple integration. I'm thinking something like
```julia
Comonicon.@main function pluto(
    notebooks::String...;
    server::Configuration.ServerOptions,
    security::Pluto.Configurations.SecurityOptions,
    compiler::Configuration.CompilerOptions,
    evaluation::Configuration.EvaluationOptions
)
    options = Options(; server, security, evaluation, compiler)
    session = ServerSession(; options)
    run(session)
end
```

One issue this currently has is that `@option` fields with type `Union{Nothing, T}` are not supported well at the moment. I think in the end there is a `tryparse(Union{Nothing, T}, input)` call that fails. We could pirate that, or make a PR to Comonicon to try to fix that. Or perhaps we wrap every `Union` in some wrapper type `PlutoUnion{Union{...}}` and then override `tryparse` for that...

RomeoV:
~The `Union{Nothing, T}` problem could probably be solved by using [`Configurations.Maybe`](https://configurations.rogerluo.dev/stable/ref/#Configurations.Maybe) instead, and then telling `MsgPack` about that type.~

Nvm that just aliases `Maybe{T} = Union{Nothing, T}` as per https://github.com/Roger-luo/Configurations.jl/blob/f7da275772610d0a9b58eab0438fdcf6491024cd/src/types.jl#L2

fonsp:
Okay! What do you think we should do next? Maybe we can write an issue for Roger to ask for help?

j-fu:
Comonicon makes it easy to define argument lists from a function call, but I guess it is hard to create programmatically. May be what @fonsp is asking for is easier to implement with ArgParse, because this has a function based API.

I used this in my own version of a pluto script: https://github.com/j-fu/julia-cli-tools/blob/main/pluto .

j-fu:
Ah and from that experience it appeared to be useful to have Julia version checking, so that I don't accidentally start a notebook with the "wrong" julia version. 

Also we could add explicit manifest handling to such an app. This may be could allow to adapt notebooks to newer versions of Julia without loading the notebook. I tried it in my version, but essentially this relied on unpublished API. Here we could make this a part of the functionality.

fonsp:
What should we do next for this PR?

RomeoV:
Closing as I unfortunately don't have time anymore to work on this. Happy for someone to pick this up, though.

PR #3440, opened 2026-01-05
Fix logging with ProgressLogging v1.6

### LINKED ISSUE #3434: ProgressLogging.jl does not display nicely anymore
I started noticing recently that the logs from `ProgressLogging.jl` do not display nicely anymore:

Image

Did not have time yet to test whether this is caused by an update of ProgressLogging or an update of Pluto.

This is with julia 1.12.2 on Linux, Pluto v0.20.21 and ProgressLogging v0.1.6

disberd:
The latest version of ProgressLogging changed the object returned by the logs to not be a subtype of `AbstractString` anymore (and removed the `String` conversion) to reduce invalidations.

This PR approaches the fix by adding a new custom _preprocessing_ step for log messages (allowing customization of how to preprocess logs messages even in the future for other packages) and then adds a custom method for the `ProgressLogging.ProgressString` type inside an `Integration` in PlutoRunner.

This seemed the easiest/cleanest fix and it also allows potential reuse for eventual future problems with logs of specific packages used via Integrations

This fixes #3434 

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/disberd/Pluto.jl", rev="fix_progress_logging")
julia> using Pluto
```

### DIFF
--- a/frontend/components/CellInput.js
+++ b/frontend/components/CellInput.js
@@ -68,8 +68,7 @@ import { t } from "../common/lang.js"
 
 export const ENABLE_CM_MIXED_PARSER = window.localStorage.getItem("ENABLE_CM_MIXED_PARSER") === "true"
 export const ENABLE_CM_SPELLCHECK = window.localStorage.getItem("ENABLE_CM_SPELLCHECK") === "true"
-export const ENABLE_CM_AUTOCOMPLETE_ON_TYPE =
-    (window.localStorage.getItem("ENABLE_CM_AUTOCOMPLETE_ON_TYPE") ?? (/Mac/.test(navigator.platform) ? "true" : "false")) === "true"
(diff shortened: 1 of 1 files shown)


fonsp:
Awesome, thank you! And will this still work with past versions of ProgressLogging? Maybe a fallback?

disberd:
Yes this is also backwards compatible. If you see the change that broke Pluto (despite having checked on Pluto based on the comments. I think the author there just used the default PlutoPkg whch installed the older ProgressLogging version) is this PR https://github.com/JuliaLogging/ProgressLogging.jl/pull/50

Accessing the name directly from the progress object was already the behavior of the `String` conversion method delete in the PR above so it would behave as previously in previous ProgressLogging versions

fonsp:
Perfect, thanks!

PR #3441, opened 2026-01-12
avoid access to internal `MethodList` field removed in 1.12.

disberd:
I was playing with internals and found out that this `delete_computer` method errors due to direct access to an internal field of `MethodList` which was removed in https://github.com/JuliaLang/julia/pull/58131

I am not sure what was the original logic, and why this does not simply delete the single methods of the computer function (which should be unique as it's created everytime with a `gensym` name).

I added a `only` here to simplify catching issues where multiple methods might be returned for a specific computer function.

It is strange that this has not popped up on running Pluto notebooks in the wild. I wonder whether this path is actually not hit anymore in latest pluto versions, but couldn't find any obvious pointer to this in the code so far.

@Pangoraw maybe you know why this was done with `Base.visit` instead of directly using `Base.delete_method`? 

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/disberd/Pluto.jl", rev="update_delete_computer")
julia> using Pluto
```

### DIFF
--- a/src/runner/PlutoRunner/src/evaluation/run_expression.jl
+++ b/src/runner/PlutoRunner/src/evaluation/run_expression.jl
@@ -38,7 +38,12 @@ end
 function delete_computer!(computers::Dict{UUID,Computer}, cell_id::UUID)
     computer = pop!(computers, cell_id)
     UseEffectCleanups.trigger_cleanup(cell_id)
-    Base.delete_method(methods(computer.f) |> only) # Make the computer function uncallable
+    # Make the computer function uncallable
+    if VERSION < v"1.12.0-0"
+        Base.visit(Base.delete_method, methods(computer.f).mt)
+    else
+        foreach(Base.delete_method, methods(computer.f))
+    end
 end
 
 parse_cell_id(filename::Symbol) = filename |> string |> parse_cell_id


Pangoraw:
> It is strange that this has not popped up on running Pluto notebooks in the wild. I wonder whether this path is actually not hit anymore in latest pluto versions, but couldn't find any obvious pointer to this in the code so far.



You can try to uncomment this line to see what is going on but it can be silently failing otherwise.

https://github.com/fonsp/Pluto.jl/blob/2ce96375a1b932e11b9b19df3e17cec115d4d7a9/src/runner/PlutoRunner/src/evaluation/run_expression.jl#L254

fonsp:
Let me know how you'd like to continue @disberd :) Do you want me to review?

disberd:
Hi @fonsp, yeah if you could have a review it would be nice.
For the purpose of fixing the issue directly the proposed update should achieve the same effect and rely less on internals (it should work in 1.10 as well as 1.12).

In general we might check better why this error is not popping up in general (I stumbled upon it as I was manually calling the `delete_computer!` function on 1.12) and see if this whole code path is not really used anymore. But that can be probably left as separate issue.

fonsp:
Okay! Then without reviewing the dead code (let's leave that for another issue), I will just copy what we did in https://github.com/JuliaPluto/Pluto.jl/pull/3339/changes .

We can merge this now if the tests pass on Julia 1.10 - 1.12. Julia 1.13 will fail for unrelated reasons

fonsp:
Can you make the separate issue?

PR #3452, opened 2026-01-16
Restore per-method docstrings in Live Docs on Julia 1.13+

Nithin2113:
Julia 1.13 changed Docs.doc behavior and no longer includes method docstrings.
This restores previous behavior by collecting docs from individual methods.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/Nithin2113/Pluto.jl", rev="fix-julia113-docstrings")
julia> using Pluto
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,8 +1,8 @@
 name = "Pluto"
 uuid = "c3e4b0f8-55cb-11ea-2926-15256bba5781"
 license = "MIT"
-authors = ["Fons van der Plas <fons@plutojl.org>"]
 version = "0.20.21"
+authors = ["Fons van der Plas <fons@plutojl.org>"]
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
@@ -47,7 +47,7 @@ ExpressionExplorer = "0.5, 0.6, 1"
 FileWatching = "1"
 GracefulPkg = "2"
 HTTP = "^1.10.17"
-HypertextLiteral = "0.7, 0.8, 0.9"
+HypertextLiteral = "0.7, 0.8, 0.9, 1"
 InteractiveUtils = "1"
 LRUCache = "1.6.2"
 Logging = "1"


Nithin2113:
CI failure seems unrelated to this PR.
Failing test is with_js_link.js (frontend JS test), while this PR only changes Live Docs backend logic.
Looks like flaky timeout in CI.

Nithin2113:
Hello, I investigated the remaining CI failures and they all appear in test/MacroAnalysis.jl, specifically within the "Doc strings" testset.

My changes are limited to Live Docs handling in src/runner/PlutoRunner/src/ide features/docs.jl, so these failures seem unrelated to the modified code and may be due to differences in how docstrings or method summaries are displayed in newer Julia versions.

I did not modify the MacroAnalysis logic or tests in this PR to avoid mixing unrelated changes. If you think these tests should be updated as part of this fix, I’m happy to work on a follow-up commit or a separate PR — please let me know what you’d prefer.

Thank you for the review!

fonsp:
Hey, thanks for your PR!

Could you write a bit more about what changed in Julia 1.13? What is the new behaviour, and why? Is this new behaviour something we also want to do in Pluto? How does it work in the REPL?

The goal is not necessarily to make our old tests pass, but to give the best experience with the new Julia version.


Can you try to make your PR smaller? It contains many whitespace changes and removes and unrelated TODO.

You can ignore the unrelated tests

Nithin2113:
Hi @fonsp thank you for the feedback and sorry for the noise in this PR.

In Julia 1.13+, the display behavior for functions seems to have changed: instead of embedding docstring/type information in the display output, the REPL now shows the generic function summary (e.g. f (generic function with N methods)), and the docstring is accessed separately via ?f. Pluto appears to now follow this same display path, so the cell output matches the REPL summary rather than including ::Bool, ::Int, etc.

My intention was to align the tests with this new Julia behavior rather than forcing the old output format, but I agree that we should ensure this is also the best user experience in Pluto, not just make old tests pass.

I will clean up this PR by reverting the unrelated whitespace changes and the removed TODO, and keep only the changes relevant to the Julia 1.13 display behavior. I’ll update the PR shortly. Thanks again for the guidance!

Nithin2113:
@fonsp Thanks a lot for taking the time to review this and for the helpful feedback!

I checked again on a clean main branch and I’m not seeing the docstring-related test failures locally anymore. It’s possible that what I initially saw was specific to the CI Julia version, or that things have changed upstream since then.

Given that the tests now pass locally, I agree that changing the expected output just to match older assumptions wouldn’t be the right approach, especially if it affects the user experience in Pluto.

I’m happy to close this PR if it’s no longer needed, but also very happy to dig further if there’s a specific failing case on CI or with Julia 1.13 that you’d like me to look into. Thanks again!

fonsp:
Hey! I'm still not quite sure what the change in Julia 1.13 is that this PR fixes. Could you try to explain the change? Maybe you can show me before and after screenshots? I find it difficult to understand without a concrete example.

Is this https://github.com/JuliaLang/julia/issues/60681 or another issue?

Nithin2113:
Hey @fonsp thanks for the detailed question.

I re-tested on a clean main branch and I’m not currently seeing a consistent before/after difference on Julia 1.13 that I can demonstrate with screenshots, so I can’t confidently point to a specific behavior change this PR is fixing.

Because of that, I don’t think this is clearly tied to julia#60681, and it may not be addressing a real regression. I agree it’s probably best to close this PR unless there’s a concrete failing case on CI or locally that you’d like me to focus on.

Really appreciate you taking the time to review this.

PR #3467, opened 2026-02-04
scopestate: ignore call-site kwarg labels

younes-io:
Fixes #3459.

Problem:
- `f(; kw=1)` made `kw` look like a local variable (and a usage) in scopestate.
- This skewed autocomplete ranking and could mis-link labels in go-to-definition/LiveDocs.

Fix:
- Treat call-site kwarg labels as neither definitions nor usages (still traverse values).
- Keep function-definition kwargs (`function foo(; kw=1)`) as locals.

Tests:
- `cd test/frontend && npm test -- __tests__/scopestate_test.js`

PR #3469, opened 2026-02-09
Do not explore Symbol in scopestate statefield

### LINKED ISSUE #3468: Global variable detection in symbol
Image

Pangoraw:
Fixes #3468.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/JuliaPluto/Pluto.jl", rev="sym-var-def")
julia> using Pluto
```


fonsp:
I added a test using the new setup from #3467 and #3470

PR #3529, opened 2026-05-01
Ask the user before running long-running cells

fonsp:
In this PR, any reactive run that will take a long time to run (based on summing the runtimes of all downstream cells), will ask the user for confirmation before.

You can use `Esc`/`Enter` to quickly dismiss/confirm.

https://github.com/user-attachments/assets/d8ebae5c-4fad-4c77-91a0-67f81b5ea231

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/JuliaPluto/Pluto.jl", rev="confirm-before-long-runtime")
julia> using Pluto
```

# TODO
- [x] Auto dismiss in tests
- [x] Find a good timeout setting
- [x] Dark color scheme


fonsp:
Discussion with @pankgeorg :

> [02-05-2026, 11:52:21] ~Παναγιώτης: And I'd also add something more subtle; a countdown that doesn't take your focus
> [02-05-2026, 11:53:39] fns: yeahh i was wondering…
> 
> countdown is nice but maybe stressful / not accessible?
> 
> but without taking focus is an interesting idea - if it takes 2 mins then a 10 sec delay (like Undo Send in email) doesnt matter
> [02-05-2026, 11:54:47] ~Παναγιώτης: yea Undo Send ui is great

fonsp:
I really like the suggestion to make it like "Undo Send" but I'm a bit intimidated by the implementation. By making the dialog blocking (like it is now) we don't need to worry about supporting other user interactions while the window is open, which makes it much easier.

I think I will merge this in its current state, and we can pick it up again later (or leave it as an issue for newcomers). I can play a bit with the styling to make it less intimidating

fonsp:
I worked on the UI a bit more. This is the current state (in this video I put the threshold at -1 second just for testing, that's why it shows for very small runtimes).


https://github.com/user-attachments/assets/1aff21ed-aa0d-4f46-bd09-ec0cf92818f1

It's now closer to the "Undo send" UI but it still traps focus.

PR #3531, opened 2026-05-06
Detect JULIAHUB_APP_URL for launch URL

### LINKED ISSUE #3530: URL printing for non-local use
When running Pluto inside of a system such as JuliaHub, where you're running a code-server that proxies local ports the default printing on start up of a server obviously doesn't work:

```
┌ Info: 
│ Go to http://localhost:1234/?secret=PqfnSkGq in your browser to start writing ~ have fun!
└
```

The ports are exposed through the scheme below:

```
https://***.juliahub.com/proxy/1234/
```

It would be useful if Pluto could detect such cases and print out

```
┌ Info: 
│ Go to https://***.juliahub.com/proxy/1234/?secret=PqfnSkGq in your browser to start writing ~ have fun!
└
```

to make the experience seamless.

MichaelHatherly:
JuliaHub renamed JH_APP_URL to JULIAHUB_APP_URL on its current job images. Pluto only checked the old name, so the printed launch URL fell back to the bare localhost form even though a proxy URL was available.

Check JULIAHUB_APP_URL first and keep JH_APP_URL as a fallback for older job environments. Add a pretty_address testset covering the default, root_url override, both env var names, precedence, and root_url winning over the env vars.

Fixes #3530

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/MichaelHatherly/Pluto.jl", rev="mh/fix-juliahub-app-url")
julia> using Pluto
```

### DIFF
--- a/src/webserver/WebServer.jl
+++ b/src/webserver/WebServer.jl
@@ -355,7 +355,10 @@ function pretty_address(session::ServerSession, hostIP, port)
     root = if session.options.server.root_url !== nothing
         @assert endswith(session.options.server.root_url, "/")
         replace(session.options.server.root_url, "{PORT}" => string(Int(port)))
+    elseif haskey(ENV, "JULIAHUB_APP_URL")
+        "$(ENV["JULIAHUB_APP_URL"])proxy/$(Int(port))/"
     elseif haskey(ENV, "JH_APP_URL")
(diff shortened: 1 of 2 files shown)


pankgeorg (approved):
I can say with both my hats that this looks ok!

Issue #3449, opened 2026-01-14
Julia 1.13 tests: Macro analysis - docstrings


fonsp:
Test failure:

```

Doc strings: Test Failed at /Users/fons/Documents/Pluto.jl/test/MacroAnalysis.jl:874
  Expression: occursin("::Bool", bool.output.body)
   Evaluated: occursin("::Bool", "f (generic function with 1 method)")
  Stacktrace:
   [1] top-level scope
     @ ~/Documents/Pluto.jl/test/MacroAnalysis.jl:7
   [2] macro expansion
     @ ~/.julia/juliaup/julia-1.13.0-beta1+0.aarch64.apple.darwin14/share/julia/stdlib/v1.13/Test/src/Test.jl:1961 [inlined]
   [3] macro expansion
     @ ~/Documents/Pluto.jl/test/MacroAnalysis.jl:849 [inlined]
   [4] macro expansion
     @ ~/.julia/juliaup/julia-1.13.0-beta1+0.aarch64.apple.darwin14/share/julia/stdlib/v1.13/Test/src/Test.jl:1961 [inlined]
   [5] macro expansion
     @ ~/Documents/Pluto.jl/test/MacroAnalysis.jl:874 [inlined]
   [6] macro expansion
     @ ~/.julia/juliaup/julia-1.13.0-beta1+0.aarch64.apple.darwin14/share/julia/stdlib/v1.13/Test/src/Test.jl:753 [inlined]

Doc strings: Test Failed at /Users/fons/Documents/Pluto.jl/test/MacroAnalysis.jl:876
  Expression: occursin("::Bool", int.output.body)
   Evaluated: occursin("::Bool", "f (generic function with 2 methods)")
  Stacktrace:
   [1] top-level scope
     @ ~/Documents/Pluto.jl/test/MacroAnalysis.jl:7
   [2] macro expansion
     @ ~/.julia/juliaup/julia-1.13.0-beta1+0.aarch64.apple.darwin14/share/julia/stdlib/v1.13/Test/src/Test.jl:1961 [inlined]
   [3] macro expansion
     @ ~/Documents/Pluto.jl/test/MacroAnalysis.jl:849 [inlined]
   [4] macro expansion
     @ ~/.julia/juliaup/julia-1.13.0-beta1+0.aarch64.apple.darwin14/share/julia/stdlib/v1.13/Test/src/Test.jl:1961 [inlined]
   [5] macro expansion
[...]

Issue #3389, opened 2025-10-13
Julia 1.13 support


fonsp:
If you find an issue with Pluto on Julia 1.13, please **open a new issue**.

- https://github.com/JuliaPluto/Malt.jl/issues/101
- https://github.com/JuliaPluto/Malt.jl/issues/93
- https://github.com/JuliaPluto/PlutoDependencyExplorer.jl/issues/18
  - https://github.com/fonsp/Pluto.jl/pull/3447
- https://github.com/fonsp/Pluto.jl/issues/3445
  - https://github.com/fonsp/Pluto.jl/pull/3447
- https://github.com/fonsp/Pluto.jl/pull/3448
- Update version warning to 1.14
- https://github.com/fonsp/Pluto.jl/issues/3449
  - Caused by https://github.com/JuliaLang/julia/issues/60681
- Check GracefulPkg.jl
  - https://github.com/JuliaPluto/GracefulPkg.jl/commit/003e727ddb4ecdeb18fb59296d1c36c48c06d975
- 

---

For Julia 1.12, see #3005

Issue #93, opened 2020-04-27
Wrong backtrace for error on cell output display


karlwessel:
When there is an error while displaying the output of a cell the wrong backtrace is shown.

This is because in `runner/PlutoRunner.jl` at line 142 it uses `backtrace` instead of `catch_backtrace`.

MWE notebook:
```
### A Pluto.jl notebook ###
# v0.6.4

using Markdown
# ╔═╡ a6ebc12e-884a-11ea-1734-bf9535b30668
struct Dummy end

# ╔═╡ 01be751a-884b-11ea-2f53-f3b983edaaa7
Base.show(io::IO, t::Dummy) = error("test")

# ╔═╡ 06538a46-884b-11ea-2bb4-057a42ebb71d
Dummy()

# ╔═╡ Cell order:
# ╠═a6ebc12e-884a-11ea-1734-bf9535b30668
# ╠═01be751a-884b-11ea-2f53-f3b983edaaa7
# ╠═06538a46-884b-11ea-2bb4-057a42ebb71d
```

gives output
```
Failed to show value: 
test
Stacktrace:
 [1] format_output(::Main.workspace2.Dummy) at .julia/packages/Pluto/3iUXu/src/runner/PlutoRunner.jl:142
 [2] fetch_formatted_ans() at .julia/packages/Pluto/3iUXu/src/runner/PlutoRunner.jl:24
 [3] top-level scope at .julia/packages/Pluto/3iUXu/src/react/WorkspaceManager.jl:0
 [4] (::Distributed.var"#104#106"{Distributed.CallMsg{:call_fetch}})() at /build/julia/src/julia-1.4.1/usr/share/julia/stdlib/v1.4/Distributed/src/process_messages.jl:294
 [5] run_work_thunk(::Distributed.var"#104#106"{Distributed.CallMsg{:call_fetch}}, ::Bool) at /build/julia/src/julia-1.4.1/usr/share/julia/stdlib/v1.4/Distributed/src/process_messages.jl:79
 [6] macro expansion at /build/julia/src/julia-1.4.1/usr/share/julia/stdlib/v1.4/Distributed/src/process_messages.jl:294 [inlined]
 [7] (::Distributed.var"#103#105"{Distributed.CallMsg{:call_fetch},Distributed.MsgHeader,Sockets.TCPSocket})() at ./task.jl:358
```

fonsp:
[...]

PR #3546, opened 2026-05-14
Julia 1.13: re-wrap @doc cell result as Docs.Binding

pankgeorg:
## Summary

Restore the docstring-as-cell-output rendering on Julia 1.13+ that was broken by [JuliaLang/julia#59882](https://github.com/JuliaLang/julia/pull/59882). Fixes the failing assertions in [#3449](https://github.com/JuliaPluto/Pluto.jl/issues/3449); part of the [#3389](https://github.com/JuliaPluto/Pluto.jl/issues/3389) umbrella. Continues — but does not stack on — the closed [#3452](https://github.com/JuliaPluto/Pluto.jl/pull/3452), which had been abandoned because the author could not pin down the upstream behavior change.

## What changed in Julia 1.13

Same parsed expression, different result:

```julia
julia> Meta.parse("\"::Bool\"\nf(::Bool) = 1") |> e -> Core.eval(Main, e) |> typeof
# 1.12: Base.Docs.Binding
# 1.13: typeof(f)
```

`@doc str def` now returns the value of `def` instead of a `Base.Docs.Binding` ([JuliaLang/julia#60681](https://github.com/JuliaLang/julia/issues/60681), closed as not planned). Pluto's `format_output(::Base.Docs.Binding)` dispatch no longer fires, so documented function cells render as `f (generic function with N methods)` rather than the docstring HTML.

`Docs.doc(binding)` itself still works correctly on 1.13 and aggregates all method docstrings as before — only the cell result type changed.

## The fix

Detect the (pre-macroexpansion) `@doc` macrocall in `run_expression`, extract the bound name from the definition, and re-wrap the result as `Docs.Binding(workspace, name)` before storing in `cell_results`. This puts the cell back on the existing rendering path; no display-side changes required.

- No-op on Julia < 1.13 (where the result is already a `Docs.Binding`).
- Skipped for `CapturedException` results so error rendering is untouched.
- Skipped if the bound name isn't defined in the workspace, so a half-failed definition still falls through to the default formatter.
- Handles common doc targets: `f(x) = …`, `function f end`, `f(::T) where T`, parametric methods, `struct`, `macro`, `module`, `const`.

## Test plan

- [ ] CI: `test/MacroAnalysis.jl` "Doc strings" testset (the 11 failing assertions from #3449) passes on Julia 1.13.
- [ ] CI: same testset still passes on 1.11 and 1.12 (helper is a no-op).
- [ ] Manual: open a notebook on 1.13, run a cell like `"docstring"\nf(::Bool) = 1`, verify the rich docstring panel renders.
- [ ] Manual: redefine a method in another cell, verify both method docstrings appear in either cell's output.

## Out of scope (follow-ups from #3389)

- `warn_julia_compat()` in `src/Pluto.jl` still warns on 1.13 — should be bumped to 1.14 once the testset is green.
- [Malt.jl#101](https://github.com/JuliaPluto/Malt.jl/issues/101) and [#93](https://github.com/JuliaPluto/Malt.jl/issues/93) (1.13 interrupt failures) are still open.
- GracefulPkg.jl 1.13 audit per the umbrella issue.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/JuliaPluto/Pluto.jl", rev="pg/julia-1.13-docstring-display")
julia> using Pluto
```

### DIFF
--- a/src/runner/PlutoRunner/src/evaluation/run_expression.jl
+++ b/src/runner/PlutoRunner/src/evaluation/run_expression.jl
@@ -199,6 +199,11 @@ function doc_target_name(@nospecialize(e))
     if e.head === :(=) || e.head === :function || e.head === :macro || e.head === :(->) ||
        e.head === :where || e.head === :call || e.head === :(::) || e.head === :curly
         return isempty(e.args) ? nothing : doc_target_name(e.args[1])
+    elseif e.head === :.
+        # qualified name like `Base.conj` — pick the rightmost symbol; `Docs.Binding`
+        # in any workspace resolves to the same docs via Docs.aliasof.
+        return length(e.args) == 2 && e.args[2] isa QuoteNode && e.args[2].value isa Symbol ?
+            e.args[2].value::Symbol : nothing
     elseif e.head === :struct
         return length(e.args) >= 2 ? doc_target_name(e.args[2]) : nothing
     elseif e.head === :module
--- a/src/runner/PlutoRunner/src/evaluation/run_expression.jl
+++ b/src/runner/PlutoRunner/src/evaluation/run_expression.jl
@@ -163,6 +163,22 @@ so we re-derive the binding from the (pre-macroexpansion) expression. See JuliaL
 and Pluto issue #3449.
 """
 function doc_macrocall_binding_name(@nospecialize(expr))
+    # Peel off the `:toplevel`/`:block` wrapper that Pluto's Parse.jl always builds, ignoring
+    # `LineNumberNode`s, until we either reach the macrocall or something else.
+    while expr isa Expr && (expr.head === :toplevel || expr.head === :block)
+        inner = nothing
+        for arg in expr.args
+            arg isa LineNumberNode && continue
+            if inner === nothing
+                inner = arg
+            else
+                # more than one non-LineNumberNode arg: not a `@doc` wrapper shape
+                return nothing
+            end
+        end
+        inner === nothing && return nothing
+        expr = inner
+    end
     Meta.isexpr(expr, :macrocall) || return nothing
     name = expr.args[1]
     is_doc_macro =
--- a/src/runner/PlutoRunner/src/evaluation/run_expression.jl
+++ b/src/runner/PlutoRunner/src/evaluation/run_expression.jl
@@ -155,6 +155,64 @@ contains_macrocall(other) = false
 
 
 """
+Detect a top-level `@doc str def` expression and return the bound name (e.g. `:f`) if found.
+
+The Julia 1.13 release (JuliaLang/julia#59882) changed `@doc` so it returns the value of `def`
+rather than a `Docs.Binding`. That breaks `format_output(::Base.Docs.Binding)` cell rendering,
+so we re-derive the binding from the (pre-macroexpansion) expression. See JuliaLang/julia#60681
+and Pluto issue #3449.
+"""
+function doc_macrocall_binding_name(@nospecialize(expr))
+    Meta.isexpr(expr, :macrocall) || return nothing
+    name = expr.args[1]
+    is_doc_macro =
+        name === GlobalRef(Core, Symbol("@doc")) ||
+        name === GlobalRef(Base, Symbol("@doc")) ||
+        name === Symbol("@doc") ||
+        name == :(Core.var"@doc") ||
+        name == :(Base.var"@doc")
+    is_doc_macro || return nothing
+    # Layout: @doc <LineNumberNode> <docstring> <def>
+    length(expr.args) >= 4 || return nothing
+    return doc_target_name(expr.args[end])
+end
+
+doc_target_name(s::Symbol) = s
+function doc_target_name(@nospecialize(e))
+    e isa Expr || return nothing
+    if e.head === :(=) || e.head === :function || e.head === :macro || e.head === :(->) ||
+       e.head === :where || e.head === :call || e.head === :(::) || e.head === :curly
+        return isempty(e.args) ? nothing : doc_target_name(e.args[1])
+    elseif e.head === :struct
+        return length(e.args) >= 2 ? doc_target_name(e.args[2]) : nothing
+    elseif e.head === :module
+        return length(e.args) >= 2 && e.args[2] isa Symbol ? e.args[2] : nothing
+    elseif e.head === :const && length(e.args) >= 1
+        return doc_target_name(e.args[1])
+    end
+    return nothing
+end
+
+"""
+On Julia 1.13+, re-wrap the result of a `@doc` cell as a `Docs.Binding` so the cell renders as
+documentation rather than as a generic function summary. No-op on earlier versions, where `@doc`
+already returns a `Docs.Binding`.
+"""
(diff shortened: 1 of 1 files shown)


pankgeorg:
Another idea: 

- Make a pluto-specific show method for Functions that always gets the latest docs (after reactivity is done rendering)
  - in order to get the function to display the "same" docstring across methods, and not different, based on the timing of its execution
- Make that work in the pre - 1.12
- Keep the docs renderer in case someone actually explicitly requests it.

fonsp:
Little nsippet

```html
<div style="
font-family: JuliaMono;
padding: .4em .4em;
font-size: .8rem;
">

<span style='font-style: italic; opacity: .8;'>function</span>

<span style='font-weight: bold;'>hello_world</span>

<a style='
	font-family: system-ui;
margin-inline-start: 2ch;
' href="#">(View docs)</a>

</div>

```

j-fu:
Hi, with 1.13 release approaching I tested this PR (with current main merged in) - CI tests including those from MacroAnalysis are all green for both 1.12.6 and 1.13-rc1  (openSUSE Leap 15.6 Linux).
Also no issues with Malt.

So this would tick all the items in the test plan green.

I am living on 1.13 for quite while now including a multitude of Pluto notebook uses, and I have seen no other hiccups.

So IMHO the "unsupported" warning can be bumped and this can be merged.

j-fu:
This is what Claude Sonnet has to say:

I compared `jf/julia-1.13-docstring-display` against `main` (only one file changed: src/runner/PlutoRunner/src/evaluation/run_expression.jl) and empirically tested the fix against a real Julia 1.13.0-rc1 install plus 1.12.6 for comparison. Summary of the change first, then risks.

## What the change does
Julia 1.13 changed `@doc str def` to return the value of `def` instead of a `Docs.Binding` ([JuliaLang/julia#59882](https://github.com/JuliaLang/julia)), which broke Pluto's special `format_output(::Docs.Binding)` rendering (the nice docs panel). The fix adds `doc_macrocall_binding_name`/`doc_target_name` to re-derive the bound name from the **pre-macroexpansion** cell expression, and `maybe_rewrap_as_doc_binding` re-wraps the result as a `Docs.Binding` on Julia ≥ 1.13 only.

## Validated as working
- Ran the existing `test/MacroAnalysis.jl` "Doc strings" testset against Julia 1.13.0-rc1 with the branch code (via `TestEnv.activate()` in the repo's own project) — **all 22 assertions pass**, including the trickier `Base.conj()` doc-merging case.
- The `@static if VERSION < v"1.13.0-DEV"` guard is correct: since prerelease versions sort below their release, `1.13.0`, `1.13.0-rc*`, and `1.13.0-DEV.xxxx` nightlies all take the new path, while all 1.12.x take the no-op path.
- `@doc x` / `@doc sin` (lookup form) are correctly left alone (guarded by `length(expr.args) >= 4`), and a `begin...end` block with a stray string + multiple statements is correctly *not* misdetected as a doc cell.
- `cell_results` is only read by `format_output` (display), not by Pluto's reactivity/dependency graph, so this only affects rendering, not notebook semantics — no risk to reactivity.
- No new attack surface: this only introspects the user's own already-trusted AST; no security concern.

## Risk: the heuristic silently misses several common definition shapes

I confirmed three concrete cases where `doc_target_name` fails to find a name, so `maybe_rewrap_as_doc_binding` falls through to the pre-1.13 (broken/plain) rendering — **no error is raised, it just silently degrades**, which is what makes this risky (it won't show up as a test failure or exception, only as "the docs panel doesn't appear"):

1. **Macros**: `"doc" \n macro foo(x) ... end` extracts the bare name `:foo`, but the macro is bound as `var"@foo"`, not `foo`. `isdefined(m, :foo)` is `false`, so it's never rewrapped. Verified: cell shows `@mymac (macro with 1 method)` instead of the docs panel, whereas Julia 1.12 returns a `Docs.Binding` directly for the exact same code (so this is a real regression vs. pre-1.13 Pluto for macros).
2. **Structs with an explicit supertype** (`struct MyStruct2 <: Number ... end`): the AST for the type signature has head `:<:`, which isn't among the handled heads in `doc_target_name`, so it returns `nothing`. Verified: a plain `struct MyStruct3` gets the nice `pluto-docs-binding` HTML, but `struct MyStruct2 <: Number` renders as a bare `text/plain` cell with an **empty body**.
3. **`abstract type`/`primitive type` declarations**: their AST heads are `:abstract`/`:primitive`, neither handled at all, so any docstring on a type hierarchy declaration (a very common Julia pattern) is silently dropped.

These aren't edge cases nobody writes — subtyped structs, abstract type hierarchies, and documented macros are all idiomatic Julia. They currently work fine on 1.12 (via the old direct-`Docs.Binding` return), so this is a partial, not full, fix of the regression, and it fails invisibly.

## Other observations
- No new tests were added for `doc_macrocall_binding_name`/`doc_target_name`/`maybe_rewrap_as_doc_binding` themselves; the PR relies entirely on the pre-existing "Doc strings" testset, which happens to only exercise the shapes that already work (plain function/struct/const definitions, dotted `Base` extension). None of the three gaps above would be caught by CI.
- The approach is inherently coupled to the exact AST shape Julia's parser produces for `@doc` and to the assumption that "Pluto's Parse.jl always wraps cells in `:toplevel`/`:block`". That's fine today (validated), but it's a second, separate implicit contract with Julia's parser that could silently drift in a future Julia release without any test failure (the function fails open, returning the un-wrapped value, rather than erroring).

## Recommendation
I would not block merging over this — it's a strict improvement over the current broken 1.13 rendering for the common cases, is well-guarded by version checks, has no security/reactivity implications, and passes existing tests. But I'd flag it for follow-up before/soon after merging:
- Extend `doc_target_name` to handle `:abstract`, `:primitive`, and `:<:` (treat like `:where`/`:curly` — recurse into `args[1]`).
- Fix macro name resolution to check for the `var"@name"` binding (or `Symbol("@", name)`) in addition to the plain name.
- Add regression tests for these three shapes (and maybe `mutable struct <: X`, `abstract type <: X`) so future changes don't silently regress them again.

I asked them to implement these recommendations.

j-fu:
See #3600

j-fu:
Hi, it seems that the Julia 1.13 release is quite close now. I think we should release this (with #3600 included), and declare 1.13 as supported. This would also be a statement from plutojl.org that Pluto will live on. 

As Pluto has now full semantic versioning, it should be relatively straightforward to observe things in the wild and to make patch versions if needed.

pankgeorg:
Merged `main` into this branch (no conflicts). #3600 is rebased on top of it and is now mergeable, and #3613 stacks on #3600 to bump the compat warning and CI matrix.

Suggested merge order: #3600 into this branch, then this PR into `main`, then #3613 into `main`.

Local run of `test/MacroAnalysis.jl` on Julia 1.13.0 with the full stack: 191 passed, 1 broken, 0 failed.

🤖 Posted for @pankgeorg by an AI agent — harness: Claude Code 2.1.272 · model: claude-fable-5-1
Session: local Claude Code session session_01APWWhQTK3fPcfLdFFb5LjE

Pangoraw (approved):

PR #3554, opened 2026-05-21
fix: html render improvement

pankgeorg:
## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/JuliaPluto/Pluto.jl", rev="pg/html")
julia> using Pluto
```

### DIFF
--- a/frontend/components/Editor.js
+++ b/frontend/components/Editor.js
@@ -124,7 +124,7 @@ const statusmap = (/** @type {EditorState} */ state, /** @type {LaunchParameters
     recording_waiting_to_start: state.recording_waiting_to_start,
     is_recording: state.is_recording,
     isolated_cell_view: launch_params.isolated_cell_ids != null && launch_params.isolated_cell_ids.length > 0,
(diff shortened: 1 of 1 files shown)


frontend/components/Editor.js
```diff
@@ -124,7 +124,7 @@ const statusmap = (/** @type {EditorState} */ state, /** @type {LaunchParameters
     recording_waiting_to_start: state.recording_waiting_to_start,
     is_recording: state.is_recording,
     isolated_cell_view: launch_params.isolated_cell_ids != null && launch_params.isolated_cell_ids.length > 0,
-    sanitize_html: state.notebook.process_status === ProcessStatus.waiting_for_permission,
+    sanitize_html: state.notebook.process_status === ProcessStatus.waiting_for_permission || state.static_preview,
```
pankgeorg, line 127: > think we should make this PR conditional on where the launch paramaters came from (query or html file)

-- 🐣

PR #3561, opened 2026-05-27
Update finnish translations

eteppo:
I added remaining entries to `"t_status_names": {]` that do not exist in `english.json`. It appeared to work regardless and (I think) should not cause problems for other languages.

As a sidenote, I noticed a hardcoded "or" from /frontend/components/ErrorMessage.js, line 386. Gemini wanted to edit `, or` to `, ${t("t_key_or")}` but I didn't include it here. 

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/eteppo/Pluto.jl", rev="add-finnish-translations")
julia> using Pluto
```


fonsp:
Woohoo, very nice, thank you!

Well found with the `or`! I decided to just remove the `, or` part. It would take some extra work to localize, and it's not adding that much.

PR #3569, opened 2026-06-01
Add Russian localization

bvdmitri:
## What

Adds a Russian (`ru`) localization to Pluto's frontend.

- New translation file `frontend/lang/russian.json`
- Registers the `ru` locale in `frontend/common/lang.js` and `frontend/imports/lang_imports.js`
- Adds plural forms (`_one`/`_other`) for several keys (`t_split_this_cell_into_cells`, `t_confirm_delete_multiple_cells`, `t_logs_truncated`, `t_ready_notif_body`) so the i18n validation passes, and updates `english_generated_bases.json` accordingly

## Why

Provide a complete Russian translation for Pluto's UI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/bvdmitri/Pluto.jl", rev="russian-localization")
julia> using Pluto
```


fonsp:
Спасибо! Awesome 🎉

Some notes:
- The changes to `english.json` are not needed, because these forms are never called with 1, so it was intentional. But anything more than 2 is possible, so the russian `_few` and `_many` are both possible. You could remove the `_one` forms in russian too, but it's not needed. 
- This language is currently "Russian" with code `ru`. We could also make it more specific, like "Russian (Russia)" or `ru-RU` to distinguish it from other variants of the language. (We did this for some other languages.) We could also do this later if we add more russian variants. It depends on the context – what do you think?
- We require the localization to be reviewed and tested by a human. Could you still do that, and let me know how the quality was? (Also to give me a better idea of how good the AI systems are at this.)


Thanks again!

bvdmitri:
Thanks for feedback @fonsp , I've manually proof-read the translation, its mostly fine with me, but I replaced some of words that I think are more appropriate. In particular the translation of "Live documentation" was kinda of funny, was sort of "Living documentation" instead, I fixed it

PR #3574, opened 2026-06-03
update Persian translation

shosseinib:
As requested I've updated `persian.json` to be synced with the latest `english.json`. It adds missing translations, removes outdated keys, and minor updates for previous translations.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/shosseinib/Pluto.jl", rev="persian-translation")
julia> using Pluto
```


fonsp:
Thank you! This looks great, nice to have the complete language support again 🎉

---

I have another question:

What should the runtime and status progress look like? For example, in the status tab, it currently looks like this:

Scherm­afbeelding 2026-06-04 om 12 00 35

If a task took 400 milliseconds, and it is 90% done, it would say in English:

> TaskName (9/10) 400 ms

What should this look like in Persian, with RTL? And should we localize the time units (`s`, `ms`, `μs`, `ns`)?

Thank you!

shosseinib:
It's not the standardized way to display time in Persian literature. Some books and articles use explicit full-word units (e.g., میلی‌ثانیه for milliseconds and میکروثانیه for microseconds), while using Latin/Greek is more popular.
When using Latin/Greek letters for units, they typically write the number first, followed by the unit, like this: 
`۴۰۰ ms` for `400 ms`
not `ms ۴۰۰` and not `ms 400`.
I've tried to change the CSS for these parts, but I'm not sure if it's possible to apply it to the entire source:
Screenshot_20260604_155739

fonsp:
Alright! Is this correct?

Scherm­afbeelding 2026-06-04 om 15 10 46

shosseinib:
No, time unit signs should come after the digits. `۴۰۰ μs` not `μs ۴۰۰`. Here is the example:
image

PR #3363, opened 2025-09-15
🇮🇹 add italian translation

disberd:
Attempt to add italian translation following similar examples for other languages

One notable thing I kept `(clipboard)` next to the italian translation `appunti` as to me that still sounds a bit weird and I think a lot of people even in italy might be more familiar with `clipboard` than `appunti`, but I may be biased as it's too long I live outside of italy :D

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.activate(temp=true)
julia> Pkg.add(url="https://github.com/disberd/Pluto.jl", rev="italian_translation")
julia> using Pluto
```

PR #3584, opened 2026-06-23
🇮🇹 update italian translation with new feature strings

disberd:
This adds translation for new entries that were added after the original PR #3363 

Should address the request from @fonsp in https://github.com/JuliaPluto/Pluto.jl/pull/3363#issuecomment-4544473342

I made claude do a draft of the translation and only applied minor changes to the suggestions

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/disberd/Pluto.jl", rev="update_italian_translation")
julia> using Pluto
```


fonsp:
Amazing, thank you!!

PR #3474, opened 2026-02-10
Add Persian Translation

shosseinib:
I added complete Persian translation for Pluto.jl.
The file is attached too
[persian.json](https://github.com/user-attachments/files/25204238/persian.json)

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.activate(temp=true)
julia> Pkg.add(url="https://github.com/shosseinib/Pluto.jl", rev="main")
julia> using Pluto
```

PR #3585, opened 2026-06-24
Added arabic language support

ariguiba:
Added complete Arabic translation for Pluto.jl. This is based on the right-to-left localization from #3474 

From what I've tested so far: the selection and typing flow are similar to writing right-to-left in VS Code. Some things can be quite intuitive if you have to include some English words in between but other than that all works well. The default font size may be a bit small, especially for the settings but that could be personal preferences :) 

I've added a full notebook to show how to use this directly in Pluto here: https://github.com/JuliaPluto/featured/pull/107

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/ariguiba/Pluto.jl", rev="main")
julia> using Pluto
```


fonsp:
Amazing, it looks really good!! Thank you ✨

People can change their font size by zooming. But do you mean that the arabic text looks smaller compared to english text in the same place? Is this for cell output (like the arabic-english text side by side in your featured notebook) or for Pluto UI? How is this on other websites?

I will merge this now, and we can look at the scaling in another PR :)

ariguiba:
Hi :) Glad you like it! 

So I researched this (because I noticed the same thing when coding on VS Code in Arabic) and turns out it's a common issue that Arabic font in general all over the internet seem smaller than their Latin counterpart - This [blogpost](https://alexstrick.com/personal/2017-03-12-why-are-arabic-fonts-so-small.html) explains it well in case someone is interested :) 

They say: 

> _“Arabic letters have a smaller body relative to the extenders above and below, so the most common elements are smaller relative to everything else. Because the height of the font needs to more-or-less fit within the body size, Arabic looks smaller than Latin at the same point size.”_

My original idea was to increase the font +2/3, I mainly noticed this in PlutoUI but for the Markdown or variable names in the cells it was fine! 
So I think to stay consistent with how this is done all over the internet we should keep the same font size, since Arabic users are probably used to this

PR #3595, opened 2026-07-10
Improve ANSI white / light white output

### LINKED ISSUE #3593: `:white` and `:light_white` output can't be read
Currently, Pluto maps both ANSI 'white' and 'bright white' to `#FFFFFF`, which is completely unreadable on a white background.

Image

The code is @ https://github.com/JuliaPluto/Pluto.jl/blob/main/frontend/ansi-colors.css#L11-L19

So firstly there's no way to distinguish between white and bright white, which defeats the purpose of having two different colours. They should ideally be mapped to different hex codes. Secondly, I read through #3225 and, while it's true that there is no stipulation that white *must* be distinguishable from a white background, I think it's also just not an optimal choice and choosing to render it as `#FFFFFF` is an active way of making text harder to read for users.

Many terminal emulators / color themes don't map white to `#FFFFFF`, they tend to use a gray.

For example, the "Basic" theme on macOS Terminal.app uses `#BFBFBF` for white and `#E5E5E5` for bright white:

Image

The default iTerm2 theme uses `#C7C7C7` for white and `#FFFFFF` for bright white:

Image

Documenter.jl maps ANSI white to CSS `color: gray` which is `#808080`, and bright white is `#F5F5F5` (https://github.com/JuliaDocs/Documenter.jl/blob/master/assets/html/themes/documenter-light.css, the `sgr37` and `sgr97` classes)

And so on.

I think it would be a simple improvement to make white a gray? If you want to keep bright white at `#FFFFFF` that'd be fine (IMO).

penelopeysm:
Closes #3593 

Screenshots in light/dark mode:

Screenshot 2026-07-10 at 14 06 54

Screenshot 2026-07-10 at 14 07 09

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/penelopeysm/Pluto.jl", rev="main")
julia> using Pluto
```

### DIFF
--- a/frontend/ansi-colors.css
+++ b/frontend/ansi-colors.css
@@ -8,7 +8,7 @@
     --ansi-blue: rgb(0, 0, 187);
     --ansi-magenta: rgb(187, 0, 187);
     --ansi-cyan: rgb(0, 187, 187);
-    --ansi-white: rgb(255, 255, 255);
+    --ansi-white: rgb(190, 190, 190);
     --ansi-bright-black: rgb(85, 85, 85);
     --ansi-bright-red: rgb(255, 85, 85);
     --ansi-bright-green: rgb(0, 255, 0);
@@ -16,7 +16,7 @@
     --ansi-bright-blue: rgb(85, 85, 255);
     --ansi-bright-magenta: rgb(255, 85, 255);
     --ansi-bright-cyan: rgb(85, 255, 255);
-    --ansi-bright-white: rgb(255, 255, 255);
+    --ansi-bright-white: rgb(240, 240, 240);
 }
 
 /* Foreground colors */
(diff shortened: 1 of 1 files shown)


frontend/ansi-colors.css
```diff
@@ -126,11 +126,10 @@ pluto-log-dot.Stdout {
     --ansi-red: rgb(220, 0, 0);
     --ansi-magenta: rgb(200, 0, 200);
     --ansi-white: rgb(220, 220, 220);
-    --ansi-bright-black: rgb(50, 50, 50);
```
penelopeysm, line 129: This was being shadowed by the line below

Pangoraw:
Thanks!

PR #3600, opened 2026-07-22
Update 1.13 docstring rendering to properly include structs + macros

j-fu:
Suggested and implemented by Claude Sonnet.

* doc_target_name extended to handle :<: (subtyped structs/abstract types), :abstract, and :primitive heads
* plus a proper :macro case that returns Symbol("@name") instead of the bare name.
* Regression tests added to test/MacroAnalysis.jl covering macros, abstract types (with/without supertype), primitive types, and subtyped/parametric structs.

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/j-fu/Pluto.jl", rev="julia-1.13-struct-docstrings")
julia> using Pluto
```

### DIFF
--- a/src/runner/PlutoRunner/src/evaluation/run_expression.jl
+++ b/src/runner/PlutoRunner/src/evaluation/run_expression.jl
@@ -199,6 +199,11 @@ function doc_target_name(@nospecialize(e))
     if e.head === :(=) || e.head === :function || e.head === :macro || e.head === :(->) ||
        e.head === :where || e.head === :call || e.head === :(::) || e.head === :curly
         return isempty(e.args) ? nothing : doc_target_name(e.args[1])
+    elseif e.head === :.
+        # qualified name like `Base.conj` — pick the rightmost symbol; `Docs.Binding`
+        # in any workspace resolves to the same docs via Docs.aliasof.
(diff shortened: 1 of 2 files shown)


pankgeorg:
Rebased this branch onto the updated base `pg/julia-1.13-docstring-display` (which now has `main` merged in). The branch previously carried its own copies of the base commits on top of an older `main`, which is what made GitHub report a conflict. Your two commits are unchanged, only re-parented; the branch is now mergeable.

Verified locally on Julia 1.13.0: `test/MacroAnalysis.jl` passes (191 passed, 1 broken) including the new "Doc strings on macros types structs" testset.

The macOS 1.10/1.12 failures in the previous CI run were the unrelated `Match compiler options` precompile test in `test/packages/Basic.jl`, which also fails on `main`.

Follow-up PR that declares 1.13 supported and is stacked on this one: #3613.

🤖 Posted for @pankgeorg by an AI agent — harness: Claude Code 2.1.272 · model: claude-fable-5-1
Session: local Claude Code session session_01APWWhQTK3fPcfLdFFb5LjE

pankgeorg (approved):

PR #3603, opened 2026-08-23
add integration for AbstractDictionary

hogrobinson:
Hi, I am opening this because I cannot find a way outside of Pluto to do it. I want to update Pluto's integrations.jl to make AbstractDictionary from Dictionaries.jl display as nice expandable trees in the same way as Base Dict does already. This means that the dictionary and items inside are inspectable which was not possible before. It seems better to fix the display rather than choose different types because of how they display in pluto.

This PR:
image
and expanded:
image

Main:
image

It is particularly useful for a long type that quickly stop getting printed otherwise.
This PR:
image

Main:
image

## Try this Pull Request!
Open Julia and type:
```jl
julia> import Pkg
julia> Pkg.add(url="https://github.com/hogrobinson/Pluto.jl", rev="display_AbstractDictionary_as_tree")
julia> using Pluto
```

### DIFF
--- a/src/runner/PlutoRunner/src/integrations.jl
+++ b/src/runner/PlutoRunner/src/integrations.jl
@@ -204,6 +204,46 @@ const integrations = Integration[
             preprocess_log_msg(ps::ProgressLogging.ProgressString) = ps.progress.name
         end,
     ),
+    Integration(
+        id = Base.PkgId(UUID("85a47980-9c8c-11e8-2b9f-f7ca1fa99fb4"), "Dictionaries"),
+        code = quote
+            pluto_showable(::MIME"application/vnd.pluto.tree+object", ::Dictionaries.AbstractDictionary) = true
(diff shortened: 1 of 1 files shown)


src/runner/PlutoRunner/src/integrations.jl
```diff
@@ -204,6 +204,46 @@ const integrations = Integration[
...
+
+                    elements = []
+
+                    my_limit = get_my_display_limit(x, 1, depth, context, tree_display_limit, tree_display_limit_increase)
+                    row_index = 1
+
+                    k=keys(x)
+                    v=values(x)
+                    for pair in zip(k,v)
+                        k, v = pair
```
Pangoraw: A more julian way to do this would be:

```suggestion
                    for (k, v) in x
```

hogrobinson: Thanks for looking at this. I have updated the PR. I could not get that suggestion to work exactly but 'for (k, v) in pairs(x)' works instead which is tidier than before.

